Issue #14, opened 2017-10-04
cli: configurable spec file name 


tiborsimko:
Currently `.reana.yaml` looks hardcoded, see for example `reana_yaml_file_path`.

It would be useful to allow specifying it from CLI.  For example:

```console
$ reana-client -f reana-spec-quick-test.yaml run`
$ reana-client -f reana-spec-trying-out-different-weights.yaml  run`
[...]
$ reana-client -f reana-spec-final-run.yaml run`
```

This is similar to how e.g. `docker-compose -f ...` permits to specify any wanted file to serve as `Dockerfile`.

Note that this would be handy not only when we have several different parameter sets for the given workflow, but also for testing several different workflow engines, say Yadage and CWL, on the same code base.

tiborsimko:
@hjhsalo You may want to look at this under the "no hardcoded strings" umbrella...

tiborsimko:
BTW, we may want to ponder making the REANA spec file to be a required argument to `run`, rather than optional `-f` one, as this is more of a use case similar to `condor_submit` rather than `.travis.yml`.

hjhsalo:
PR #16 implements -f flag as optional.
Let's not close this yet, since it might be a good idea to have fileinput as a mandatory argument instead of optional flag.

tiborsimko:
Keeping it optional for REANA v0.2.0 release.

PR #16, opened 2017-10-05
cli: pass spec file as parameter `reana run`

hjhsalo:
* Add `-f / --file` option to `reana run` command that specifies which
    REANA specifications file should be used.
    (addresses #14)

  * Add `--skip-validation` option to `reana run` command in order to
    skip validation of the given specifications file, before submitting
    it to Reana Server.
    (addresses #19)

### DIFF
--- a/reana_client/utils.py
+++ b/reana_client/utils.py
@@ -23,7 +23,7 @@
 
 import json
 import logging
-from config import reana_yaml_file_path, reana_yaml_schema_file_path
+from config import reana_yaml_schema_file_path
 
 import yadageschemas
 import yaml
@@ -59,23 +59,28 @@ def load_workflow_spec(workflow_type, workflow_file, **kwargs):
     return workflow_load[workflow_type](workflow_file, **kwargs)
 
 
-def load_reana_spec():
+def load_reana_spec(filepath, skip_validation=False):
     """Load and validate reana specification file.
 
-    :raises IOError: Error while reading `reana_yaml_file_path`.
-    :raises ValidationError:g `reana_yaml_file_path` does not validate against
+    :raises IOError: Error while reading REANA spec file from given filepath`.
+    :raises ValidationError: Given REANA spec file does not validate against
         REANA specification.
     """
     try:
-        with open(reana_yaml_file_path) as f:
+        with open(filepath) as f:
             reana_yaml = yaml.load(f.read())
 
-        _validate_reana_yaml(reana_yaml)
+        if not (skip_validation):
+            logging.info('Validating REANA specification file: {filepath}'
+                         .format(filepath=filepath))
+            _validate_reana_yaml(reana_yaml)
+
         return reana_yaml
     except IOError as e:
         logging.info(
-            'Something went wrong when reading .reana.yaml: {0}'.format(
-                e.strerror))
+            'Something went wrong when reading specifications file from '
+            '{filepath} : \n'
+            '{error}'.format(filepath=filepath, error=e.strerror))
         raise e
     except Exception as e:
         raise e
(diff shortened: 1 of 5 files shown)


hjhsalo:
Should this be in a separate commit / PR?
* Add --validate_only option to reana run command in order to
  only validate the given specifications file.
  No workflows are submitted to Reana Server when using this flag.

hjhsalo:
I thought of adding default values for `user` and `organization` options for `reana run` to config.py. (like `reana_yaml_default_file_path`)

Should I include them in this PR or make a new issue and a separate PR?

tiborsimko:
>  --validate_only 

Mixing dashes and underscores may not be aesthetically pleasing :) Better to use one word only.

However, if we don't run but only validate, then it would be better not to use an option, but rather introduce a new command called `validate` that is distinct from `run`:

```
$ reana validate my-spec-file.yaml
$ reana run my-spec-file.yaml
```

If we want we can later have:

```
$ reana run --skip-validation my-spec-file.xml
```

tiborsimko:
>  `user` and `organization` 

We should probably attack these later... not in this PR. 

(But yes it would be good to have a way to specify the "operational" argumens as CLI options, plus have a way of storing them in the spec files for later archiving...)

hjhsalo:
* > Mixing dashes and underscores may not be aesthetically pleasing :) Better to use one word only.

  Agree. I didn't read documentation carefully enough yesterday, as I didn't know how click supported dashes in option and arguments names. Now I know about http://click.pocoo.org/5/parameters/#parameter-names (second paragraph), and will fix this. 

* > However, if we don't run but only validate, then it would be better not to use an option, but rather introduce a new command called validate that is distinct from run.

  I will open a new Issue about this and include `reana run --skip-validation my-spec-file.xml` in this as well.

* > > user and organization

  > would be good to have a way to specify the "operational" argumens as CLI options, plus have a way of storing them in the spec files for later archiving

  I will open a new Issue about this.

hjhsalo:
PR updated and WIP-status removed.

Regarding to comment made in #14 about making the REANA spec file to be a required argument to run, rather than optional -f one.
Should we implement this now in this PR or open a new issue and address it in near future?

reana_client/utils.py
```diff
@@ -59,23 +60,28 @@ def load_workflow_spec(workflow_type, workflow_file, **kwargs):
     return workflow_load[workflow_type](workflow_file, **kwargs)
 
 
-def load_reana_spec():
+def load_reana_spec(filepath, skip_validation=False):
```
diegodelemos: Regarding specification validation I would rather make `_validate_reana_yaml` a public function and make another command `reana validate -f <spec_file>` which uses it.

Inside `load_reana_spec` I would make validation mandatory.

hjhsalo: > make another command reana validate -f <spec_file> which uses it.

Yes. It's coming in https://github.com/hjhsalo/reana-client/blob/19_add_validate_command/reana_client/cli/analyses.py
No PR opened yet, since I want to see what gets merged to master by this PR.

> _validate_reana_yaml a public function
> Inside load_reana_spec I would make validation mandatory.

If I make changes to `_validate_reana_yaml`, I need to reuse the yaml loading part from `load_reana_spec` either in calling function or in the new `validate_reana_yaml`?

diegodelemos: If we move `yaml.load()` inside `validate_reana_yaml` I would say that this function returns the spec as a dict so you can use it in `load_reana_spec`. But if we do so, `validate_reana_yaml` and `load_reana_spec` are the same function in the end so there is no need to have two, so we have two options:

1. We go for just one function which includes both loading and validation specific code, so when we want to validate or load we call the same.
2. We keep then as they are, and we always call `load_reana_yaml` either for validating or loading but having place for validation code and another for loading.

The difference for me lies in future changes to validation and loading keeping concerns separate or in the same function. I would go for the second 😄

PR #24, opened 2017-10-11
cli: add `reana validate` command

### LINKED ISSUE #19: cli: add `validate` command
In order to support validation of a REANA specification file, a command like `reana validate [specfile]` should be added.

In addition a flag which skips the validation step in `reana run` could be added as well (e.g `reana run --skip-validation`).

hjhsalo:
* Add `reana validate` command to validate a REANA specification file.
    Uses either default value (e.g. .reana.yml) or a file specified with
    `-f / --file` -option (e.g. `reana validate -f my-spec-file.yml`).
    (closes #19)

### DIFF
--- a/reana_client/cli/analyses.py
+++ b/reana_client/cli/analyses.py
@@ -30,6 +30,28 @@ from ..config import (default_organisation, default_user,
 from ..utils import load_reana_spec, load_workflow_spec
 
 
+@click.command()
+@click.option('-f',
+              '--file',
+              type=click.Path(exists=True, resolve_path=True),
+              default=reana_yaml_default_file_path,
+              help='REANA specifications file describing the workflow and '
+                   'context which REANA should execute.')
+@click.pass_context
+def validate(ctx, file):
+    """Validate given REANA specification file."""
+    try:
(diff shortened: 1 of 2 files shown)


reana_client/cli/analyses.py
```diff
@@ -30,6 +30,28 @@
...
 
+@click.command()
+@click.option('-f',
+              '--file',
+              type=click.Path(exists=True, resolve_path=True),
+              default=reana_yaml_default_file_path,
+              help='REANA specifications file describing the workflow and '
+                   'context which REANA should execute.')
+@click.pass_context
+def validate(ctx, file):
```
diegodelemos: There is only left to add this to `reana_client/cli/__init__.py`:
```python
cli.add_command(analyses.validate)
```
so click takes it into account

diegodelemos:
ping @hjhsalo

reana_client/cli/analyses.py
hjhsalo: :man_facepalming: 
Fixed

diegodelemos: thanks!! we :shipit:

Issue #43, opened 2017-11-03
cli: better error reporting


anton-khodak:
Currently, reana-client reports errors in debug mode only and with a very small fraction of the accessible information so additional debugging is necessary in almost all cases to understand what caused the error. The goal is to help a user to diagnose a problem right after the first failure.

Example of current debugging output:
```
/Users/anton/.virtualenvs/reana-py3/bin/python /Users/anton/Programming/NYU-CERN/ak-reana-client/reana_client/cli/cwl_runner.py binding-test.cwl bwa-mem-job.json
/Users/anton/Programming/NYU-CERN/ak-reana-client/reana_client/cli/cwl_runner.py 1.0.20171219150536
INFO:cwltool:/Users/anton/Programming/NYU-CERN/ak-reana-client/reana_client/cli/cwl_runner.py 1.0.20171219150536
Resolved 'binding-test.cwl' to 'file:///Users/anton/Programming/CWL/cwltool/cwltool/schemas/v1.0/v1.0/binding-test.cwl'
INFO:cwltool:Resolved 'binding-test.cwl' to 'file:///Users/anton/Programming/CWL/cwltool/cwltool/schemas/v1.0/v1.0/binding-test.cwl'
DEBUG:past.translation:Running find_module: rdflib.plugins.memory...
DEBUG:past.translation:Running load_module for rdflib.plugins.memory...
DEBUG:past.translation:Excluded rdflib.plugins.memory from translation
DEBUG:past.translation:Running find_module: netrc...
DEBUG:past.translation:Running load_module for netrc...
DEBUG:past.translation:Excluded netrc from translation
DEBUG:past.translation:Running find_module: shlex...
DEBUG:past.translation:Running load_module for shlex...
DEBUG:past.translation:Excluded shlex from translation
Traceback (most recent call last):
  File "/Users/anton/Programming/NYU-CERN/ak-reana-client/reana_client/cli/cwl_runner.py", line 91, in cwl_runner
    reana_spec)
  File "/Users/anton/Programming/NYU-CERN/ak-reana-client/reana_client/api/client.py", line 129, in create_workflow
    raise e
  File "/Users/anton/Programming/NYU-CERN/ak-reana-client/reana_client/api/client.py", line 119, in create_workflow
    reana_spec, sort_keys=True))).result()
  File "/Users/anton/.virtualenvs/reana-py3/lib/python3.6/site-packages/bravado/http_future.py", line 56, in wrapper
    return func(self, *args, **kwargs)
[...]

PR #44, opened 2017-11-03
cli: change logging mode for errors and print richer info

anton-khodak:
Addresses #43

### DIFF
--- a/reana_client/cli/analyses.py
+++ b/reana_client/cli/analyses.py
@@ -22,6 +22,7 @@
 """REANA client workflow related commands."""
 
 import logging
+import traceback
 
 import click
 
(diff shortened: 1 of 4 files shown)


reana_client/cli/analyses.py
```diff
@@ -61,7 +62,7 @@ def analysis_validate(ctx, file):
     except Exception as e:
         logging.info('Something went wrong when trying to validate {0}'
                      .format(click.format_filename(file)))
-        logging.debug(str(e))
+        logging.error(traceback.print_exc())
```
tiborsimko: `traceback.format_exc()` here and elsewhere?

anton-khodak: force pushed the commit with `print_exc()`

PR #59, opened 2018-01-25
cli: create intermediate dirs when writing outputs

diegodelemos:


reana_client/cli/outputs.py
```diff
@@ -162,7 +162,12 @@ def outputs_download(ctx, user, organization, workflow, file_, output_folder):
                                                              file_name)
             logging.info('{0} binary file downloaded ... writing to {1}'.
                          format(file_name, output_folder))
-            with open(os.path.join(output_folder, file_name), 'wb') as f:
+
+            outputs_file_path = os.path.join(output_folder, file_name)
+            if not os.path.exists(os.path.dirname(outputs_file_path)):
+                    os.makedirs(os.path.dirname(outputs_file_path))
```
tiborsimko: `exist_ok` is by default `False`, so if a user attempts to re-download a file an exception would be raised... Shall we create some nice message back to the user?  

(We could add CLI a flag later to let the user govern this behaviour, but not right now...)

reana_client/cli/outputs.py
```diff
@@ -162,10 +162,18 @@ def outputs_download(ctx, user, organization, workflow, file_, output_folder):
...
+
+            with open(outputs_file_path, 'wb') as f:
                 f.write(binary_file)
             click.echo('File {0} downloaded to {1}'.format(file_name,
                                                            output_folder))
+        except OSError as e:
```
diegodelemos: This `exists_ok` is python 3 (see [here](https://docs.python.org/3/library/os.html#os.makedirs)) but makedirs in python 2 would also through an `OSError` (see [here](https://docs.python.org/2/library/os.html#os.makedirs)). I have tried to `reana-client outputs download fitdata/plot.png` in two cases: existing `outputs/fitdata/plot.png` and removing the file but keeping the directory so `outputs/fitdata` exists. In both cases it works and overwrites. Just in case we add an OSError except block for writing errors.

PR #78, opened 2018-03-01
[WIP] global: addition of .dockerignore

dinosk:
Signed-off-by: Dinos Kousidis <konstantinos.kousidis@cern.ch>

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,5 +1,5 @@
 # This file is part of REANA.
-# Copyright (C) 2017 CERN.
+# Copyright (C) 2018 CERN.
 #
 # REANA is free software; you can redistribute it and/or modify it under the
 # terms of the GNU General Public License as published by the Free Software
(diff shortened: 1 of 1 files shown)


dinosk:
A first idea for the .dockerignore. When we finalize what we need in it I'll add it and in the other components.

.dockerignore
```diff
@@ -0,0 +1,31 @@
...
+#
+# In applying this license, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization or
+# submit itself to any jurisdiction.
+
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
+
+.git
```
tiborsimko: We discussed keeping the `git` in the container for now, as it may help with debugging of "shared" containers while we are in the alpha stage.

Also, `reana-client` does not really need a dockerignore file, since we are not building the Docker image here.
Basically only the repositories with `Dockerfile` would need those.

dinosk: Ah yes right!

PR #79, opened 2018-03-01
cli: support file seeding with absolute paths

### LINKED ISSUE #67: cli: seeding inputs and code with absolute paths
Currently `reana-client code upload FILE` and `reana-client inputs upload FILE` expect as arguments filenames that are located in the `./code` and `./inputs` directories.  One cannot upload from another place:

```console
$ reana-client inputs upload ./README.rst 
[INFO] REANA Server URL ($REANA_SERVER_URL) is: http://192.168.99.100:32658
[INFO] Workflow "e4a3305b-a11f-401e-b836-eb426fc191dd" selected
Traceback (most recent call last):
  File "/home/simko/.virtualenvs/reana-client/bin/reana-client", line 11, in <module>
    sys.exit(cli())
  File "/home/simko/.virtualenvs/reana-client/lib/python2.7/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/home/simko/.virtualenvs/reana-client/lib/python2.7/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/home/simko/.virtualenvs/reana-client/lib/python2.7/site-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/simko/.virtualenvs/reana-client/lib/python2.7/site-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/simko/.virtualenvs/reana-client/lib/python2.7/site-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/simko/.virtualenvs/reana-client/lib/python2.7/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/home/simko/.virtualenvs/reana-client/lib/python2.7/site-packages/click/decorators.py", line 17, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/home/simko/.virtualenvs/reana-client/lib/python2.7/site-packages/reana_client/cli/inputs.py", line 155, in inputs_upload
    format(f.name),
UnboundLocalError: local variable 'f' referenced before assignment
```

However, using `code` and `inputs` subdirectories is simply a convention advising people how to structure their analyses.  It is very well possible that some people prefer working in a single workspace directory only, or that some people prefer calling their directories `data` (instead of `inputs`) and `src` (instead of `code`), etc.  

We should be agnostic to the naming people use, and translate their convention into REANA newspeak (that would continue using inputs and code internally).
 
This can be achieved by simply requiring `FILE` to be an absolute/relative path to the current working directory, not assuming anything about the required structure.  (An additional positive side outcome out of this would be that we wouldn't need `--inputs-directory` argument in this case, which would simplify simplify the function signature when someone runs `-help`.)

Some of the use case examples:

```console
# all these will upload to /reana/.../inputs/mydata.csv
$ reana-client inputs upload mydata.csv # look inside `inputs` subdir for backwards compatibility
$ reana-client inputs upload ./mydata.csv 
$ reana-client inputs upload ./data/mydata.csv 
$ reana-client inputs upload ./inputs/mydata.csv
$ reana-client inputs upload ./stuff/mydata.csv

# possible later extension to upload several files at once:
$ reana-client inputs upload ./mydata/*.csv

# beware of uploading files with leading subdirectories
$ reana-client inputs upload ./data/small-sample/mydata.csv
$ reana-client inputs upload ./data/big-sample/mydata.csv

# possible later extension to recursively upload everything
$ reana-client inputs upload -r ./inputs/*

# possible later extension for non-file-upload seeding
$ reana-client inputs upload http://example.org/mydata.csv
$ reana-client inputs upload root://example.org/mydata.csv
```

Note that using absolute/relative path for FILE will make a nice coherent picture with `reana.yaml` where we already use those paths:

```yaml
inputs:
  files:
    - inputs/names.txt
```

Someone preferring other nomenclature could use:

```yaml
inputs:
  files:
    - mydata/names.txt
```

and in CLI:

```console
$ reana-client inputs upload # seed everything listed in reana yaml file
$ reana-client inputs upload mydata/names.txt # as seen in reana yaml file
```

**P.S.** Note that some of the above use cases, such as the recursive upload of directories, will be for future milestones... It's simply good to keep all these various use cases in mind when we start refactoring the current inputs-name-is-hardcoded into the wanted inputs-name-is-optional behaviour.

hjhsalo:
* Add support for specifying files that should be
  uploaded to workflow workspace by absolute paths (closes #67).

### DIFF
--- a/reana_client/api/client.py
+++ b/reana_client/api/client.py
@@ -479,26 +479,21 @@ class Client(object):
                 for root, dirs, files in os.walk(path, topdown=False):
                     responses = []
                     for next_path in files + dirs:
-                        try:
-                            response = self.upload_to_server(
-                                user, organization, workflow,
-                                os.path.join(root, next_path),
-                                upload_type)
-                            responses.append(os.path.join(root, next_path))
-                        except Exception as e:
-                            logging.error(traceback.format_exc())
-                            logging.error("Something went wrong"
-                                          " while uploading '{0}'".
-                                          format(os.path.join(root,
-                                                              next_path)))
+                        response = self.upload_to_server(
+                            user, organization, workflow,
+                            os.path.join(root, next_path),
+                            upload_type)
+                        responses.append(os.path.join(root, next_path))
                 return responses
 
             # Check if input is an absolute path and upload file.
             else:
                 with open(path) as f:
                     fname = os.path.basename(f.name)
-                    if len(path.split('/')) > 1 and \
-                            path.split('/')[0] == UploadType(upload_type).name:
+                    # Remove prepending dirs named "." or as the upload type
+                    while len(path.split('/')) > 1 and \
+                            path.split('/')[0] in \
+                            [UploadType(upload_type).name, '.']:
                         path = "/".join(path.strip("/").split('/')[1:])
                     logging.debug("'{}' is an absolute filepath."
                                   .format(os.path.basename(fname)))
@@ -517,9 +512,9 @@ class Client(object):
                         if response:
                             logging.info("File '{}' was successfully "
                                          "uploaded.".format(fname))
-                        return response
                     except Exception as e:
-                        logging.error(traceback.format_exc())
-                        logging.error("Something went wrong"
-                                      " while uploading '{0}'".
-                                      format(path))
+                        logging.debug(traceback.format_exc())
+                        logging.debug(str(e))
+                        logging.info("Something went wrong while uploading {}".
+                                     format(fname))
+                    return response
--- a/reana_client/api/client.py
+++ b/reana_client/api/client.py
@@ -31,6 +31,7 @@ import pkg_resources
 from bravado.client import SwaggerClient
 from bravado.exception import HTTPError
 from reana_client.errors import FileUploadError
+from reana_client.utils import get_analysis_root
 
 
 class UploadType(enum.Enum):
@@ -472,6 +473,7 @@ class Client(object):
 
             # Check if input is a directory and upload everything
             # including subdirectories.
+
             if os.path.isdir(path):
                 logging.debug("'{}' is a directory.".format(path))
                 logging.info("Uploading contents of folder '{}' ..."
@@ -490,21 +492,31 @@ class Client(object):
             else:
                 with open(path) as f:
                     fname = os.path.basename(f.name)
+                    analysis_root = get_analysis_root()
+                    if not path.startswith(analysis_root):
+                        raise FileUploadError(
+                            'Files and directories to be uploaded'
+                            'must be under the analysis root directory.')
+                    # Calculate the path that will store the file
+                    # in the workflow controller, by subtracting
+                    # the analysis root path from the file path
+                    save_path = path.replace(analysis_root, '')
                     # Remove prepending dirs named "." or as the upload type
-                    while len(path.split('/')) > 1 and \
-                            path.split('/')[0] in \
+                    while len(save_path.split('/')) > 1 and \
+                            save_path.split('/')[0] in \
                             [UploadType(upload_type).name, '.']:
-                        path = "/".join(path.strip("/").split('/')[1:])
+                        save_path = "/".join(
+                            save_path.strip("/").split('/')[1:])
                     logging.debug("'{}' is an absolute filepath."
                                   .format(os.path.basename(fname)))
                     logging.info("Uploading '{}' ...".format(fname))
                     try:
                         if upload_type is UploadType.code:
                             response = self.seed_analysis_code(
-                                user, organization, workflow, f, path)
+                                user, organization, workflow, f, save_path)
                         elif upload_type is UploadType.inputs:
                             response = self.seed_analysis_inputs(
-                                user, organization, workflow, f, path)
+                                user, organization, workflow, f, save_path)
                         else:
                             logging.warning("Unknown upload type of '{}'."
                                             "File '{}' was not uploaded."
--- a/reana_client/api/client.py
+++ b/reana_client/api/client.py
@@ -21,6 +21,8 @@
 # submit itself to any jurisdiction.
 """REANA REST API client."""
 
+import traceback
+import enum
 import json
 import logging
 import os
@@ -28,6 +30,14 @@ import os
 import pkg_resources
 from bravado.client import SwaggerClient
 from bravado.exception import HTTPError
+from reana_client.errors import FileUploadError
+
+
+class UploadType(enum.Enum):
+    """Possible workflow status list enum."""
+
+    inputs = 0
+    code = 1
 
 
 class Client(object):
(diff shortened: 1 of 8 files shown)


reana_client/cli/code.py
```diff
@@ -120,49 +125,20 @@ def code_list(ctx, user, organization, workflow, filter, output_format):
...
-
-    else:
-        click.echo(
-            click.style('Workflow name must be provided either with '
-                        '`--workflow` option or with `$REANA_WORKON` '
-                        'environment variable',
-                        fg='red'),
-            err=True)
-
+    logging.debug('command: {}'.format(ctx.info_name))
```
hjhsalo: For the reviewer(s): Which one of these do you consider to be most suitable?

hjhsalo: For the reviewer(s): Which one of these do you consider to be most suitable?

reana_client/utils.py
```diff
@@ -127,3 +130,104 @@ def _validate_reana_yaml(reana_yaml):
...
+    """
+    if workflow:
+        logging.info('Workflow "{}" selected'.format(workflow))
+        if filenames:
+            for filename in filenames:
+                try:
+                    # Check if input is a folder and upload everything
+                    # including subdirectories.
+                    if os.path.isdir(filename):
+                        # TODO: Implement actual folder upload (with globbing)
```
diegodelemos: @tiborsimko should we move `TODO`s into issues and removing them from the code ?

reana_client/utils.py
```diff
@@ -127,3 +130,104 @@ def _validate_reana_yaml(reana_yaml):
...
+    Shared e.g. by `code upload` and `inputs upload`.g
+
+    :param user: User ID
+    :param organization: Organization ID
+    :param workflow: ID of that Workflow whose workspace should be
+    used to store the files.
+    :param filenames: Absolute filepath(s) of files to be uploaded.
+    :param client: Instance of Bravado Client.
+    """
+    if workflow:
```
diegodelemos: What do you think about raising `ValueError` with the message you have inside the `else`?

```python
if not workflow:
    raise ValueError('Workflow name must be provided either with `--workflow` option or with `$REANA_WORKON environment variable` ')
```
And let it be catch by the CLI layer so it will format using click superpowers. It sounds a bit weird to use `click` inside the generic `utils.py`. Moreover, the message shouldn't have `cli` "things" like `--workflow`, imagine someone installs `reana-client` to use the Python API and while using `upload_files_to_server` they get `cli` related messages.

hjhsalo: Agree.

hjhsalo: @diegodelemos 
I actually did a "click-free" implementation of `upload_files_to_server` in `utils.py`, but then I thought that we might run into similar cli-level code (de)duplication situation in future and introduced a `common.py` that is under `cli-module`.
Basically this file would include stuff that multiple cli commands use.
Tell me what you think and if it is not a good idea I will change the PR to use "click-free" implementation of `upload_files_to_server`.

reana_client/cli/common.py
```diff
@@ -0,0 +1,125 @@
...
+                    # filepath.
+                    elif os.path.isabs(filename):
+                        with click.open_file(filename) as f:
+                            name = os.path.basename(f.name)
+                            logging.debug('{} is an absolute filepath.'
+                                          .format(os.path.basename(name)))
+                            click.echo('Uploading \'{}\' ...'.format(name))
+
+                            ctx = click.get_current_context()
+                            if 'code upload' in ctx.command_path:
```
diegodelemos: I find this a bit fragile, if `command_path` changes the format at come point this would break, could we maybe have a new parameter which would represent the type instead?

reana_client/cli/common.py
```diff
@@ -0,0 +1,125 @@
...
+    Shared e.g. by `code upload` and `inputs upload`.
+
+    :param user: User ID
+    :param organization: Organization ID
+    :param workflow: ID of that Workflow whose workspace should be
+        used to store the files.
+    :param filenames: Absolute filepath(s) of files to be uploaded.
+    :param client: Instance of Bravado Client.
+    """
+    if workflow:
```
diegodelemos: I think this https://github.com/reanahub/reana-client/pull/79#discussion_r171538225 regarding throwing an exception still applies here, just to make the code more readable 😉

reana_client/cli/common.py
```diff
@@ -0,0 +1,125 @@
...
+        logging.info('Workflow "{}" selected'.format(workflow))
+        if filenames:
+            for filename in filenames:
+                try:
+                    # Check if input is a folder and upload everything
+                    # including subdirectories.
+                    if os.path.isdir(filename):
+                        # TODO: Implement actual folder upload (with globbing)
```
diegodelemos: I think that this https://github.com/reanahub/reana-client/pull/79#pullrequestreview-100404621 still applies as well here,  please recheck

reana_client/cli/common.py
```diff
@@ -0,0 +1,125 @@
...
+
+    """
+    command = []
+    while ctx.parent:
+        command.append(ctx.info_name)
+        ctx = ctx.parent
+    return '.'.join(str(x) for x in reversed(command))
+
+
+def upload_files_to_server(user, organization, workflow, filenames, client):
```
diegodelemos: Regarding having this as a `click` utility function, I still think that maybe not all the function but part of it should  be part of the Pythonic API the module offers, really soon CAP will ingest not only one but several file because the inputs/code for an analysis won't be just one or two files, what do you think, shall we have a quick talk about the design IRL?

hjhsalo: Just to double check the function signature would be
```
def upload_files_to_server(user, organization, workflow, filenames, api_client, upload_type):
    :param user:
    :type user: str
    :param organization:
    :type organization: str
    :param workflow:
    :type workflow: str
    :param filenames:
    :type filenames: str[]
    :param api_client:
    :type api_client: Bravado SwaggerClient
    :param upload_type:
    :type upload_type: str/enum either 'code' or 'upload'
```

reana_client/api/client.py
```diff
@@ -420,3 +430,95 @@ def get_analysis_code(self, user, organization, analysis_id):
             raise Exception(e.response.json()['message'])
         except Exception as e:
             raise e
+
+    def upload_to_server(self, user, organization, workflow,
+                         filenames, upload_type):
+        """Upload file or directory to REANA-Server.
```
tiborsimko: Regarding the work on supporting directory upload,  perhaps we should be using "pathname" instead of "filename" in arguments etc, in order to better indicate that the inputs could be files or directiories?

We should beware of situations where the same file name can be nested in several directories:

```
code/filtering/run.py
code/generating/run.py
code/something/even/deeper/nested/run.py
```
We should properly define the behaviour, just reading the file from that local place, or even creating the subdirectory structure on the server[1]. This is basically why we chose with @hjhsalo to simply say "directory upload not supported yet" when the input argument were a directory.

[1] A technique I was thinking about regarding subdirectory upload could be something simple like `rsync`-ing files with their full parent directories to the destination place, e.g. by zero-zipping the input directory on the client and zero-unzipping it on the server.  Could be useful in case there are tons of inputs files. Could speak IRL about this.)

tiborsimko: Quick summary of IRL discussion:

- two nested file upload cases:

  ``r-c type upload a/b/c/foo.ext -> type/a/b/c/foo.ext OR type/b/c/foo.ext if type==a``

- two directory walk through upload cases:

  ``r-c type upload dirname/... -> type/dirname/... OR type/... if type==dirname``

Beware of mkdir-p-style creation of nested directiories in r-w-controller and beware of `../../..` like arguments. 

This should accommodate people using the recommended structure:

```
./code/mycode.py
./inputs/mydata.csv
```

or any other structure such as flat one:

```
./mycode.py
./mydata.csv
```

or arbitrary nested one:


```
./measurements/paris-2017/test-run-12b/mydata.csv
./thesis/chapter7/scripts/mycode.py
```

In the latter cases, we would recommend using `inputs` if people don't need the runtime code distinction, but that's another story where `type` could be eventually even left out...

dinosk:
@tiborsimko I added in this PR the '-w/--workflow' option that was missing

reana_client/api/client.py
```diff
@@ -420,3 +430,96 @@ def get_analysis_code(self, user, organization, analysis_id):
...
+                            logging.warning("Unknown upload type of '{}'."
+                                            "File '{}' was not uploaded."
+                                            .format(upload_type, path))
+                        if response:
+                            logging.info("File '{}' was successfully "
+                                         "uploaded.".format(fname))
+                        return response
+                    except Exception as e:
+                        logging.error(traceback.format_exc())
+                        logging.error("Something went wrong"
```
tiborsimko: Currently this gives:

```console
$ reana-client code upload ./code/somethingthatdoesnotexist
Something went wrong while uploading ./code/somethingthatdoesnotexist
```

This looks obscure; it would be nicer to say that the file or directory does not exist.

Can you please introduce a check whether the input is a file or directory and fail with an appropriate error message? E.g. does not exist, does not have read permissions, etc...

reana_client/api/client.py
tiborsimko: I think the `OR type/... if type==dirname` parts are not implemented yet.  

I tried running the ROOT6 example and I saw:

```
/reana/default/00000000-0000-0000-0000-000000000000/analyses/f1f421fc-b6cb-4c3f-9ec7-47c02e55ec71/yadage_inputs
/reana/default/00000000-0000-0000-0000-000000000000/analyses/f1f421fc-b6cb-4c3f-9ec7-47c02e55ec71/yadage_inputs/code
/reana/default/00000000-0000-0000-0000-000000000000/analyses/f1f421fc-b6cb-4c3f-9ec7-47c02e55ec71/yadage_inputs/code/code
/reana/default/00000000-0000-0000-0000-000000000000/analyses/f1f421fc-b6cb-4c3f-9ec7-47c02e55ec71/yadage_inputs/code/code/fitdata.C
/reana/default/00000000-0000-0000-0000-000000000000/analyses/f1f421fc-b6cb-4c3f-9ec7-47c02e55ec71/yadage_inputs/code/code/gendata.C
/reana/default/00000000-0000-0000-0000-000000000000/analyses/f1f421fc-b6cb-4c3f-9ec7-47c02e55ec71/yadage_inputs/inputs
/reana/default/00000000-0000-0000-0000-000000000000/analyses/f1f421fc-b6cb-4c3f-9ec7-47c02e55ec71/code
/reana/default/00000000-0000-0000-0000-000000000000/analyses/f1f421fc-b6cb-4c3f-9ec7-47c02e55ec71/code/code
/reana/default/00000000-0000-0000-0000-000000000000/analyses/f1f421fc-b6cb-4c3f-9ec7-47c02e55ec71/code/code/fitdata.C
/reana/default/00000000-0000-0000-0000-000000000000/analyses/f1f421fc-b6cb-4c3f-9ec7-47c02e55ec71/code/code/gendata.C
```

Note the `./code/code/...` nested directory.  This makes the example not run.

We should upload `./code/...` in this case since type==code.

reana_client/api/client.py
dinosk: Could we return the exception message?
For a file or dir that doesn't exist it would show:
```
(reana-cwl) ➜  reana-demo-root6-roofit git:(master) ✗ reana-client code upload code/code_3
[Errno 2] No such file or directory: 'code/code_3'
```

I could add the check for the 2 cases, but this would show information about other exceptions that could occur.

reana_client/api/client.py
dinosk: I think this is in: https://github.com/reanahub/reana-client/pull/79/files#diff-4f27f08c3f46f0b755c1d5ac1a1e2dcaR501
The check if the first directory is equal to the upload type is done only for files, as if it is a directory it will walk it, and call the function again, eventually with the file path.

When uploading with the command from the example, it returns an error because it doesn't assume I'm pointing to the code/ dir:
```
reana-client code upload fitdata.C
[Errno 2] No such file or directory: 'fitdata.C'
```

By specifying the code/ dir:
```
reana-client code upload code/fitdata.C
code/fitdata.C was uploaded successfully.

reana-client code upload code/gendata.C 
code/gendata.C was uploaded successfully.
```

and in the `workflow-controller` container the files are at the path:
```
/reana/default/00000000-0000-0000-0000-000000000000/analyses/b6cce0ad-569c-4cdf-a17e-fee6204ec5e1/code
```

If uploading the whole code/ dir with:
```
reana-client code upload code
code/gendata.C was uploaded successfully.
code/fitdata.C was uploaded successfully.
```
 they are again at the same path.

If we upload sth outside of the code/ directory:
```
reana-client code upload a.txt
a.txt was uploaded successfully.
```
It'll be placed in code/ in the `workflow-controller` and the example runs ok.
Maybe I'm missing sth else, but I think this should be ok.

reana_client/api/client.py
dinosk: Right, in the last changes, click.Path(exists=True, ...) shows a nice message, and the exception message is reverted.

reana_client/api/client.py
```diff
@@ -420,3 +430,96 @@ def get_analysis_code(self, user, organization, analysis_id):
...
+                        elif upload_type is UploadType.inputs:
+                            response = self.seed_analysis_inputs(
+                                user, organization, workflow, f, path)
+                        else:
+                            logging.warning("Unknown upload type of '{}'."
+                                            "File '{}' was not uploaded."
+                                            .format(upload_type, path))
+                        if response:
+                            logging.info("File '{}' was successfully "
+                                         "uploaded.".format(fname))
```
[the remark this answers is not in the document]
dinosk: This will be very helpful, but it'll require moving some of the logic from `client.py` to `cli.py`.

PR #80, opened 2018-03-01
cli: less verbosity

### LINKED ISSUE #72: cli: less verbosity by default
Currently `reana-client` is too verbose, for example:

```console
$ reana-client code upload gendata.C
[INFO] REANA Server URL ($REANA_SERVER_URL) is: http://192.168.99.100:32658
[INFO] Workflow "3be010aa-b3b5-408c-9d16-17f0518a6995" selected
Uploading ./code/gendata.C ...
File ./code/gendata.C was successfully uploaded.
```

The `[INFO]` lines should not be printed by default.  They could be printed only when the user requires higher verbosity.

diegodelemos:
* (closes #72)

* Prettifies commands outputs.

### DIFF
--- a/reana_client/cli/__init__.py
+++ b/reana_client/cli/__init__.py
@@ -54,15 +54,15 @@ class Config(object):
     '--loglevel',
     '-l',
     help='Sets log level',
-    type=click.Choice(['debug', 'info']),
-    default='info')
+    type=click.Choice(['DEBUG', 'INFO', 'WARNING']),
+    default='WARNING')
 @click.pass_context
 def cli(ctx, loglevel):
     """REANA Client for interacting with REANA Server."""
     logging.basicConfig(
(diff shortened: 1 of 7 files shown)


reana_client/cli/__init__.py
```diff
@@ -54,15 +54,15 @@ def __init__(self):
...
     help='Sets log level',
-    type=click.Choice(['debug', 'info']),
-    default='info')
+    type=click.Choice(['DEBUG', 'INFO', 'WARNING']),
+    default='WARNING')
 @click.pass_context
 def cli(ctx, loglevel):
     """REANA Client for interacting with REANA Server."""
     logging.basicConfig(
         format=DEBUG_LOG_FORMAT if loglevel == 'debug' else LOG_FORMAT,
```
dinosk: I think it should be 'DEBUG'

PR #81, opened 2018-03-01
cli: support persistent workflow run names

### LINKED ISSUE #70: cli: persistent workflow run names
Currently, the workflow run name as it is displayed by `reana-client` is very volatile and changes in between two calls:

```console
$ reana-client workflow status 2> /dev/null | tail -1
objective_euclid|a9329917-5888-44a2-b57e-b7f882799d20|00000000-0000-0000-0000-000000000000|default     |finished
$ reana-client workflow status 2> /dev/null | tail -1
dreamy_kare|a9329917-5888-44a2-b57e-b7f882799d20|00000000-0000-0000-0000-000000000000|default     |finished
```

The fact that NAME does not persist between calls will be rather confusing to users.

There are two ways how to improve the situation:

1. Hide the name in all outputs and expose only UUIDs.  *(easier but less pretty)*

2. Keep name presistent across calls, once it is generated.  *(harder but more user-friendly)*

I favour the second, because then people would be able to give their workflow runs a meaningful name:

```console
$ reana-client workflow create -f reana-small-samples.yaml -n mytestrun
$ reana-client workflow create -f reana-big-samples.yaml -n mybigrun
```

It would be easier later to do:

```console
$ reana-client workflow list
mytestrun
mybigrun
$ reana-client workflow status -n mytestrun
$ reana-client workflow status -n mybigrun
```

without using UUIDs.

Here, using a new CLI option `-n` is just an example. We could have two persistent workflow run identifiers UUID (-w) and NAME (-n) that `reana-client` and `REANA_WORKON` would understand. Or we could use only one NAME (-w) that either would be supplied by users, or that we would generate (e.g. `amazing_feynman`) if the user does not specify it.  But once we generate  `amazing_feynman`, then we should use it persistently.

hjhsalo:
* Add support for persistent workflow run names
  (closes #70)


reana_client/cli/common.py
```diff
@@ -0,0 +1,67 @@
...
+        Name might be in format 'reana.workflow.123' with arbitrary
+        number of dot-delimited substrings, where last substring specifies
+        the run number of the workflow this workflow name refers to.
+
+        If name does not contain a valid run number, name without run number
+        is returned.
+    """
+    # Try to split a dot-separated string.
+    try:
+        name, run_number = workflow_name.rsplit('.', maxsplit=1)
```
diegodelemos: the `maxsplit` is not keyword argument in python 2, if you go for blah.rsplit('.', 1) it will work for both 2 and 3

reana_client/cli/common.py
```diff
@@ -0,0 +1,67 @@
...
+        click.echo(click.style(
+            'Workflow name must be provided either with '
+            '`--workflow` option or with `$REANA_WORKON` '
+            'environment variable', fg='red'),
+            err=True)
+    else:
+        return value
+
+
+def get_workflow_name_and_run_number(workflow_name):
```
diegodelemos: I would move this to `utils.py` since it is not click specific

reana_client/cli/common.py
```diff
@@ -0,0 +1,67 @@
...
+#
+# In applying this license, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization or
+# submit itself to any jurisdiction.
+"""Cli functionality shared between multiple modules of REANA-Client."""
+
+import click
+
+
+def workflow_uuid_or_name(ctx, param, value):
```
diegodelemos: I would maybe delete this file and add `workflow_uuid_or_name ` to the top of `workflow.py`

Issue #74, opened 2018-02-26
cli: nicer name for `-of`


tiborsimko:
The CLI option called `-of, --ouput-format` is composed of two names. This is not fully necessary. In order to ease the mnemotechnic recall, we can simplify the option to use one word only.  The options include:

1. using one option name with separate values, e.g. ``--format json`` or ``--format yaml``
2. using separate options ``--json`` and ``--yaml``

The second option seems preferable in case we don't need to export a plethora of diverse options. The two might suffice: (1) the text output (=the default human-friendly interaction channel) and (2) the json output (=the default machine-friendly output channel).  E.g. users could use tools like `json2yaml` and `yaml2json` if they need to convert it.

PR #82, opened 2018-03-02
cli: change in output_format flag

dinosk:
* Changes the `--output-format,  -of` flag to `--json`

(Closes https://github.com/reanahub/reana-client/issues/74)

Signed-off-by: Dinos Kousidis <konstantinos.kousidis@cern.ch>

### DIFF
--- a/docs/gettingstarted.rst
+++ b/docs/gettingstarted.rst
@@ -71,7 +71,7 @@ and check its status:
    $ reana-client workflow status --workflow 57c917c8-d979-481e-ae4c-8d8b9ffb2d10
    [INFO] REANA Server URL ($REANA_SERVER_URL) is: http://192.168.99.100:31201
    [INFO] Workflow "afbbf6d1-a129-4e4f-ab8a-b8df325351d2" selected
-   Name       |UUID                                |User                                |Organization|Status
+   name       |id                                  |user                                |organization|status
    -----------|------------------------------------|------------------------------------|------------|-------
    lucid_kirch|57c917c8-d979-481e-ae4c-8d8b9ffb2d10|00000000-0000-0000-0000-000000000000|default     |created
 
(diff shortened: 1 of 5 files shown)


dinosk:
@hjhsalo yes I agree with @tiborsimko in https://github.com/reanahub/reana-client/issues/74, json2yaml is an easy option. What do you think?

hjhsalo:
Well, I would have it since it isn't (or at least doesn't seem to be) affecting anything else than output, but 1) I don't have strong arguments why it should be kept, 2) I usually lose debates, so let's drop it :)

dinosk:
@hjhsalo :D

diegodelemos:
We have a miss match between what `reana-server` API returns and `reana-client * --json` returns. For instance, `id` in the API response and `UUID` written by `tablib` (as configured [here](https://github.com/reanahub/reana-client/pull/82/files#diff-19bed1e7334d00a24f059e517c55f01dR254)):

```json
$ diff -y <(curl -s "$REANA_SERVER_URL/api/analyses?user=00000000-0000-0000-0000-000000000000&organization=default") \
          <(reana-client workflow list --json | jq .)
[INFO] REANA Server URL ($REANA_SERVER_URL) is: http://192.168.99.100:30376
[                                                               [
  {                                                               {
    "id": "e68c59e2-c43e-408c-82fb-dcd32836b048",             |     "Name": "epic_hodgkin",
    "organization": "default",                                |     "UUID": "e68c59e2-c43e-408c-82fb-dcd32836b048",
    "status": "created",                                      |     "User": "00000000-0000-0000-0000-000000000000",
    "user": "00000000-0000-0000-0000-000000000000"            |     "Organization": "default",
                                                              >     "Status": "created"
  }                                                               }
]                                                               ]
```

dinosk:
@diegodelemos  Good point, thanks. 465ac43 should sync the outputs

PR #90, opened 2018-04-16
global: $REANA_WORKON -> REANA_WORKON

dinosk:


reana_client/cli/code.py
```diff
@@ -123,7 +123,7 @@ def code_list(ctx, user, organization, workflow, filter, output_format):
     '--workflow',
     default=os.environ.get('REANA_WORKON', None),
     help='Name or UUID of the workflow where the files should be uploaded to. '
-         'Overrides value of $REANA_WORKON.')
+         'Overrides value of `REANA_WORKON`.')
```
tiborsimko: For the help strings maybe write fully: "Overrides the value of REANA_WORKON environment variable".  Will be more understandable to users.

PR #91, opened 2018-04-16
cli: tab separated output

dinosk:

### DIFF
--- a/reana_client/cli/code.py
+++ b/reana_client/cli/code.py
@@ -29,9 +29,10 @@ import click
 
 import tablib
 
-from ..config import default_organization, default_user
 from ..errors import FileUploadError
 from ..api.client import UploadType
+from ..config import default_organization, default_user
+from ..utils import cli_printer
 
 
 @click.group(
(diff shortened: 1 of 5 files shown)


dinosk:
There is a TODO in this one, it prints out ugly if there are no data.

reana_client/cli/code.py
```diff
@@ -94,7 +96,7 @@ def code_list(ctx, user, organization, workflow, filter, output_format):
         if output_format:
             click.echo(data.export(output_format))
         else:
-            click.echo(data)
+            click.echo(data.export('df'))
```
tiborsimko: The `df` export requires `pandas` which may be an unnecessarily big dependency... Perhaps use `tsv`? It does not look as pretty as the docker output though :smile: Perhaps we may massage ourselves and/or find another helper library that would pretty-print table-like dictionaries.

reana_client/utils.py
```diff
@@ -209,3 +209,25 @@ def get_analysis_root():
                 analysis_root = parent_dir
     analysis_root += '/'
     return analysis_root
+
+
+def cli_printer(headers, _filter, data):
```
tiborsimko: We could call this `cli_table_printer()` to be more descriptive, but we can refactor when we move this to `reana-commons`...

PR #96, opened 2018-04-19
utils: import cli printer from reana-commons

dinosk:
This will require a release of reana-commons

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -35,6 +35,7 @@ history = open('CHANGES.rst').read()
 tests_require = [
     'check-manifest>=0.25',
     'coverage>=4.0',
+    'docutils==0.12',
     'httpretty>=0.8',
     'isort>=4.2.2',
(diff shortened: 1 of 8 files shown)


setup.py
```diff
@@ -66,7 +67,7 @@
 install_requires = [
     'bravado>=9.0.6',
     'click>=6.7',
-    'cwltool',
+    'cwltool==1.0.20180326152342',
```
diegodelemos: @dinosk why this? wasn't CWL test suite passing already?

dinosk: @diegodelemos requirements-builder requires a specified version

setup.py
```diff
@@ -35,6 +35,7 @@
 tests_require = [
     'check-manifest>=0.25',
     'coverage>=4.0',
+    'docutils==0.12',
```
dinosk: solves [docutils 0.13.1] assert not self.context

Issue #127, opened 2018-07-04
cli: richer users functionality


tiborsimko:
Currently, when one is an administrator, one can create users as follows:  

```$ reana-client users create -e <somebody@example.org> -t <admin-token>```

It would be useful to have:

- [x] a possibility to provide some wanted access token value for a user, e.g. if DB goes down.  

- [x] a possibility to list all users registered in the system for import/export purposes.

The former could be achieved by introducing:

```$ reana-client users create -e someemail -u wanteduseraccesstoken -a existingadmintoken```

The latter could be achieved by extending ``reana-client users get`` funcionality to make the inputs optional.  (Or just introduce new `list` verb similarly to what we have for `r-c workflow list`.)

diegodelemos:
Closed by https://github.com/reanahub/reana-server/pull/56.

PR #129, opened 2018-07-06
global: remove user management

dinosk:
Addresses #127 


dinosk:
After the discussion on API refactoring, the `users` cli group will be moved to `reana-server`, so this can be closed.

PR #133, opened 2018-07-12
global: renaming analyses -> workflows

dinosk:
Addresses https://github.com/reanahub/reana-server/issues/58


README.rst
```diff
@@ -24,17 +24,17 @@ About
 -----
 
 REANA-Client is a component of the `REANA <http://www.reana.io/>`_ reusable and
-reproducible research data analysis platform. It provides a command-line tool
```
tiborsimko: The word "analysis" should stay here and in other places that are using user-oriented language... You cannot "wide replace" everywhere.

dinosk: Right, though I didn't do a `replace all` actually as the word analysis here is just deleted.

PR #134, opened 2018-07-13
global: remove organization

dinosk:
Connects https://github.com/reanahub/reana-commons/issues/25

### DIFF
--- a/reana_client/config.py
+++ b/reana_client/config.py
@@ -34,9 +34,6 @@ reana_yaml_schema_file_path = pkg_resources.resource_filename(
 default_user = '00000000-0000-0000-0000-000000000000'
 """Default user to use when submitting workflows to Reana Server."""
 
-default_organization = 'default'
-"""Default organisation to use when submitting workflows to Reana Server."""
-
 ERROR_MESSAGES = {
         'missing_access_token':
         'Please provide your access token by using'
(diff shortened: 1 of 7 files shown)


reana_client/config.py
```diff
@@ -34,12 +34,12 @@
...
-"""Default organisation to use when submitting workflows to Reana Server."""
-
 ERROR_MESSAGES = {
         'missing_access_token':
         'Please provide your access token by using'
         ' the -at/--access-token flag, or by setting the'
         ' REANA_ACCESS_TOKEN environment variable.'
 }
+
+default_download_path = './outputs/'
```
diegodelemos: It is unused right now (removed in the outputs/inputs/code unification). However, I though that you plugged it again for safety (which actually makes sense) otherwise the file will be downloaded in current directory potentially overwriting important files. @dinosk WDYT about plugging it again? It is just that since we went to a totally plain workspace I did it like this.

reana_client/openapi_connections/reana_server.json
```diff
@@ -869,18 +801,8 @@
...
-              }
-            },
             "schema": {
-              "properties": {
-                "message": {
-                  "type": "string"
-                }
-              },
-              "type": "object"
+              "type": "file"
```
diegodelemos: `upload_file` doesn't return the file. To be changed in REANA Server

reana_client/config.py
dinosk: I saw that it gets downloaded to the current directory indeed, maybe we could add an optional parameter download_path, allowing the user to do:
```
reana-client files download plot.png outputs/myanalysis2
```

Either way, an issue should be opened about this and the fix be done in a separate PR, and I'll remove the config var here

PR #137, opened 2018-07-18
cli: better output message for status

### LINKED ISSUE #115: Nicer "Could not connect to the server." messages
Currently in case of troubles we have:

```console
$ reana-client ping
Could not connect to the server.
```
What we could show instead:

```console
$ reana-client ping
Could not connect to the selected REANA cluster server at http://reana-qa.cern.ch
```

This would be more informative.

Note also that we have in these cases:

```
$ reana-client status
User: 00000000-0000-0000-0000-000000000000
REANA cluster selected: http://reana-qa.cern.ch
Could not retrieve current status. Error: Extra data: line 1 column 5 - line 2 column 1 (char 4 - 19)
```

We could print nicer message.

dinosk:
Closes #115


dinosk:
This would print messages as:
```
User: 00000000-0000-0000-0000-000000000000
REANA cluster selected: http://192.168.99.100:31355
Workflow selected: workflow.1
Could not retrieve workflow status. Error: Token not valid.
```

```
User: 00000000-0000-0000-0000-000000000000
REANA cluster selected: http://192.168.99.100:31355
No workflow is selected currently.
```

diegodelemos:
Would it make sense to make the message shorter like:
`REANA cluster: http://192.168.99.100:31355`
Instead of:
`REANA cluster selected: http://192.168.99.100:31355 `
Same for `Workflow:` instead `Workflow selected:`.
WDYT?

dinosk:
I like `REANA cluster: ...`, I'm thinking that the `Workflow: workflow.1` line might read confusing.
Some ideas:
```
Active workflow: workflow.1
Current workflow: workflow.1
Working on: workflow.1
```
Maybe also to sync with this we can print the `REANA cluster` as:
```
Connected to: https://reana.cern.ch
```

diegodelemos:
+1 for `Working on:`. Regarding `Connected to:`, I like it but I am not sure if it may convey the correct message since I think that mentioning `REANA cluster: blah.blah` reminds that you are connected to a computing cluster whereas `Connected to:` might sound to abstract? WDYT? (:

dinosk:
Imo this is too much information, I think the message that you're connected to a URL which will probably have `reana` in it, is enough

reana_client/cli/status.py
```diff
@@ -23,45 +23,51 @@
...
+                click.echo(click.style(
+                    'Workflow status: {}'.
+                    format(workflow_status_response['status']),
+                    fg='green'))
+            except Exception as e:
+                click.echo(
+                    click.style('Could not retrieve workflow status.'
+                                ' Error: {}'.format(str(e)), fg='red'),
+                    err=True)
+                sys.exit(1)
```
diegodelemos: Probably we should add here a `finally` where we format the user message since if there is an exception thrown while calling the API the message is going to look like this:
```console
$ # Setting wrong REANA_SERVER_URL.
$ reana-client status
Could not retrieve current status. Error: HTTPConnectionPool(host='192.168.99.100', port=31758): Max retries exceeded with url: /api/ping (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x112b2a450>: Failed to establish a new connection: [Errno 61] Connection refused',))
```
Instead of something like (source #115):
```console
$ reana-client status
Connected to: http://192.168.99.100:31758
Working on: workflow.1
Could not retrieve current status. Error: HTTPConnectionPool(host='192.168.99.100', port=31758): Max retries exceeded with url: /api/ping (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x112b2a450>: Failed to establish a new connection: [Errno 61] Connection refused',))
```

dinosk: I refactored the messages, and now `status` returns:
```
1. # correct REANA_SERVER_URL, no workflow selected
Connected to: http://192.168.99.100:30362
No workflow is selected currently.

2. # correct REANA_SERVER_URL, workflow.1 selected
Connected to: http://192.168.99.100:30362
Working on: workflow.1
Workflow status: created

3. # correct REANA_SERVER_URL, non existing workflow selected
Connected to: http://192.168.99.100:30362
Working on: workflow.2
Could not retrieve workflow status. Error: REANA_WORKON is set to workflow.2, but that workflow does not exist. Please set your REANA_WORKON environment variable appropriately.

4. wrong REANA_SERVER_URL, regardless of workflow selected
Could not connect to the selected REANA cluster server at http://192.168.99.100:30363.
Please make sure you have set your REANA_SERVER_URL environment variable correctly.
```

Issue #125, opened 2018-06-29
cwl: nicer progress command


tiborsimko:
Prettify progress display for CWL workflows.  Currently one gets:

```console
$ reana-client workflow status -w workflow.2
NAME       RUN_NUMBER   CREATED               STATUS     PROGRESS   COMMAND

workflow 2 2018-06-28T08:57:32 finished 2/2 /bin/sh -c 'export
PATH="/usr/local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin";export
TMPDIR="/data/00000000-0000-0000-0000-000000000000/analyses/291323df-feb5-44ce-a36b-7067e40f80ff/workspace/cwl/tmpdir/_A33Un";export
HOME="/data/00000000-0000-0000-0000-000000000000/analyses/291323df-feb5-44ce-a36b-7067e40f80ff/workspace/cwl/docker_outdir";cp
-a
/data/00000000-0000-0000-0000-000000000000/analyses/291323df-feb5-44ce-a36b-7067e40f80ff/workspace/cwl/outdir/w_pMyc/data.root
/data/00000000-0000-0000-0000-000000000000/analyses/291323df-feb5-44ce-a36b-7067e40f80ff/workspace/cwl/docker_outdir/data.root
;cp -a
/data/00000000-0000-0000-0000-000000000000/analyses/291323df-feb5-44ce-a36b-7067e40f80ff/inputs/fitdata.C
/data/00000000-0000-0000-0000-000000000000/analyses/291323df-feb5-44ce-a36b-7067e40f80ff/workspace/cwl/docker_outdir/fitdata.C
;mkdir -p
/data/00000000-0000-0000-0000-000000000000/analyses/291323df-feb5-44ce-a36b-7067e40f80ff/workspace/cwl/docker_outdir
&& cd
/data/00000000-0000-0000-0000-000000000000/analyses/291323df-feb5-44ce-a36b-7067e40f80ff/workspace/cwl/docker_outdir
&& root -b -q
'"'"'fitdata.C("data.root","/data/00000000-0000-0000-0000-000000000000/analyses/291323df-feb5-44ce-a36b-7067e40f80ff/workspace/cwl/docker_outdir/plot.png")'"'"';
cp -r
/data/00000000-0000-0000-0000-000000000000/analyses/291323df-feb5-44ce-a36b-7067e40f80ff/workspace/cwl/docker_outdir/*
[...]

PR #141, opened 2018-07-27
cli: workflow status progress columns

dinosk:
* ADD step column to show current running step in workflow,
  and show Job.prettified_cmd in command column.

Connects https://github.com/reanahub/reana-client/issues/125

### DIFF
--- a/reana_client/cli/workflow.py
+++ b/reana_client/cli/workflow.py
@@ -308,6 +308,51 @@ def workflow_start(ctx, workflow, access_token):
 def workflow_status(ctx, workflow, _filter, output_format,
                     access_token, verbose):
     """Get status of previously created workflow."""
+    def _show_progress(succeeded_jobs, total_jobs):
+        if total_jobs:
+            return '{0}/{1}'.format(succeeded_jobs, total_jobs)
+        else:
+            return None
+
+    def _get_data_from_row(row, data, headers):
+        name, run_number = get_workflow_name_and_run_number(
+            row['name'])
+        total_jobs = row['progress'].get('total_jobs')
+        succeeded_jobs = row['progress'].get('succeeded')
+        if row['progress']['total_jobs'] > 0:
+            if 'progress' not in headers:
+                headers += ['progress']
+        current_command = ''
+        current_step_name = ''
+        if 'current_command' in row['progress'] and \
+                row['progress'].get('current_command'):
+            if 'command' not in headers:
+                headers += ['command']
+            current_command = row['progress'].\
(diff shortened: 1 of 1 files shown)


reana_client/cli/workflow.py
```diff
@@ -324,62 +364,15 @@ def workflow_status(ctx, workflow, _filter, output_format,
                                                           access_token)
             verbose_headers = ['id', 'user']
             headers = ['name', 'run_number', 'created',
-                       'status', 'progress', 'command']
+                       'status', 'progress']
```
diegodelemos: Shall we add progress only if the status response contains progress info? To handle the case in which  [progress is not still there](https://github.com/reanahub/reana-workflow-controller/blob/master/reana_workflow_controller/rest.py#L1100)?

dinosk: Note this won't raise an error but show 0/0 as the progress fields are initialized to 0 when a row gets added in `Job` table. I also think it shouldn't be shown in this case nonetheless.

reana_client/cli/workflow.py
```diff
@@ -308,6 +308,46 @@ def workflow_start(ctx, workflow, access_token):
...
+                headers.remove('command')
+        data.append(list(map(
+            str,
+            [name,
+             run_number,
+             row['created'],
+             row['status'],
+             '{0}/{1}'.
+             format(
+                 row['progress']['succeeded'],
```
diegodelemos: Can you reproduce?
```console
$ export REANA_WORKON=$(reana-client create -f reana-yadage.yaml)
$ reana-client status
Workflow status could not be retrieved:
'succeeded'
```

dinosk: yes, I'm working on it I forgot to set the label

reana_client/cli/workflow.py
```diff
@@ -308,6 +308,51 @@ def workflow_start(ctx, workflow, access_token):
 def workflow_status(ctx, workflow, _filter, output_format,
                     access_token, verbose):
     """Get status of previously created workflow."""
+    def _show_progress(succeeded_jobs, total_jobs):
+        if total_jobs:
+            return '{0}/{1}'.format(succeeded_jobs, total_jobs)
+        else:
+            return '-/-'
+
+    def _get_data_from_row(row, data, headers):
```
diegodelemos: Could we maybe remove `data` as parameter, and return a new list every time? So down, in `workflow_status` scope, when we call it we are always in control of what goes inside `data` making it easier to debug?

Issue #143, opened 2018-08-01
cli: start --no-cache


tiborsimko:
Now that `r-w-e-serial` has caching option, it would be useful to have the possibility to do:

```console
$ reana-client start --no-cache
```

to override engine's default "use cache" behaviour.

```console
$ kubectl logs serial-default-worker-77484d79cf-wh82f
...
[2018-07-31 16:54:54,179: WARNING/ForkPoolWorker-2] ~~~~~ Copied from cache
[2018-07-31 16:54:54,238: WARNING/ForkPoolWorker-2] ~~~~~ Copied from cache
```

Useful not only for users but also developers debugging components etc.

Issue #21, opened 2017-10-06
spec: add support for specifying some `reana run` operational arguments in spec file


hjhsalo:
Following the discussion in https://github.com/reanahub/reana-client/pull/16 it might be good to have a way to specify the "operational" argumens as CLI options (e.g. `reana run --user "jdoe" --organization "acme-dep"`), plus have a way of storing them in the spec files for later archiving.

What do you think, is this something that one would want to preserve in REANA specifications file or something that should definitely not be there?

Implementation would probably be something like:
* a key `user` in spec file where CLI reads it from.
* `user`-value passed to the CLI would take preference if it exists.
* Regarding to passing `user`-value as env_var (currently not supported), which one would take preference, the env_var or the cli-submitted one?
* If no `user`-value exists in spec file, nor one is passed to CLI and env_var doesn't specify it either, default value defined in config.py would be used.

diegodelemos:
Superseded by https://github.com/reanahub/reana-client/issues/152.

PR #151, opened 2018-08-08
cli: addition of parameters in workflow_start()

dinosk:
* ADD arbitrary number of parameters in workflow_start().

* Updates the reana-server openapi spec with parameters in
  set_workflow_status, and the status passed in the query instead of
  the request body.

Connects https://github.com/reanahub/reana-client/issues/21
Closes https://github.com/reanahub/reana-client/issues/143

### DIFF
--- a/reana_client/openapi_connections/reana_server.json
+++ b/reana_client/openapi_connections/reana_server.json
@@ -585,13 +585,10 @@
           },
           {
             "description": "Required. New workflow status.",
-            "in": "body",
+            "in": "query",
             "name": "status",
             "required": true,
-            "schema": {
-              "description": "Required. New status.",
-              "type": "string"
-            }
+            "type": "string"
           },
           {
             "description": "Required. The API access_token of workflow owner.",
@@ -599,6 +596,15 @@
             "name": "access_token",
             "required": true,
             "type": "string"
+          },
+          {
+            "description": "Optional. Extra parameters for workflow status.",
+            "in": "body",
+            "name": "parameters",
+            "required": false,
+            "schema": {
+              "type": "object"
+            }
           }
         ],
         "produces": [
--- a/reana_client/api/client.py
+++ b/reana_client/api/client.py
@@ -162,14 +162,15 @@ class Client(object):
         except Exception as e:
             raise e
 
-    def start_workflow(self, workflow, access_token):
+    def start_workflow(self, workflow, access_token, parameters):
         """Start a workflow."""
         try:
             (response,
              http_response) = self._client.api.set_workflow_status(
                  workflow_id_or_name=workflow,
                  status='start',
-                 access_token=access_token).result()
+                 access_token=access_token,
+                 parameters=parameters).result()
             if http_response.status_code == 200:
                 return response
             else:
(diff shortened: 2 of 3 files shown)


dinosk:
To test running a workflow without caching: `reana-client start CACHING=false`

reana_client/openapi_connections/reana_server.json
```diff
@@ -585,20 +585,36 @@
...
             "type": "string"
+          },
+          {
+            "description": "Optional. Extra parameters for workflow status.",
+            "in": "body",
+            "name": "parameters",
+            "required": false,
+            "schema": {
+              "properties": {
+                "input_parameters": {
```
diegodelemos: Is this schema being used? I just see a free dict here:
```console
    parsed_parameters = {'parameters':
                         dict(p.split('=') for p in parameters)}
```
If this is the desired schema, we have to think exactly where we want each parameter to end up and not mix responsibilities. As I see in [`RWES` PR](https://github.com/reanahub/reana-workflow-engine-serial/pull/32), this parameters are not just "parameters", they are "workflow_engine_parameters" as in `cwltool` could be the `--timestamps` flag (what I guess `run_parameters` means) but they are passed as [workflow parameters](https://github.com/reanahub/reana-workflow-controller/blob/master/docs/openapi.json#L220-L222).

If we want also the "workflow_parameters" (i.e. in demo [root6-roofit `events`](https://github.com/reanahub/reana-demo-root6-roofit/blob/master/workflow/cwl/input.yml#L1)) (what I guess `input_parameters` are) to be overwritten by this call, then we have to update the database since [they are written to the DB at creation time](https://github.com/reanahub/reana-workflow-controller/blob/master/reana_workflow_controller/rest.py#L295).

In any case they are two different things we should differentiate. To keep it simple, it the `PUT /api/workflows/{workflow_id_or_name}/status` I would for now, only pass "workflow_engine_parameters" and add a new parameter to the [Celery task](https://github.com/reanahub/reana-workflow-engine-serial/pull/32/files#diff-6d29ceb19895382845929513c9bc9099R86) and this one would be the one representing workflow engine parameters.

dinosk: The schema needs updating, regarding the differentiation I can see input parameters (number of events for example) and workflow run parameters (caching true|false), the schema was created with these 2 types, but I went with a single dictionary for now, to keep it short.
We should ticketize this splitting parameters into 2, but since now only the `CACHING=true|false` is read from them we can use it as is.

diegodelemos: Makes sense 👍 . So for now we implement this just with workflow engine parameters and we add this additional parameter in the task (`engine_parameters` or similar). Later we think about how we want to handle the `workflow_parameters` (I will create issue from this discussion).

I would emphasize in those two names or similar since: one, the `workflow_parameters` is as  [we have modeled it](https://github.com/reanahub/reana-db/blob/master/reana_db/models.py#L79-L90) (or `input_parameters` as it is common terminology in `yadage` and `CWL`), and second the workflow engine parameters (related to https://github.com/reanahub/reana/pull/73#discussion_r207264526 dicussion), since they are literally parameters for the concrete workflow engine implementation (i.e. [`cwltool` ](https://manpages.debian.org/testing/cwltool/cwltool.1.en.html) ones or [`yadage`](https://yadage.readthedocs.io/en/latest/executingworkflows.html#example) ones).

PR #154, opened 2018-08-09
cli: initialise API client at command level

### LINKED ISSUE #153: cli: reana-client validate fails when not server_url or token is set
https://travis-ci.org/diegodelemos/reana-demo-atlas-recast/builds/413629487#L582
Output from a Travis build on a REANA Demo:
```console
$ reana-client validate | grep -q ' is a valid '
[ERROR] REANA client is not connected to any REANA cluster.
Please set REANA_SERVER_URL environment variable to the remote REANA cluster you would like to connect to.
For example: export REANA_SERVER_URL=https://reana.cern.ch/
```
`reana-client validate` should work even if there is no `REANA_SERVER_URL` nor `REANA_ACCESS_TOKEN`.

diegodelemos:
* Introduces a decorator to be added to the commands that need
  an API client (closes #153).

### DIFF
--- a/reana_client/cli/files.py
+++ b/reana_client/cli/files.py
@@ -32,6 +32,7 @@ import tablib
 
 from ..config import ERROR_MESSAGES, default_user
 from ..errors import FileUploadError
+from reana_client.decorators import with_api_client
 from reana_commons.utils import click_table_printer
 
 
@@ -69,6 +70,7 @@ def files(ctx):
     default=os.environ.get('REANA_ACCESS_TOKEN', None),
     help='Access token of the current user.')
 @click.pass_context
+@with_api_client
 def get_files(ctx, workflow, _filter,
               output_format, access_token):
     """List workflow workspace files."""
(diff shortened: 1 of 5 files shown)


reana_client/cli/files.py
```diff
@@ -32,12 +32,14 @@
...
 from ..config import ERROR_MESSAGES, default_user
 from ..errors import FileUploadError
+from reana_client.decorators import with_api_client
 from reana_commons.utils import click_table_printer
 
 
 @click.group(
     help='All interaction related to files.')
 @click.pass_context
+@with_api_client
```
dinosk: I think we need to add `@with_api_client` in all commands in files.py

diegodelemos: Didn't test that, I thought that the object would be modified for next wrapped calls.. that makes it a bit inconvenient... I will add it everywhere, later we can come up with better solution. Thanks for spotting it!

diegodelemos:
Updated and rebased !

PR #160, opened 2018-08-28
validate: check if workflow parameters are expandable 

diegodelemos:
Connects https://github.com/reanahub/reana-workflow-engine-serial/issues/34.

Depends on https://github.com/reanahub/reana-commons/pull/40.


diegodelemos:
Done

PR #164, opened 2018-09-03
docs: document serial CACHING parameter

### LINKED ISSUE #161: parameters: document serial caching parameter
Currently caching parameter for Engine Serial is not documented see [here](https://reana-client.readthedocs.io/en/latest/cliapi.html#cmdoption-reana-client-start-arg-parameters).

diegodelemos:
* Closes #161.

### DIFF
--- a/reana_client/cli/workflow.py
+++ b/reana_client/cli/workflow.py
@@ -212,7 +212,14 @@ def workflow_create(ctx, file, name, skip_validation, access_token):
 
 @click.command(
     'start',
-    help='Start previously created workflow.')
+    help="""
+    Start previously created workflow.
+
+    The workflow execution can be further influenced by setting operational
+    prameters using `-p` or `--parameter` option.  The option can be
+    repetitive. For example, to disable caching for the Serial workflow
+    engine, you can set ``-p CACHING=false``.
+    """)
 @click.option(
     '-w',
     '--workflow',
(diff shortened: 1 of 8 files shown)


reana_client/cli/workflow.py
```diff
@@ -212,7 +212,17 @@ def workflow_create(ctx, file, name, skip_validation, access_token):
 
 @click.command(
     'start',
-    help='Start previously created workflow.')
+    help="""
+    Start previously created workflow
+
+    The workflow execution can be modified by setting operational
```
tiborsimko:
```
Start previously created workflow.

The workflow execution can be further influenced by setting operational
prameters using `-p` or `--parameter` option.  The option can be repetitive.
For example, to disable caching for the Serial workflow engine, you can 
set ``-p CACHING=false``. 
 ```

P.S. BTW the value `false` is not very commonly used in the CLI world, it would be better to use 0 and 1.

File: reana_client/cli/workflow.py [hunk 2]
```diff
@@ -225,13 +235,15 @@ def workflow_create(ctx, file, name, skip_validation, access_token):
     '--access-token',
     default=os.environ.get('REANA_ACCESS_TOKEN', None),
     help='Access token of the current user.')
-@click.argument(
-    'parameters',
-    nargs=-1,
+@click.option(
+    '-p', '--parameter',
+    multiple=True,
+    help='Operatinal parameters. For now only CACHING=[true|false] is '
```
tiborsimko: `s/Operatinal/Operational`

Instead of "for now only" you could say something short like:

```
Optional operational parameters for the workflow execution. E.g. CACHING=false.
```

PR #176, opened 2018-09-12
installation: workflow engine options

### LINKED ISSUE #174: pip install reana-client[cwl,serial,yadage]
As the number of supported REANA workflow engines grows, it is useful to offer an option to users to install only some workflow engine libraries that the given user uses. E.g. only Yadage and no CWL, which can come handy to both (1) save some disk space from dependencies and (2) ease possible incompatibility situations such as older CWL with Python 3.7.

The user would do: 

```console
$ pip install reana-client[yadage]
```

which would only install Yadage-related optional dependencies.

P.S. We should still consider which "popular" dependencies should be installed by default when people do not specify anything... Probably all three engines for now.

dinosk:
Closes #174

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -52,6 +52,16 @@ extras_require = {
         'sphinx-click>=1.0.4',
     ],
     'tests': tests_require,
+    'yadage': [
+        'pyOpenSSL==17.3.0',
+        'rfc3987==1.3.7',
+        'strict-rfc3339==0.7',
+        'webcolors==1.7',
+        'yadage-schemas==0.7.16',
+    ],
+    'cwl': [
+        'cwltool==1.0.20180326152342',
+    ],
 }
 
 extras_require['all'] = []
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -67,16 +77,12 @@
...
-    'pyOpenSSL==17.3.0',  # FIXME remove once yadage-schemas solves deps.
     'reana-commons>=0.3.1,<0.4',
-    'rfc3987==1.3.7',  # FIXME remove once yadage-schemas solves deps.
-    'strict-rfc3339==0.7',  # FIXME remove once yadage-schemas solves deps.
     'tablib>=0.12.1,<0.13',
-    'webcolors==1.7',  # FIXME remove once yadage-schemas solves deps.
-    'yadage-schemas==0.7.16',
 ]
 
+# install_requires += extras_require['yadage'] + extras_require['cwl']
```
tiborsimko: Should probably uncomment it for now to support all the workflow engines we have... Otherwise we would have to loudly specify `pip install reana-client[yadage,cwl]` everywhere

diegodelemos: Shall we uncomment this and rebase? Seems to be ready

diegodelemos:
This PR needs to come together with a tested change in REANA Makefile, adding the workflow engine installation options.

diegodelemos:
Closing as we've decided to go for a solution like https://appimage.org/ to package and distribute `reana-client`.

PR #183, opened 2018-09-25
release: v0.3.1

diegodelemos:


CHANGES.rst
```diff
@@ -1,6 +1,17 @@
 Changes
 =======
 
+Version 0.3.1 (2018-09-25)
+--------------------------
+
+- Introduces new feature for files upload and download, now these commands
```
tiborsimko: We can make it more user-orientated by putting the commands that the change concerns prominently in front, such as:

- Amends `upload` and `download` commands that will now upload/download all the files specified in `reana.yaml` in case no arguments are provided.
- Fixes `status` command's JSON output mode.
- Upgrades CWL reference implementation to version ``1.0.20180912090223``.
- Renames Serial workflow operational parameter from ``CACHING` to `CACHE`. 
- Adds support for Python 3.7.

Issue #106, opened 2018-06-14
cli: workflow status order


tiborsimko:
Currently we get:

```console
$ reana-client workflow status
NAME       RUN_NUMBER   ID                                     USER                                   ORGANIZATION   STATUS   COMMAND                                                                                             PROGRESS
workflow   15           2889f737-51a6-499a-9562-7e8a017d0e03   00000000-0000-0000-0000-000000000000   default        failed   source fix-env.sh && source env.sh && aliroot -b -q generate.C > generation.log 2> generation.err   4/8     
```

It would be good to display STATUS - PROGRESS - COMMAND  in this order, so that the table will be more easily readable.  (I.e. display fixed-width columns first,  and leave varchar-columns like command for later.)

See also #105 for musings about whether we need all the ID and USER and ORGANIZATION information being displayed by default... Better hide them and show them only when verbosity requires it.

dinosk:
PR #105 changes the `reana-client workflow status` as described here:
```
$ rc workflow status
NAME       RUN_NUMBER   STATUS     PROGRESS   COMMAND                                               
workflow   16           finished   2/2        root -b -q '../code/fitdata.C("data.root","plot.png")'
$ rc workflow status -v
NAME       RUN_NUMBER   STATUS     PROGRESS   COMMAND                                                  ID                                     USER                                   ORGANIZATION
workflow   16           finished   2/2        root -b -q '../code/fitdata.C("data.root","plot.png")'   1ea9c854-ecc5-40af-a2e5-7a284a6ecac8   00000000-0000-0000-0000-000000000000   default   
```
the extra columns are just appended.

PR #190, opened 2018-10-23
docs: virtualenv instead of mkvirtualenv

okraskaj:
(connects #106)

### DIFF
--- a/docs/gettingstarted.rst
+++ b/docs/gettingstarted.rst
@@ -12,7 +12,10 @@ environment:
 
 .. code-block:: console
 
-   $ mkvirtualenv reana-client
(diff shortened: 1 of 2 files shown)


docs/gettingstarted.rst
```diff
@@ -12,7 +12,9 @@ environment:
 
 .. code-block:: console
 
-   $ mkvirtualenv reana-client
+   $ # create new virtual environment
+   $ virtualenv ~/.virtualenvs/myreana
+   $ source ~/.virtualenvs/myreana/bin/activate
```
tiborsimko: Please add `$ # install reana-client` comment before the pip commands

PR #195, opened 2018-10-31
global: workflow deletion

### LINKED ISSUE #192: cli: workflow deletion
Should have option for soft/hard delete, delete all failed workflows.
Soft delete could be done through a new status `deleted`, hard delete could delete the row in the DB completely.

dinosk:
* ADDS cli and api command for deleting workflow runs,
  with option for deleting all runs of a given workflow.

Closes #192 

### DIFF
--- a/docs/userguide.rst
+++ b/docs/userguide.rst
@@ -154,6 +154,48 @@ analysis inputs, and start the workflow run.
    myanalysis   1            2018-11-07T12:45:18   running   1/1
    $ reana-client download results/plot.png
 
+Deleting workflows
+~~~~~~~~~~~~~~~~~~
+
+You can mark a workflow as deleted with:
+
+.. code-block:: console
+
+   $ reana-client delete
+
+Passing no argument will mark the workflow selected by your REANA_WORKON
+variable as deleted. To specify a different workflow than your
+currently selected one use the -w/--workflow flag and set the workflow name
+and run number.
+If e.g. workflow run number 123 of your analysis failed, you can delete it
+as follows:
+
+.. code-block:: console
+
+   $ reana-client delete --workflow=myanalysis.123
+
+After simple deletion the workspace can be accessed to retrieve files uploaded
+there. If you are sure the workspace can also be deleted pass the
+--include-workspace flag.
+
+.. code-block:: console
+
+   $ reana-client delete --workflow=myanalysis.123 --include-workspace
+
+To delete all runs of a given workflow, pass the --include-all-runs flag and
+run:
+
+.. code-block:: console
+
+   $ reana-client delete --workflow=myanalysis --include-all-runs
+
+and to completely remove a workflow run and its workspace from REANA
+pass the --include-records flag:
+
+.. code-block:: console
+
+   $ reana-client delete --workflow=myanalysis.1 --include-records
+
 Examples
 --------
 
--- a/reana_client/cli/workflow.py
+++ b/reana_client/cli/workflow.py
@@ -61,16 +61,22 @@ def workflow(ctx):
     flag_value='json',
     default=None,
     help='Get output in JSON format.')
-@add_access_token_options
+@click.option(
+    '--all',
+    'show_all',
+    count=True,
+    help='Show all workflows including deleted ones.'
+)
 @click.option(
     '-v',
     '--verbose',
     count=True,
     help='Set status information verbosity.')
+@add_access_token_options
 @click.pass_context
 @with_api_client
 def workflow_workflows(ctx, _filter, output_format, access_token,
-                       verbose):
+                       show_all, verbose):
     """List all workflows user has."""
     logging.debug('command: {}'.format(ctx.command_path.replace(" ", ".")))
     for p in ctx.params:
@@ -90,6 +96,8 @@ def workflow_workflows(ctx, _filter, output_format, access_token,
             headers += verbose_headers
         data = []
         for workflow in response:
+            if workflow['status'] == 'deleted' and not show_all:
+                continue
             name, run_number = get_workflow_name_and_run_number(
                 workflow['name'])
             data.append(list(map(str, [name,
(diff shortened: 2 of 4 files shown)


reana_client/cli/workflow.py
```diff
@@ -477,9 +477,71 @@ def workflow_validate(ctx, file):
...
+
+        except Exception as e:
+            logging.debug(traceback.format_exc())
+            logging.debug(str(e))
+            click.echo(
+                click.style('Workflow could not be started: \n{}'
+                            .format(str(e)), fg='red'),
+                err=True)
+    else:
+        click.secho('Workflow name and number must be provided'
```
roksys: In case $REANA_WORKON var is not set and --workflow flag is not given, error message is printed twice. Do we want that?

`callback=workflow_uuid_or_name`

```
(reana-client3) rokas:reana-client rokas$ reana-client delete
Workflow name must be provided either with `--workflow` option or with REANA_WORKON environment variable
Workflow name and number must be provided either with `--workflow` option or with REANA_WORKON environment variable
```

dinosk: No, this is happening for all other cli commands in `workflow.py` and it should be fixed with the next commit. This is still wip, but as it seems wafflebot puts it `in review` after every push

docs/userguide.rst
```diff
@@ -115,6 +115,36 @@ file of the analysis, by just running:
...
+
+   $ reana-client delete --workflow=otherworkflow.1
+
+Deleting a workflow removes its workspace from REANA, making any
+files present there inaccessible.
+To delete all runs of a given workflow, pass the --all flag and run:
+
+.. code-block:: console
+
+   $ reana-client delete --workflow=failedworkflow.1 --all
```
tiborsimko: In this case there would be no ".1".

dinosk: Yes, currently this works with both

reana_client/cli/workflow.py
```diff
@@ -477,9 +464,65 @@ def workflow_validate(ctx, file):
...
 
 
+@click.command(
+    'delete',
+    help='Delete a workflow run.')
+@click.option(
+    '-A',
+    '--all',
+    'all_runs',
+    count=True,
```
tiborsimko: Hmm, the option `--all` may be perhaps misunderstood... Perhaps we could name it more explicitly (e.g. `--all-runs`) or support some wildcard thing such as `-w workflow.*` or `-w workflow.[0-9]*`?

The most frequent functionality that people may want to do and need to understand is (1) deleting runs regarding DB and logs and other such accounting vs (2) deleting runs regarding workspace files and disk space.  It may be easy to misunderstand what "soft" and "hard" do here.  BTW this is also what brought me to muse about `--all`, because the switch could be understood to act on deleting all files and all DB entries and all logs and everything.) 

It may be useful good to think it terms the following thinking framework: *(just an example)* 

command | hide DB entries in CLI? | delete DB entries? | delete logs? | delete workspace? | delete cache?
--- | --- | --- | --- | --- | ---
"soft" | yes  |  no | no | no | yes 
"hard" | no | yes | yes | yes | yes 

A frequent wish may be to remove the files (=to save space) but to keep an entry in the DB (=to see which parameters failed for past runs).  This wish would not be really easily expressible via soff/hard delete dichotomy anymore... It may be more advantageous to offer a CLI like:

```console
$ reana-client delete -w workflow.17 --erase-files --erase-logs
```

with some sane defaults. 

In this way users can compose what they wish to keep and what they wish to delete.  (And `--all` or `--erase-all` would do everything for the given run...)

We can muse further IRL.

reana_client/cli/workflow.py
```diff
@@ -539,10 +533,71 @@ def workflow_run(ctx, file, filenames, name, skip_validation,
...
+@click.pass_context
+@with_api_client
+def workflow_delete(ctx, workflow, all_runs, hard_delete, access_token):
+    """Delete a workflow run given the workflow name and run number."""
+    logging.debug('command: {}'.format(ctx.command_path.replace(" ", ".")))
+    for p in ctx.params:
+        logging.debug('{param}: {value}'.format(param=p, value=ctx.params[p]))
+
+    if not access_token:
+        click.echo(
```
okraskaj: Why not click.secho?

dinosk: `click.echo(click.style())` you'll see is in a lot of places in the codebase, we can change it everywhere in one go if we think this is important

docs/userguide.rst
```diff
@@ -154,6 +154,36 @@ analysis inputs, and start the workflow run.
...
+
+and to completely remove a workflow from REANA pass the --hard flag:
+
```
diegodelemos: Missing `.. code-block:: console`

docs/userguide.rst
tiborsimko: Better than calling workflow runs "otherworkflow" and "failedworkflow", we can write a more realistic example as follows:

- use `-w myanalysis` always
- run number 122 was good
- run number 123 failed so we want to delete it

Please also use "workflow run" in the verbose text to make it more clear that people will be deleting failed runs, not the abstract concept of workflow as such.  Example:  "If the workflow run number 123 of your analysis failed, you can delete it as follows: ...  Deleting a workflow run will do ... You can delete all runs using ... "

reana_client/cli/workflow.py
```diff
@@ -576,10 +570,89 @@ def workflow_run(ctx, file, filenames, name, skip_validation,
...
+    count=True,
+    help='Delete all runs of a given workflow.')
+@click.option(
+    '--files',
+    'workspace',
+    count=True,
+    help='Delete workflow workspace from REANA.')
+@click.option(
+    '-a',
+    '--all',
```
tiborsimko: The `--all` option looks

reana_client/cli/workflow.py
```diff
@@ -576,10 +570,89 @@ def workflow_run(ctx, file, filenames, name, skip_validation,
...
+         'the workflow list.\n'
+         'Workspaces of deleted workflows'
+         'are accessible to retrieve files, to remove the workspace'
+         ' too pass --files flag.\n'
+         'By passing -r/--every-run all workflows with the same'
+         ' will be deleted.\n'
+         'The -a/--all flag will hard delete'
+         ' the workflow data from the database and remove its workspace.')
+@click.option(
+    '-r',
```
tiborsimko: For dangerous options, we should not make a one-letter shortcut.  Similarly to how `dpkg` uses only long forms for `--force-*` commands.

Since we are in the context of deleting workflows, the option to delete all runs could be simply:

```console
$ reana-client delete -w myanalysis --all-runs
```

I would not use any one-letter shortcut for this dangerous command...

PR #204, opened 2018-11-09
cli: workflow diff

dinosk:
* Adds diff command returning the differences between two workflows.
  Includes optional flags to show short diff or full extensive diff,
  including differences in the file contents between the two
  workflow workspaces.

Closes https://github.com/reanahub/reana-client/issues/193

### DIFF
--- a/reana_client/api/client.py
+++ b/reana_client/api/client.py
@@ -465,7 +465,7 @@ class Client(BaseAPIClient):
             parameters = {'force_stop': force_stop}
             (response, http_response) = self._client.api.set_workflow_status(
                 workflow_id_or_name=workflow,
-                status='stopped',
+                status='stop',
                 access_token=access_token,
                 parameters=parameters).result()
             if http_response.status_code == 200:


reana_client/api/client.py
```diff
@@ -485,3 +485,45 @@ def stop_workflow(self, workflow, force_stop, access_token):
...
+
+    def diff_workflows(self, workflow_id_a, workflow_id_b,
+                       brief, access_token, context_lines):
+        """Return the list of differences between two workflows.
+
+        :param workflow_id_a: UUID which identifies the first workflow.
+        :param workflow_id_b: UUID which identifies the second workflow.
+        :param brief: Flag specifying desired detail in diff.
+        :param access_token: API token of user requesting diff.
+        :returns: A list of dictionaries composed by `asset`, `type`, `lines,
```
diegodelemos: Missing closing back quote in `lines`.

reana_client/api/client.py
```diff
@@ -485,3 +485,45 @@ def stop_workflow(self, workflow, force_stop, access_token):
...
+            `a` and `b`. Asset refers to the workflow asset where a
+            difference was found, type refers to the asset type, lines refer
+            to the lines of the file where the differences are and a, b fields
+            are the actual lines that differ.
```
diegodelemos: Missing `context_lines` parameter docstring.

reana_client/cli/workflow.py
```diff
@@ -702,6 +702,89 @@ def workflow_delete(ctx, workflow, all_runs, workspace,
                 err=True)
 
 
+@click.command(
+    'diff',
+    help='Show differences between two workflows.')
+@click.option(
+    '-a',
+    '--workflow-a',
```
diegodelemos: Should we make the workflows (`workflow_a` and `workflow_b`) arguments instead of options? I think it should would be clearer, like:
```diff
- reana-client diff -a myanalysis.34 -b myotheranalysis.1
+ reana-client difff myanalysis.34 myotheranalysis.1
```

run-tests.sh
```diff
@@ -7,7 +7,7 @@
 # under the terms of the MIT License; see LICENSE file for more details.
 
 pydocstyle reana_client && \
-isort -rc -c -df **/*.py && \
+#isort -rc -c -df **/*.py && \
```
diegodelemos: We are seeing this problem everywhere, we should fix it and propagate the change to the rest of problematic repos. @okraskaj did you fix it already? If so please share with @dinosk.

reana_client/api/client.py
dinosk: thanks

diegodelemos:
Works well except for the `--brief` mode. For me, it still shows the detailed view, can you confirm?

dinosk:
It works only for the `workspace_listing` part. `difflib` doesn't provide it for the `reana_specification`

Issue #194, opened 2018-10-31
cli: workflow stop


dinosk:

tiborsimko:
Use case scenario: a user submits some long-running workflow and realises a mistake.  It is wanted to `stop` it soon in order to liberate resources. The stop could be:

- (i) `--soft`, i.e. terminate the steps you are working on and do not proceed further; in this case the workflow run will end up in a consistent state;

- (ii) `--hard`, i.e. kill any running jobs the sooner the bettter; in this case the workflow may end up in a totally broken state.  

In the former use case, the user may be interesting in submitting corrected files and `continue`-ing the execution from some given step.  In the latter use case, the user may want to `delete` the workflow run and its workspace.

We can distinguish the above two possibilities via CLI options, even though we may implement only the latter option in this sprint.

PR #205, opened 2018-11-14
cli: add stop command

okraskaj:

Connects #194.

### DIFF
--- a/docs/userguide.rst
+++ b/docs/userguide.rst
@@ -196,6 +196,25 @@ pass the --include-records flag:
 
    $ reana-client delete --workflow=myanalysis.1 --include-records
 
+Stopping workflows
+~~~~~~~~~~~~~~~~~~
+
+You can stop a workflow with:
+
+.. code-block:: console
+
+    $ reana-client stop --force
+
+The workflow assigned to REANA_WORKON variable will be stopped. To specify a
+different workflow than your currently selected one use the -w/--workflow flag
+and set the workflow name or UUID.
+
+.. code-block:: console
+
+    $ reana-client stop --force --workflow=otherworkflow.1
+
+Note that currently only force stop is implemented.
+
 Examples
 --------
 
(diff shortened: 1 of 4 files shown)


docs/userguide.rst
```diff
@@ -196,6 +196,25 @@ pass the --include-records flag:
...
+Passing no argument will stop the workflow selected by your REANA_WORKON
+variable. To specify a different workflow than your
+currently selected one use the -w/--workflow flag and set the workflow name
+or UUID.
+
+.. code-block:: console
+
+    $ reana-client stop --workflow=otherworkflow.1
+
+
```
dinosk: We need to mention here the `--force` flag

reana_client/cli/workflow.py
```diff
@@ -513,6 +513,53 @@ def workflow_validate(ctx, file):
...
+    callback=workflow_uuid_or_name,
+    help='Name and run number to be stopped. '
+         'Overrides value of REANA_WORKON.')
+@add_access_token_options
+@click.pass_context
+@with_api_client
+def workflow_stop(ctx, workflow, force_stop, access_token):
+    """Stop given workflow."""
+    if not force_stop:
+        click.secho('Graceful stop not implement yet. If you really want to '
```
dinosk: Works fine like this, we could consider showing a confirmation message and having `--force` set by default

docs/userguide.rst
```diff
@@ -196,6 +196,26 @@ pass the --include-records flag:
...
+Stopping workflows
+~~~~~~~~~~~~~~~~~~
+
+You can stop a workflow with:
+
+.. code-block:: console
+
+    $ reana-client stop --force
+
+Passing no argument will stop the workflow selected by your REANA_WORKON
```
roksys: I would rephrase this sentence to something like - 

> Workflow assigned to REANA_WORKON variable will be stoped. 

What do you think?

PR #242, opened 2019-01-29
cli: change message on start depending on status

### LINKED ISSUE #235: cli: "... has been started"
Now that workflows are queued for execution and not always started, the output message may be misleading.  

Example:

```console
[INFO] Starting workflow...
reana-demo-worldpopulation.6 has been started.
$ reana-client status -w reana-demo-worldpopulation.6                                                
NAME                         RUN_NUMBER   CREATED               STATUS   PROGRESS
reana-demo-worldpopulation   6            2019-01-28T13:17:35   queued   -/-     
```

Instead of "... has been started", it would be nicer to say something like "... has been queued for execution. There are currently 10 jobs in queue."

diegodelemos:
* (closes #235).


reana_client/utils.py
```diff
@@ -24,6 +25,15 @@
 
 from reana_client.config import reana_yaml_schema_file_path
 
+workflow_status_verbs_mapping = {
+    'created': 'has been',
+    'running': 'is',
+    'finished': 'has',
+    'failed': 'has',
+    'deleted': 'is',
```
tiborsimko: Could also be "has been deleted" and "has been stopped".

PR #243, opened 2019-01-30
cli: adding column filtering

### LINKED ISSUE #236: cli: workflows --filter
Provide nice examples of usage in `workflows --filter` documentation.

Currently one gets not much `workflows --help` there, and the output  like: 

```console
$ reana-client workflows --filter created
CREATED                             
myanalysis                          
workflow_name                       
reana-demo-worldpopulation          
reana-demo-alice-lego-train-test-run
reana-demo-helloworld               
reana-demo-bsm-search               
reana-demo-atlas-recast             
reana-demo-root6-roofit             
myanalysis                          
workflow                            
reana-demo-atlas-recast             
reana-demo-worldpopulation          
reana-demo-bsm-search               
reana-demo-alice-lego-train-test-run
myanalysis                          
workflow                            
reana-demo-worldpopulation          
reana-demo-atlas-recast             
recast                              
reana-demo-bsm-search               
reana-demo-bsm-search               
reana-demo-atlas-recast             
reana-demo-worldpopulation          
reana-demo-bsm-search               
reana-demo-worldpopulation          
reana-demo-bsm-search               
reana-demo-worldpopulation          
reana-demo-bsm-search               
```

don't seem that useful :wink: since there is no creation information returned there. 

E.g. one could expect usage scenario going like:

```console
$ reana-client workflows --filter status=queued 
```

roksys:
* Allows to filter workflows list by column name and value.
  Closes #236


roksys:
What do you think about renaming this flag to `--format` instead of `--filter`?
E.g. in condor
`$ condor_q`
> ...
> -format <fmt> <attr>	 Print attribute attr using format fmt
> ...

Slurm also uses `-format`

> -O <output_format>, --Format=<output_format>
> Specify the information to be displayed. Also see the -o <output_format>, --format=<output_format> option described below (which supports greater flexibility in formatting, but does not support access to all fields because we ran out of letters). Requests a comma separated list of job information to be displayed.

https://slurm.schedmd.com/squeue.html

reana_client/cli/utils.py
```diff
@@ -22,3 +24,42 @@ def add_access_token_options(func):
     def wrapper(*args, **kwargs):
         return func(*args, **kwargs)
     return wrapper
+
+
+def parse_parameters(filter):
```
diegodelemos: Can you change this variable name here and elsewhere to `filter_` or a different name to prevent shadowing the built-in function?

diegodelemos:
Regarding the naming, almost all tools I am using everyday use the filter option, but is true that our users are used to other ones such as these schedulers so in my eyes, it would make sense. One thing which even though you don't mention it, it appears in the slurm help, what I wouldn't do is to use the `-O` or `-o` as short since we use this already to change the document format to be retrieved (json, yaml, ASCII table) and we could combine the data filtering with it i.e. `reana-client workflows -o json --filter status=failed`.

PR #244, opened 2019-02-04
cli: workspace disk usage

### LINKED ISSUE #238: cli: info about disk space usage
In the `reana-client` commands such as `workflows`, it would be good to list the workdir disk space occupied by this workflow run, kind of like `docker images` does.  

Example:

```console
$ reana-client workflows
NAME                   RUN_NUMBER   CREATED               STATUS        SIZE
myanalysis             1            2019-01-25T14:46:31   running       1.0GB
workflow               1            2019-01-25T16:22:49   running       200MB
```

People would more easily know which runs are to be deleted in case they approach their total quota, etc.

dinosk:
* Adds command to display workspace disk usage.
  Displays workspace tree with disk usage for each element by default,
  '-s/--summarize' parameter shows only total sum.
  Values shown in human readable format.

Closes #238 

### DIFF
--- a/reana_client/cli/workflow.py
+++ b/reana_client/cli/workflow.py
@@ -23,6 +23,7 @@ from reana_client.api.client import (create_workflow, current_rs_api_client,
                                      delete_workflow, diff_workflows,
                                      get_workflow_logs,
                                      get_workflow_parameters,
+                                     get_workflow_disk_usage,
                                      get_workflow_status, get_workflows,
                                      open_interactive_session, start_workflow,
                                      stop_workflow)
@@ -100,8 +101,8 @@ def workflow_workflows(ctx, _filter, output_format, access_token,
     if _filter:
         parsed_filters = parse_parameters(_filter)
     try:
-        response = get_workflows(access_token)
-        verbose_headers = ['id', 'user']
+        response = get_workflows(access_token, bool(verbose))
+        verbose_headers = ['id', 'user', 'size']
         headers = ['name', 'run_number', 'created', 'status']
         if verbose:
             headers += verbose_headers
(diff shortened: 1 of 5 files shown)


reana_client/cli/workflow.py
```diff
@@ -455,6 +454,61 @@ def add_verbose_data_from_response(response, verbose_headers,
...
+            parameters = {'summarize': summarize}
+            response = get_workflow_disk_usage(workflow,
+                                               parameters,
+                                               access_token)
+            headers = ['size', 'name']
+            data = []
+            if summarize:
+                click_table_printer(headers, [],
+                                    [response['disk_usage_info'][-1]['size'],
+                                     response['disk_usage_info'][-1]['name']])
```
tiborsimko: Works nicely; one remark on the output:

```console
$ reana-client du -w helloworld.serial          
SIZE   NAME                  
4.0K   /results/greetings.txt
4.0K   /results              
4.0K   /data/names.txt       
4.0K   /data                 
4.0K   /code/helloworld.py   
4.0K   /code                 
12K                          
```

It would be good to display `./results/greetings.txt' with the leading `./` so that directories wouldn't appear to be absolute but rather relative paths to the workflow workspace.

tiborsimko:
@dinosk Please squash together, the commit message of the second commit is not correct.  Please also increase `reana_client/version.py` at the same time.  It would be good to keep increasing YYYYMMDD version for each `r-client` commit, now that we have users using unstable-not-yet-released-on-PyPI versions, for easier troubleshooting.

dinosk:
@tiborsimko Squashed and fixed. There is one more PR added in r-commons for fixing the `--summarize` flag, which will bump the version of setup.py here and in r-server.

PR #245, opened 2019-02-06
cli: add open interactive processes command

diegodelemos:
* (connect reanahub/reana-workflow-controller#178).

### DIFF
--- a/reana_client/cli/workflow.py
+++ b/reana_client/cli/workflow.py
@@ -24,13 +24,14 @@ from reana_client.api.client import (create_workflow, current_rs_api_client,
                                      get_workflow_logs,
                                      get_workflow_parameters,
                                      get_workflow_status, get_workflows,
-                                     start_workflow, stop_workflow)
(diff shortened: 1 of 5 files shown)


reana_client/cli/workflow.py
```diff
@@ -820,6 +821,52 @@ def print_color_diff(lines):
...
+@click.option(
+    '-i',
+    '--image',
+    help="Image to use in the interactive session, by default it is"
+         "``jupyter/scipy-notebook``.")
+@click.option(
+    '-p',
+    '--port',
+    type=int,
+    help="Port exposed by the service run inside the container spawned"
```
roksys: space missing. 

>   -p, --port INTEGER        Port exposed by the service run inside the
>                             container spawnedwith ``image``, by default

Maybe it is a bit too techincal help description for user. What about something like this?

> Specify a port for the interactive session.

What do you think?

diegodelemos: Definitely 👍 changing ...

PR #246, opened 2019-02-07
cli: adding mv command

### LINKED ISSUE #213: cli: ls rm cp mv
**Current status**

When a user wants to transfer files from client to cluster and vice versa, the user can do:

- `reana-client upload` to transfer files from client to cluster's workspace 

- `reana-client download` to transfer files from cluster's workspace to client

This ticket is not about these commands, but rather about operating files inside the workspace directly, without need to transfer anything.  This is useful for speed.

So, when a user wants to work with files directly in the workspace, there are currently only a few commands for now:

- `reana-client list` is listing files in the current workspace.

- `reana-client remove` is being added in order to remove files from the workspace. (#210)

**Use case**

For interactive development, people can frequently amend file names or change directory structure, so it would be useful to have an easy way to "copy" files around, to "move" and "rename" them, etc.  This will save the time so that people won't have to do "upload/download" for big data files, for example.

**Proposal**

Let us introduce a Unix like syntax for workspace-file-oriented commands and let us enrich the command set to to include copying and moving:

- `reana-client ls '*.C'` -- list files in workspace
- `reana-client rm mytmp/myfiltereddata.res` -- remove files in workspace
- `reana-client cp data/mytestdata.csv data/myinputdata.csv` -- copy files inside workspace 
- `reana-client mv data/mydata.csv data/mymeasurement/mydata.csv` -- move files inside workspace

Possibly also: *(unless the above commands already create any necessary directories and subdirectories by themselves, as we do already)*

- `reana-client mkdir data/mymeasurement/` -- create directories inside workspace
- `reana-client rmdir mytmp/` -- remove directories inside workspace

**Note**

- The `ls/rm/cp/mv` convention for commands would be always indicating operating *inside* the workspace only. Hence `cp` would operate directly there, not to copy files from the client to the cluster. For transferring flies between the client and the cluster, we have "upload" and "download" commands.  We should remember to document this properly in the `reana-client` user guide.   

- Having `mv` would be much more useful than `cp` for users. We could even leave `cp` out for clarity for now. 

- An alternative would be to open a restricted shell session on the workspace and let people run some restricted set of commmands, but that can of worms had better stay closed for now...

roksys:
* Closes #213

### DIFF
--- a/reana_client/cli/files.py
+++ b/reana_client/cli/files.py
@@ -15,17 +15,17 @@ import traceback
 
 import click
 import tablib
+from reana_commons.errors import MissingAPIClientConfiguration
+from reana_commons.utils import click_table_printer
 
 from reana_client.api.client import (current_rs_api_client, delete_file,
-                                     download_file, list_files,
-                                     upload_to_server)
-from reana_client.cli.utils import (add_access_token_options, parse_parameters,
-                                    filter_data)
+                                     download_file, get_workflow_status,
+                                     list_files, mv_files, upload_to_server)
+from reana_client.cli.utils import (add_access_token_options, filter_data,
+                                    parse_parameters)
 from reana_client.config import ERROR_MESSAGES
 from reana_client.errors import FileDeletionError, FileUploadError
 from reana_client.utils import get_workflow_root, load_reana_spec
-from reana_commons.errors import MissingAPIClientConfiguration
-from reana_commons.utils import click_table_printer
 
 
 @click.group(
@@ -37,7 +37,7 @@ def files(ctx):
 
 
 @click.command(
-    'list',
+    'ls',
     help='List workflow workspace files.')
 @click.option(
     '-w',
@@ -319,7 +319,7 @@ def upload_files(ctx, workflow, filenames, access_token):
 
 
 @click.command(
-    'remove',
+    'rm',
     help='Delete the specified file or pattern.')
 @click.argument(
     'filenames',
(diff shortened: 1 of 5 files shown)


reana_client/cli/files.py
```diff
@@ -388,7 +388,68 @@ def delete_files(ctx, workflow, filenames, access_token):
...
+                                                 access_token).get('status')
+            if current_status == 'running':
+                click.echo(
+                    click.style('File(s) could not be moved for running '
+                                'workflow', fg='red'),
+                    err=True)
+                sys.exit(1)
+            files = list_files(workflow, access_token)
+            current_files = [file['name'] for file in files]
+            for item in current_files:
```
dinosk: I think this condition should be: `if source not in current_files`
or even just do the check directly in RWC to not have 2 api calls
```
(reana-3.6) ➜  reana-demo-root6-roofit git:(master) ✗ rc ls 
NAME             SIZE   LAST-MODIFIED      
code/fitdata.C   1648   2019-02-12T15:40:32
code/gendata.C   1937   2019-02-12T15:40:31
(reana-3.6) ➜  reana-demo-root6-roofit git:(master) ✗ rc mv code/fitdata.C code/fitdata2.C
Target file(s) code/fitdata2.C does not exist in workspace ['code/fitdata.C', 'code/gendata.C']
```

dinosk: if this is changed it works nicely :+1:

roksys: I had condition `if source not in current_files` before but, then movement of dirs does not work.

roksys: Yes, you are right, but then `mv` won't work for directories. I am thinking how to allow it.

roksys: I added more logic, still not perfect, but allow movement of dirs.

PR #254, opened 2019-02-19
installation: reana-commons with kubernetes

dinosk:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -54,7 +54,7 @@ install_requires = [
     'click>=7,<8',
     'cwltool==1.0.20181118133959',
     'pyOpenSSL==17.5.0',  # FIXME remove once yadage-schemas solves deps.
-    'reana-commons>=0.5.0.dev20190215,<0.6.0',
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -54,7 +54,7 @@
     'click>=7,<8',
     'cwltool==1.0.20181118133959',
     'pyOpenSSL==17.5.0',  # FIXME remove once yadage-schemas solves deps.
-    'reana-commons>=0.5.0.dev20190215,<0.6.0',
+    'reana-commons[kubernetes]>=0.5.0.dev20190220,<0.6.0',
```
tiborsimko: Do we need `kubernetes` here for the end users?

Please also upgrade package version to `0.5.0.dev20190220` so that we have the same version as in `r-commons`.

dinosk: We need it just because there's an import from `reana_commons.utils` and we import `kubernetes.client.rest.ApiException` there. If that import is moved inside the util function that may raise it we can remove it from here.

PR #257, opened 2019-02-27
interactive-execution: use closed session types

diegodelemos:
Closes https://github.com/reanahub/reana-client/issues/258.

### DIFF
--- a/reana_client/api/client.py
+++ b/reana_client/api/client.py
@@ -629,30 +629,28 @@ def diff_workflows(workflow_id_a, workflow_id_b,
         raise e
 
 
-def open_interactive_session(workflow, access_token, image=None, port=None):
+def open_interactive_session(workflow, access_token,
+                             interactive_session_type,
+                             interactive_session_configuration):
     """Open an interactive session inside the workflow workspace.
 
     :param workflow: Workflow which workspace will be available inside the
         interactive session.
     :param access_token: Workflow owner REANA access token.
-    :param image: Image to use in the interactive session, by default it is
-        ``jupyter/scipy-notebook``.
-    :param port: Port exposed by the service run inside the container spawned
-        with ``image``, by default ``8888`` for Jupyter notebooks.
+    :param interactive_session_type: Type of interactive session to spawn.
+    :param interactive_session_configuration: Specific configuration for
+        the interactive session.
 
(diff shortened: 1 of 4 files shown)


reana_client/api/client.py
```diff
@@ -629,30 +629,23 @@ def diff_workflows(workflow_id_a, workflow_id_b,
...
     """Open an interactive session inside the workflow workspace.
 
     :param workflow: Workflow which workspace will be available inside the
         interactive session.
     :param access_token: Workflow owner REANA access token.
-    :param image: Image to use in the interactive session, by default it is
-        ``jupyter/scipy-notebook``.
-    :param port: Port exposed by the service run inside the container spawned
-        with ``image``, by default ``8888`` for Jupyter notebooks.
+    :param interactive_session_type: Type of interactive session to spawn.
```
tiborsimko: Going for type `jupyter` is good, but it'll bd still interesting to offer an option to users to provide a customised image that will match the `jupyter` type protocol... E.g. there are [lots of stacks](https://jupyter-docker-stacks.readthedocs.io/en/latest/using/selecting.html) that people may want to choose from, including a wish to specify their own version. Kind of like:

```console
$ reana-client open jupyter     # detault, e.g. jupyter/scipy-notebook:latest
$ reana-client open jupyter --image jupyter/tensorflow-notebook:2343e33dec46
```

reana_client/cli/files.py
```diff
@@ -42,7 +43,7 @@ def files(ctx):
 @click.option(
     '-w',
     '--workflow',
-    default=os.environ.get('REANA_WORKON', None),
```
roksys: Does this line need to be deleted?

reana_client/cli/workflow.py
```diff
@@ -873,25 +874,26 @@ def print_color_diff(lines):
...
     '--image',
-    help="Image to use in the interactive session, by default it is "
-         "``jupyter/scipy-notebook``.")
-@click.option(
-    '-p',
-    '--port',
-    type=int,
-    help="Specify a port for the interactive session, by default ``8888`` "
-         "for Jupyter notebooks.")
+    help='Docker image which will be used to spawn the interactive session, '
```
roksys: I think it should be dot (.) at the end of the sentence, right?

PR #259, opened 2019-03-04
json_schema: adding cvmfs repo validation

### LINKED ISSUE #161: parameters: document serial caching parameter

### LINKED ISSUE #256: validation: cvmfs
**Observed behaviour**

 Using the following `reana.yaml` test case:

```yaml
workflow:
  type: serial
  resources:
    - cvmfs:
      - fcc.cern.ch
  specification:
    steps:
      - environment: 'cern/slc6-base'
        commands:
        - ls -l /cvmfs/ > output.txt
```

the CVMFS support is working perfectly fine :+1:

However, in case the `resources` part is not correctly set by the user, such as:

```yaml
...
 resources:
    - /cvmfs/fcc.cern.ch
...
```

then the validation succeeds (`reana-client validate` returns OK) but the workflow is not starting:

```console
$ reana-client status -w workflow.8        
NAME       RUN_NUMBER   CREATED               STATUS    PROGRESS
workflow   8            2019-02-20T16:34:01   running   -/-     
```

and there are no pods being created on the cluster side.

**Desired behaviour**

- The proper status should be reported, not "running", when the REANA cluster cannot start some workflows due to whatever internal reasons.
- On the cluster side, it would be good to log somewhere these cases when a workflow cannot be started, so that admins are alerted. (Hunting exceptions in server and controller pod logs is not easy, and I see mostly  `200` statuses there now...) For example `/var/reana/server.log` or somesuch, until the times we have larger logging and monitoring infrastructure.
- On the client side, it would be good to improve the validation so that the user would be informed about invalid `resources` declaration and would not be even able to submit such a badly formatted workflow.

roksys:
* Closes #256 

### DIFF
--- a/reana_client/schemas/reana_analysis_schema.json
+++ b/reana_client/schemas/reana_analysis_schema.json
@@ -136,8 +136,19 @@
         },
         "resources": {
           "id": "/properties/workflow/properties/resources",
-          "type": "array",
-          "title": "Workflow resources in yaml format."
+          "type": "object",
+          "title": "Workflow resources in yaml format.",
+          "properties": {
+            "cvmfs": {
+              "id": "/properties/workflow/properties/resources/properties/cvmfs",
+              "type": "array",
+              "items": {
+                "type": "string",
+                "title": "CVMFS repos",
+                "optional": false
+              }
+            }
+          }
         }
       },
       "required": ["type"]


diegodelemos:
Needs rebase :)

PR #278, opened 2019-04-12
cli: workflow option as decorator

roksys:
* Adding decorator function for passing `workflow` option
  from CLI.

### DIFF
--- a/reana_client/cli/utils.py
+++ b/reana_client/cli/utils.py
@@ -14,12 +14,27 @@ import shlex
 
 import click
 
+from reana_client.utils import workflow_uuid_or_name
+
 
 def add_access_token_options(func):
     """Add access token related options to click commands."""
-    @click.option('-at', '--access-token',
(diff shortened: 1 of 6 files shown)


reana_client/cli/utils.py
```diff
@@ -14,10 +14,12 @@
 
 import click
 
+from reana_client.utils import workflow_uuid_or_name
+
 
 def add_access_token_options(func):
     """Add access token related options to click commands."""
-    @click.option('-at', '--access-token',
+    @click.option('-t', '--access-token',
```
diegodelemos: Thanks for remembering this, we didn't ticketise it!

reana_client/cli/utils.py
```diff
@@ -26,6 +28,19 @@ def wrapper(*args, **kwargs):
     return wrapper
 
 
+def add_workflow_option(func):
```
diegodelemos: 🎊  thanks also for the refactor!

PR #285, opened 2019-05-06
cli: implement a user secrets management mechanism

### LINKED ISSUE #283: cli: implement a user secrets management mechanism
In order to implement [Condor-Slurm-Demo](https://github.com/orgs/reanahub/projects/3) we will need to allow users to upload their secrets, so REANA can instantiate jobs in their name. The CLI could look like follows:

```console
$ reana secrets add <secret-name> <secret-value>
$ reana secrets list
$ reana secrets delete <secret-name>
```

Note that this feature will help as well, later, to enable push/pull mechanisms, i.e. user wants the results of the workflow to be pushed to his EOS space, so REANA will be able to do it thanks to the secret store.

okraskaj:
* Closes #283

### DIFF
--- a/reana_client/api/client.py
+++ b/reana_client/api/client.py
@@ -14,13 +14,14 @@ import traceback
 from functools import partial
 
 from bravado.exception import HTTPError
-from werkzeug.local import LocalProxy
-
 from reana_client.config import ERROR_MESSAGES, WORKFLOW_ENGINES
 from reana_client.errors import FileDeletionError, FileUploadError
 from reana_client.utils import (_validate_reana_yaml, get_workflow_root,
                                 is_uuid_v4)
 from reana_commons.api_client import get_current_api_client
+from reana_commons.errors import REANASecretAlreadyExists, \
+    REANASecretDoesNotExist
+from werkzeug.local import LocalProxy
 
 current_rs_api_client = LocalProxy(
     partial(get_current_api_client, component='reana-server'))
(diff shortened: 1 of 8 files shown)


reana_client/cli/secrets.py
```diff
@@ -0,0 +1,196 @@
...
+    \t                            --from-file ~/.keytab
+    """
+    if not access_token:
+        click.echo(
+            click.style(ERROR_MESSAGES['missing_access_token'], fg='red'),
+            err=True)
+        sys.exit(1)
+
+    secrets = {}
+    for literal in from_literal:
```
diegodelemos: What do you think about extracting all this code to parse secrets from CLI to a function? It could be a [`click.ParamType`](https://click.palletsprojects.com/en/7.x/api/#click.ParamType) with the logic inside [`convert`](https://click.palletsprojects.com/en/7.x/api/#click.ParamType.convert).

reana_client/cli/secrets.py
```diff
@@ -0,0 +1,196 @@
...
+    """Add secrets from literal string or from file.
+
+    Examples: \n
+    \t $ reana-client secrets-add --from-literal PASSWORD=password \n
+    \t $ reana-client secrets-add --from-file ~/.keytab \n
+    \t $ reana-client secrets-add --from-literal USER=reanauser \n
+    \t                            --from-literal PASSWORD=password \n
+    \t                            --from-file ~/.keytab
+    """
+    if not access_token:
```
diegodelemos: We are repeating this check everywhere, I have created an issue to solve it later https://github.com/reanahub/reana-client/issues/286 :).

reana_client/cli/secrets.py
```diff
@@ -0,0 +1,196 @@
...
+                ', '.joint(secrets.keys())),
+                fg='green')
+        )
+
+
+@secrets_group.command()
+@add_access_token_options
+@click.argument('secrets', type=str, nargs=-1)
+def secrets_delete(secrets, access_token):  # noqa: D301
+    """Delete users secrets with a specified names.
```
diegodelemos:
```suggestion
    """Delete user secrets by name.
```

reana_client/cli/secrets.py
```diff
@@ -0,0 +1,196 @@
...
+        sys.exit(1)
+
+    if not access_token:
+        click.echo(
+            click.style(ERROR_MESSAGES['missing_access_token'], fg='red'),
+            err=True)
+        sys.exit(1)
+
+    try:
+        response = list_secrets(access_token)
```
diegodelemos: I would change the naming here and everywhere else we use `response` when we call the `api_client` and rather use a more descriptive name such as `secrets_list`, since the `list_secrets(access_token)` method returns an actual list and not an "HTTP response", what do you think?

tests/test_cli_secrets.py
```diff
@@ -0,0 +1,146 @@
...
+"""REANA client secrets tests."""
+
+from click.testing import CliRunner
+from mock import Mock, patch
+from pytest_reana.test_utils import make_mock_api_client
+
+from reana_client.cli import cli
+
+
+def test_secrets_list_server_not_reachable():
```
diegodelemos: I have created an issue, we should test this only once when we have centralised the check, to do later https://github.com/reanahub/reana-client/issues/287.

tests/test_cli_secrets.py
```diff
@@ -0,0 +1,146 @@
...
+    """Test list secrets when not connected to any cluster."""
+    message = 'REANA client is not connected to any REANA cluster.'
+    runner = CliRunner()
+    result = runner.invoke(cli, ['secrets-list'])
+    assert result.exit_code == 1
+    assert message in result.output
+
+
+def test_secrets_list_server_no_token():
```
diegodelemos: We can remove this test once we address #286.

reana_client/api/client.py
```diff
@@ -764,3 +763,100 @@ def get_workflow_disk_usage(workflow, parameters, access_token):
...
+            access_token=access_token).result()
+        if http_response.status_code in [200, 404]:
+            return response
+        else:
+            raise Exception(
+                "Expected status code 200 but replied with "
+                "{status_code}".format(
+                    status_code=http_response.status_code))
+
+    except HTTPError as e:
```
diegodelemos: The `404` will be an HTTPError, so we need to handle the `404` case in the exception body.

reana_client/cli/secrets.py
```diff
@@ -63,49 +65,22 @@ def secrets_add(from_literal, from_file, overwrite, access_token):  # noqa: D301
 
     secrets = {}
```
diegodelemos: Let's call this in a different way, such as `secrets_` to avoid shadowing `secrets` library.

reana_client/cli/secrets.py
```diff
@@ -0,0 +1,181 @@
...
+        secret = parse_secret_from_literal(literal)
+        secrets.update(secret)
+    for path in from_file:
+        secret = parse_secret_from_path(path)
+        secrets.update(secret)
+    try:
+        add_secrets(secrets, overwrite, access_token)
+    except REANASecretAlreadyExists as e:
+        logging.debug(traceback.format_exc())
+        logging.debug(str(e))
```
diegodelemos:
```suggestion
        logging.debug(str(e), exc_info=True)
```

We can use this keyword parameter so we do not have to explicitly log the `traceback` :)

PR #299, opened 2019-06-25
cli: add flag to list output URLs

### LINKED ISSUE #297: cli: add option to list workflow output files as URL's
As an enrichment of the `ls` command, we could offer users the possibility to list the files as URLs:

```console
$ reana-client ls --url
https://reana.cern.ch/api/workflows/{workflow_id_or_name}/workspace/{file_name}
https://reana.cern.ch/api/workflows/{workflow_id_or_name}/workspace/{file_name}
https://reana.cern.ch/api/workflows/{workflow_id_or_name}/workspace/{file_name}
```

The same output could be added to `reana-client run --follow` output, showing the list of files in case of success.

leticiawanderley:
Add new flag to the list workspace files command.

Allow user to retrieve files' URLs

Closes #297

Signed-off-by: leticia <leticia.farias.wanderley@cern.ch>

### DIFF
--- a/reana_client/cli/files.py
+++ b/reana_client/cli/files.py
@@ -20,10 +20,11 @@ from reana_client.api.client import (current_rs_api_client, delete_file,
                                      download_file, get_workflow_disk_usage,
                                      get_workflow_status, list_files, mv_files,
                                      upload_to_server)
+from reana_client.api.utils import get_path_from_operation_id
 from reana_client.cli.utils import (add_access_token_options,
                                     add_workflow_option, filter_data,
                                     parse_parameters)
-from reana_client.config import ERROR_MESSAGES
+from reana_client.config import ERROR_MESSAGES, JSON, URL
 from reana_client.errors import FileDeletionError, FileUploadError
 from reana_client.utils import (get_workflow_root, load_reana_spec,
                                 workflow_uuid_or_name)
@@ -54,6 +55,12 @@ def files_group(ctx):
     flag_value='json',
     default=None,
     help='Get output in JSON format.')
+@click.option(
+    '--url',
+    'output_format',
+    flag_value='url',
+    default=None,
+    help='Get URLs of output files.')
 @add_access_token_options
 @click.pass_context
 def get_files(ctx, workflow, _filter,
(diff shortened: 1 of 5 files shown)


reana_client/cli/files.py
```diff
@@ -67,10 +73,12 @@ def get_files(ctx, workflow, _filter,
     Examples: \n
     \t $ reana-client ls --workflow myanalysis.42
     """
+    JSON = 'json'
+    URL = 'url'
+    FILE_PATH = "https://reana.cern.ch/api/workflows/%s/workspace/%s"
```
leticiawanderley: Is this `reana.cern.ch` the only possible address? Or should we use the `current_rs_api_client.swagger_spec.api_url` value?

diegodelemos: To get the URL you can use [`ctx.obj`](https://github.com/diegodelemos/reana-client/blob/master/reana_client/cli/__init__.py#L87) which contains the configuration with the `REANA_SERVER_URL` set by the user: [`ctx.obj. reana_server_url`](https://github.com/diegodelemos/reana-client/blob/master/reana_client/cli/__init__.py#L29) :)

reana_client/cli/files.py
```diff
@@ -67,10 +73,12 @@ def get_files(ctx, workflow, _filter,
     Examples: \n
     \t $ reana-client ls --workflow myanalysis.42
     """
+    JSON = 'json'
+    URL = 'url'
```
diegodelemos: We might need more output formats later, what do you think about moving it to config?

reana_client/cli/files.py
```diff
@@ -67,10 +73,12 @@ def get_files(ctx, workflow, _filter,
...
+    FILE_PATH = "%s/api/workflows/%s/workspace/%s"
```
diegodelemos: If you would be able to retrieve [this string](https://github.com/reanahub/reana-commons/blob/master/reana_commons/openapi_specifications/reana_server.json#L1835) from the `api_client` you could do as follows (and we would be save from having this string outdated in the future 😄):

```python
In[1]: "/api/workflows/{workflow_id_or_name}/workspace/{file_name}".format(workflow_id_or_name='1234', file_name='test.json')                                                     
Out[2]: '/api/workflows/1234/workspace/test.json'
```

PR #308, opened 2019-07-26
cli: create -w

tiborsimko:
* Adds `-w` option as an alias for `-n` to the `create` command
  so that users can use `-w` consistently everywhere to specify
  the desired workflow.  (Principle of least astonishment.)

### DIFF
--- a/reana_client/cli/workflow.py
+++ b/reana_client/cli/workflow.py
@@ -17,7 +17,6 @@ import traceback
 import click
 import tablib
 from jsonschema.exceptions import ValidationError
-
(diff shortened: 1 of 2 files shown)


tiborsimko:
Note: the Travis CI build troubles on Python-2.7 stem from ``typing-extensions 3.7.4 has requirement typing>=3.7.4, but you have typing 3.6.6``. These dependencies are brought by ``cwltool``.

tiborsimko:
Workaround for Python-2.7 deps added. Tested locally full Python-2 based installation as well.

PR #311, opened 2019-08-09
cli: secrets cli refactor

### LINKED ISSUE #288: refactor: in secrets cli extract "file or literal" code into a custom ParamType
In secrets_add command argument ``--from-file`` and ``--from-literal`` could be extracted to custom [ParamType](https://click.palletsprojects.com/en/7.x/api/#click.ParamType). It would allow to require at least one of them.

### LINKED ISSUE #294: cli: adding non file objects fails with cryptic error
Submitting something which is not a file using `--from-file` fails with non user-friendly error:
```console
$ reana-client secrets-add --from-file PASSWORD=123456
File PASSWORD=123456 could not be uploaded: PASSWORD=123456 does not exist.
Traceback (most recent call last):
  File "/Users/rodrigdi/.virtualenvs/reana/bin/reana-client", line 10, in <module>
    sys.exit(cli())
  File "/Users/rodrigdi/.virtualenvs/reana/lib/python3.6/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/Users/rodrigdi/.virtualenvs/reana/lib/python3.6/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/Users/rodrigdi/.virtualenvs/reana/lib/python3.6/site-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/rodrigdi/.virtualenvs/reana/lib/python3.6/site-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/rodrigdi/.virtualenvs/reana/lib/python3.6/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/Users/rodrigdi/.virtualenvs/reana/lib/python3.6/site-packages/reana_client/cli/utils.py", line 27, in wrapper
    return func(*args, **kwargs)
  File "/Users/rodrigdi/.virtualenvs/reana/lib/python3.6/site-packages/reana_client/cli/secrets.py", line 72, in secrets_add
    secrets_.update(secret)
TypeError: 'NoneType' object is not iterable
```

### LINKED ISSUE #305: cli: secrets-add full path upload
When uploading secrets from file, the full path does not seem to be accepted.

This works OK:

```
$ reana-client secrets-add --from-file .keytab 
```

This does not:

```
$ reana-client secrets-add --from-file ~/.keytab 
```

Even though we advertise it like this in the docs :wink: 

We should fix the path resolution.

### LINKED ISSUE #306: cli: secrets-add --from-file
**Problem**

We are using the following syntax to upload files as secrets:

```
$ reana-client secrets-add --from-file mytoken.txt
```

I think the word "from this is very misleading. Kubernetes uses `--from-file` differently; the secret values are stored a file and are *read from* that file when they are being added. 

In our case, we are adding full file as secret directly, so it would be more correct to say `secrets-add --file mytoken.txt`.  We are not reading values *from* that file.

**Proposal**

(1) If we don't implement Kubernetes's "from file" functionality, then we can use a simplified API like:

```
$ reana-client secrets-add --env MYPARAM=myvalue
$ reana-client secrets-add --file mytoken.txt
```

Simple and clean for lambda researchers who don't speak Kubernetes secrets languages.

(2) If we implement "from file" functionality, then something this should be possible:

```
$ reana-client secrets-add --env --from-literal MYPARAM=myvalue
$ echo MYPARAM=myvalue > mysecrets.txt
$ reana-client secrets-add --env --from-file mysecrets.txt
```

We would still not be fully K8s compliant in this manner, but at least the "from" functionality would be consistent, and people could use this in their scripts without revealing the values.

Note that we would still have to have something like:

```
$ reana-client secrets-add --file mytoken.txt
```

for the file upload.

(3) We may switch later to using Vault or some such for the better secret management, so emulating Kubernetes syntax too literally in our `reana-client` API is not advantageous anyway. We can take some liberties to make syntax more user-friendly.

WDYT?

okraskaj:
* Change syntax for file and env types (Closes #306)

* Introduce NotRequiredIf functionality for click options (Closes #288)

* Change type of  argument to click.Path for native path support (Closes #294)

* Limit k8s secret name to just basename- it used to fail
because of forbiden characters (Closes #305)

### DIFF
--- a/reana_client/cli/utils.py
+++ b/reana_client/cli/utils.py
@@ -11,6 +11,7 @@ import functools
 import json
 import os
 import shlex
+import sys
 
 import click
 
@@ -85,3 +86,29 @@ def format_session_uri(reana_server_url, path, access_token):
     return "{reana_server_url}{path}?token={access_token}".format(
         reana_server_url=reana_server_url,
         path=path, access_token=access_token)
+
+
+class NotRequiredIf(click.Option):
+    """Allow only one of two arguments to be missing."""
+
+    def __init__(self, *args, **kwargs):
+        """."""
+        self.not_required_if = kwargs.pop('not_required_if')
+        assert self.not_required_if, "'not_required_if' parameter required"
+        super(NotRequiredIf, self).__init__(*args, **kwargs)
+
+    def handle_parse_result(self, ctx, opts, args):
+        """Overwritten click method."""
+        argument_present = self.name in opts
+        other_argument_present = self.not_required_if in opts
+        if not argument_present and not other_argument_present:
+            click.echo(
+                click.style(
+                    'At least one of the options: `{}` or `{}` is required\n'
+                    .format(self.name, self.not_required_if),
+                    fg='red') + ctx.get_help(),
+                err=True)
+            sys.exit(1)
+
+        return super(NotRequiredIf, self).handle_parse_result(
+            ctx, opts, args)
--- a/docs/userguide.rst
+++ b/docs/userguide.rst
@@ -141,28 +141,28 @@ the ``secrets-add`` command:
 .. code-block:: console
 
    $ # You can upload secrets from literal strings:
-   $ reana-client secrets-add --from-literal HTCONDORCERN_USERNAME=johndoe
-                              --from-literal HTCONDORCERN_KEYTAB=.keytab
+   $ reana-client secrets-add --env HTCONDORCERN_USERNAME=johndoe
+                              --env HTCONDORCERN_KEYTAB=.keytab
    Secrets HTCONDORCERN_USERNAME, HTCONDORCERN_KEYTAB were successfully uploaded.
 
    $ # ...and from files:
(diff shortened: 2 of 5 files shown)


reana_client/cli/utils.py
```diff
@@ -85,3 +86,30 @@ def format_session_uri(reana_server_url, path, access_token):
...
+
+    def handle_parse_result(self, ctx, opts, args):
+        """Overwritten click method."""
+        argument_present = self.name in opts
+        other_argument_present = self.not_required_if in opts
+
+        if not argument_present and not other_argument_present:
+            click.echo(
+                click.style(
+                    'At least one of the options: `%s` or `%s` is required\n'
```
diegodelemos: Can you switch to `format`? :)

PR #327, opened 2019-11-27
cli: updates callback function signature

### LINKED ISSUE #326: warning: use latest click validation callback signature
After merging https://github.com/reanahub/reana-client/pull/325 a warning message is showing up, we should use the new signature for Click validation callbacks, more [here](https://click.palletsprojects.com/en/7.x/options/#callbacks-and-eager-options).

```console
$ reana-client list
/Users/rodrigdi/.virtualenvs/reana/lib/python3.7/site-packages/click/core.py:940: Warning: Invoked legacy parameter callback "<function access_token_check at 0x11048f680>".  The new signature for such callbacks starting with click 2.0 is (ctx, param, value).
  value, args = param.handle_parse_result(ctx, opts, args)
...
```

roksys:
* Closes #326

### DIFF
--- a/reana_client/cli/utils.py
+++ b/reana_client/cli/utils.py
@@ -31,7 +31,7 @@ def add_access_token_options(func):
     return wrapper
 
 
-def access_token_check(ctx, access_token):
+def access_token_check(ctx, _, access_token):
     """Check if access token is present."""
     if not access_token:
         click.echo(


reana_client/cli/utils.py
```diff
@@ -31,7 +31,7 @@ def wrapper(*args, **kwargs):
     return wrapper
 
 
-def access_token_check(ctx, access_token):
+def access_token_check(ctx, param, access_token):
```
diegodelemos: I think that using `param` as a parameter name could be confusing along with `access_token`. What do you think of either renaming to `def access_token_check(ctx, param, value):` (so it is clear that one is the param name and the other the value, more generic) or ignore the `param` parameter `def access_token_check(ctx, _, access_token):` (more specific)?

diegodelemos:
The closes statement is missing the `#`.

PR #331, opened 2019-12-03
docs: reanmes HTCONDORCERN* vars to CERN*

roksys:
* Addresses reanahub/reana-job-controller/issues/71

### DIFF
--- a/docs/userguide.rst
+++ b/docs/userguide.rst
@@ -141,36 +141,36 @@ the ``secrets-add`` command:
 .. code-block:: console
 
    $ # You can upload secrets from literal strings:
-   $ reana-client secrets-add --env HTCONDORCERN_USERNAME=johndoe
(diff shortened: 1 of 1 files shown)


docs/userguide.rst
```diff
@@ -185,8 +185,8 @@ the ``secrets-list`` command:
    $ reana-client secrets-list
    NAME                    TYPE
    .keytab                 file
-   HTCONDORCERN_KEYTAB     env
-   HTCONDORCERN_USERNAME   env
+   CERN_KEYTAB             env
+   CERN_USER               env
```
tiborsimko: Please add a section about Slurm in the user documentation, and mention that for both Slurm and HTCondor the workspace is transferred, and kept after the workflow finishes, so that people should clean it themselves...

PR #332, opened 2019-12-09
cli: refactors connection check

roksys:
* No need to check connection for every command (validate, version)

### DIFF
--- a/reana_client/cli/files.py
+++ b/reana_client/cli/files.py
@@ -22,8 +22,8 @@ from reana_client.api.client import (current_rs_api_client, delete_file,
                                      upload_to_server)
 from reana_client.api.utils import get_path_from_operation_id
 from reana_client.cli.utils import (add_access_token_options,
-                                    add_workflow_option, filter_data,
-                                    parse_parameters)
+                                    add_workflow_option, check_connection,
+                                    filter_data, parse_parameters)
 from reana_client.config import ERROR_MESSAGES, JSON, URL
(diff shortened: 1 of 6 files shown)


reana_client/cli/files.py
```diff
@@ -40,6 +40,7 @@ def files_group(ctx):
 
 @files_group.command('ls')
 @add_workflow_option
+@check_connection
```
diegodelemos: Can we pass the `@check_connection` decorator to the parent Click group `files_group `? All file commands need the check so we write it fewer times.

Same happens to `@add_access_token_options` which we are writing several times. Since this one is not related to the PR you can leave, but if it is quick for you to test it would be better to move it together with `@check_connection` to the Click group.

roksys: > Can we pass the @check_connection decorator to the parent Click group files_group ? All file commands need the check so we write it fewer times.

It was the initial idea, unfortunately it does not work.

reana_client/cli/secrets.py
```diff
@@ -53,6 +54,7 @@ def secrets_group(ctx):
     default=False,
     help='Overwrite the secret if already present')
 @add_access_token_options
+@check_connection
```
diegodelemos: Same as before, moving decorators to `secrets_group`? This applies to all commands that belong to a Click group and all of them they need the check.

diegodelemos:
Before merging I just did a quick grep and it seems that `workflow diff` is missing the check, can you add it there? :)

PR #337, opened 2019-12-20
docs: adds serial partial execution example

roksys:


docs/userguide.rst
```diff
@@ -241,6 +241,37 @@ analysis inputs, and start the workflow run.
...
+
+   $ reana-client start -w workflow.1 -o target='gendata'
+   # or
+   $ reana-client run -w workflow.1 -o target='gendata'
+
+*CWL*
+
+CWL allows to execute only certain step of the workflow specified by the user.
+To do so, you need to provide target step name as operational option for the
+``reana-client start`` or ``reana-client run`` commands.
```
diegodelemos: What do you think of pointing to CWL docs here? Strictly speaking, the `target='gendata'` means something different for serial than for CWL, in serial runs until `genedata` for CWL it runs from `gendata`.

> If a target is an output parameter, it will only run only the steps that contribute to that output. If a target is a workflow step, it will run the workflow starting from that step. If a target is an input parameter, it will only run only the steps that are connected to that input.

Extracted from https://github.com/common-workflow-language/cwltool#running-only-part-of-a-workflow.

docs/userguide.rst
```diff
@@ -241,6 +241,37 @@ analysis inputs, and start the workflow run.
...
+
+.. code-block:: console
+
+   $ reana-client start -w workflow.1 -o target='gendata'
+   # or
+   $ reana-client run -w workflow.1 -o target='gendata'
+
+*CWL*
+
+`CWL allows to execute parts <https://github.com/common-workflow-language/cwltool#running-only-part-of-a-workflow>`_
```
diegodelemos: What about:
```console
`CWL allows executing workflows partially <https://github.com/common-workflow-language/cwltool#running-only-part-of-a-workflow>`_.
To do so, you need to provide the specific target as an operational option for the ``reana-client start``
or ``reana-client run`` commands.
```

PR #344, opened 2020-01-07
cli: sets default notebook to jupyter

roksys:


docs/userguide.rst
```diff
@@ -303,7 +303,7 @@ environment spawned on the remote containerised platform.
 
 .. code-block:: console
 
-   $ reana-client open -w myanalysis.1 jupyter
+   $ reana-client open -w myanalysis.1
```
tiborsimko: Hmm, I'm not sure about removing "jupyter", since "open -w myanalysis.42" can give an impression that this will open the web page for the given analysis, which we don't have yet but which will come.  It'll be good to keep the option open to use "open" without qualifier to point to the future `reana.cern.ch/myanalysis/42` page with workflow details.

This is akin to "hub" command-line client where "hub browse" opens the GitHub repository web page; similarly "reana-client open" can in the future open the REANA web page for the given workflow run.)

Perhaps revert back to having "jupyter" mandatory in order not to loose this possibility?

reana_client/cli/workflow.py
```diff
@@ -868,6 +868,7 @@ def interactive_group():
 @click.argument(
     'interactive-session-type',
     metavar='interactive-session-type',
+    default=INTERACTIVE_SESSION_TYPES[0],
```
tiborsimko: Since `close` closes jupyter, but default, I guess we can go with it, but there might be changes ahead later.

PR #351, opened 2020-01-14
cli: hide .git related files

mvidalgarcia:
fixes reanahub/reana#240

### DIFF
--- a/reana_client/cli/files.py
+++ b/reana_client/cli/files.py
@@ -31,6 +31,9 @@ from reana_client.utils import (get_workflow_root, load_reana_spec,
 from reana_commons.utils import click_table_printer
 
 
+FILES_BLACKLIST = ('.git/', '/.git/')
+
+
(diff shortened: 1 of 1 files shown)


reana_client/cli/files.py
```diff
@@ -31,6 +31,9 @@
 from reana_commons.utils import click_table_printer
 
 
+FILES_BLACKLIST = ('.git/', '/.git/')
```
roksys: Is there a need for `'/.git/'` ?

mvidalgarcia: Yes,  this is how `reana-client du` prints the information. This is what `reana-client` obtains from the server:

```python
...
{'name': '/.git/info/exclude', 'size': '4.0K'}
{'name': '/.git/info', 'size': '4.0K'}        
{'name': '/.git/branches', 'size': '0'}       
{'name': '/.git', 'size': '180K'}             
...
```

Issue #355, opened 2020-01-17
cli: rerun workflow in the same workspace


diegodelemos:
We shall implement https://github.com/reanahub/reana-client/issues/202#issuecomment-441049790 which is the first approach of `reana-client workflow continue` command but in a more simple way, allowing `reana-client run` to be called multiple times instead of creating a new command. In order to tackle this we should take into account:

(1) Currently, we have embedded logic determining whether a workflow can be change to running or stopped in [two functions `_start_workflow` and `_stop_workflow`](https://github.com/reanahub/reana-workflow-controller/blob/01c6402d4bf76341a05b8bc5f59b04d5e10037cd/reana_workflow_controller/rest.py#L1824-L1861). In the case of rerunning, we would need to allow the transition `[finished|failed] -> [running]`, however in our codebase we only allow a workflow to be run if it is in state `created` or `queued`, see the following snippet or the code [here](https://github.com/reanahub/reana-workflow-controller/blob/01c6402d4bf76341a05b8bc5f59b04d5e10037cd/reana_workflow_controller/rest.py#L1826):

```python
...
    if workflow.status in [WorkflowStatus.created, WorkflowStatus.queued]:
        # start workflow code
...
```

To address the new rerun functionality, we can either:
- Patch it there, in [`_start_workflow`](https://github.com/reanahub/reana-workflow-controller/blob/01c6402d4bf76341a05b8bc5f59b04d5e10037cd/reana_workflow_controller/rest.py#L1824)
- Do more deep work and solve the existing ticket ["state: use state machine to manage workflow state"](https://github.com/reanahub/reana-workflow-controller/issues/149) and then, on top, add the new rule.

(2) Until now we have only covered RWC part. Regarding the specific workflow engines, regardless of the approach (1) or (2), we will face different challenges depending on the workflow engine implementation, for example, if you try to run yadage twice you get:

```console
$ yadage-run . workflow/yadage/workflow.yaml -p sleeptime=0 -p inputfile=data/names.txt  -p helloworld=code/helloworld.py -d initdir=`pwd`/yadage-inputs 
2020-01-16 14:56:34,688 | packtivity.asyncback |   INFO | configured pool size to 12
Error: Workflow failed: yadage meta directory ./_yadage/ exists. Allow overwrite by using the command line option  --accept-metadir
```

Actions:
- [ ] Allow to call `run` more than one time by solving (1) in RWC
- [ ] Solve (2) for `reana-workflow-engine-serial`
[...]

PR #361, opened 2020-01-21
cli: adds rerun flag for start command

roksys:
* Addresses #355

### DIFF
--- a/docs/userguide.rst
+++ b/docs/userguide.rst
@@ -252,18 +252,18 @@ option to the ``reana-client start`` or ``reana-client run`` commands.
 
 .. code-block:: console
 
-   $ reana-client start -w myanalysis.42 -o TARGET='gendata'
+   $ reana-client start -w myanalysis.42 -o TARGET=gendata
    # or
-   $ reana-client run -w myanalysis.42 -o TARGET='gendata'
+   $ reana-client run -w myanalysis.42 -o TARGET=gendata
 
 If you want to start workflow execution from a certain step you should use
 operational option ``FROM`` with a desired step name.
 
 .. code-block:: console
 
-   $ reana-client start -w myanalysis.42 -o FROM='fitdata'
+   $ reana-client start -w myanalysis.42 -o FROM=fitdata
    # or
-   $ reana-client run -w myanalysis.42 -o FROM='fitdata'
+   $ reana-client run -w myanalysis.42 -o FROM=fitdata
 
 If you want to restart workflow on the same workspace you should use
 ``--restart`` flag of ``reana-client start``.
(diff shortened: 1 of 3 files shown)


reana_client/api/client.py
```diff
@@ -209,20 +209,22 @@ def create_workflow_from_json(workflow_json, name, access_token,
         raise e
 
 
-def start_workflow(workflow, access_token, parameters):
+def start_workflow(workflow, access_token, parameters, rerun):
```
tiborsimko: Let's rather call the option `--restart` to keep the naming harmony with the name of the command `start`.

docs/userguide.rst
```diff
@@ -252,9 +252,29 @@ option to the ``reana-client start`` or ``reana-client run`` commands.
...
+If you want to restart workflow on the same workspace you should use
+``--restart`` flag of ``reana-client start``.
+
+Note that workflow restarting can be used in a combination with operational
+options ``FROM`` or ``TARGET``.
+
+.. code-block:: console
+
+   # Restarts workflow on the same workspace from step named fitdata
+   $ reana-client start -w workflow.1 -o FROM='fitdata' --restart
```
tiborsimko: `s/workflow\./myanalysis\./g` -- since we tend to use `myanalysis.42` elsewhere in the docs, it would be good to harmonise onto that value

docs/userguide.rst
```diff
@@ -252,9 +252,29 @@ option to the ``reana-client start`` or ``reana-client run`` commands.
...
+
+   $ reana-client start -w workflow.1 -o FROM='fitdata'
+   # or
+   $ reana-client run -w workflow.1 -o FROM='fitdata'
+
+If you want to restart workflow on the same workspace you should use
+``--restart`` flag of ``reana-client start``.
+
+Note that workflow restarting can be used in a combination with operational
+options ``FROM`` or ``TARGET``.
```
tiborsimko: FROM "and" TARGET, rather than "or", since people can use both.  

(Actually we could even add another line in the example that uses both, perhaps from mystep3 to mystep7, to show the potential.)

docs/userguide.rst
```diff
@@ -252,9 +252,31 @@ option to the ``reana-client start`` or ``reana-client run`` commands.
...
+If you want to restart workflow on the same workspace you should use
+``--restart`` flag of ``reana-client start``.
+
+Note that workflow restarting can be used in a combination with operational
+options ``FROM`` and ``TARGET``.
+
+.. code-block:: console
+
+   # Restarts workflow on the same workspace from step named fitdata
+   $ reana-client start -w myanalysis.42 -o FROM='fitdata' --restart
```
tiborsimko: We don't need single quotes in CLI command invokations, so we can remove them from the examples, so that they would read more nicely:

`-o TARGET=gendata`
`-o FROM=gendata`
`...`

docs/userguide.rst
```diff
@@ -264,9 +286,9 @@ the ``reana-client start`` or ``reana-client run`` commands.
 
 .. code-block:: console
 
-   $ reana-client start -w workflow.1 -o target='gendata'
+   $ reana-client start -w myanalysis.42 -o --target=gendata
```
tiborsimko: Ah, indeed, we are using engine options as such, so dashes included.

(This raises a question about possibly harmonising these across engines, e.g. -o TARGET meaning --target in CWL world, and something else in Serial world, and something else in Snakemake world.... This would allow us to use nicer word pairs for Serial maybe? (`from ... until ...` would be a very natural English option combination. We can muse on this later...

PR #366, opened 2020-02-06
Move restart to separte cmd

roksys:
* Allows to pass workflow specification
    Closes reanahub/reana-workflow-controller/issues/289

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,7 +19,7 @@ readme = open('README.rst').read()
 history = open('CHANGES.rst').read()
 
 tests_require = [
-    'pytest-reana>=0.7.0.dev20191219,<0.8.0',
+    'pytest-reana==0.7.0.dev20191219',
 ]
 
 extras_require = {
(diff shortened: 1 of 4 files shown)


roksys:
Autodiscovery of workflow specs changes also could be done in the future. Open API specs then might n

setup.py
```diff
@@ -19,7 +19,7 @@
 history = open('CHANGES.rst').read()
 
 tests_require = [
-    'pytest-reana>=0.7.0.dev20191219,<0.8.0',
+    'pytest-reana==0.7.0.dev20191219',
```
tiborsimko: Doesn't seem good to pin to the old version in case we upgrade pytest-reana for client stuff in the future?!

But actually no REANA package uses today's released pytest-reana 0.7.0.dev20200206 minimal version yet...

tiborsimko:
Passing manually `reana.yaml` in those cases when the workflow definition changes sounds perfectly fine for now :+1:

PR #376, opened 2020-02-10
cli: adds option to get logs of a specific step

### LINKED ISSUE #369: cli: `logs` for given workflow step
The `logs` command output may be significant in case a workflow has many steps.

It will be useful to allow getting outputs of some particular step, for example:

```
$ reana-client logs -w bsm --step select_signal_hist_0
```

(stems from #348)

roksys:
* Closes #369

### DIFF
--- a/reana_client/cli/workflow.py
+++ b/reana_client/cli/workflow.py
@@ -627,9 +627,16 @@ def workflow_status(ctx, workflow, _filter, output_format,
     count=True,
     help='Get output in JSON format.')
 @add_access_token_options
+@click.option(
+    '-s',
+    '--step',
+    'steps',
+    multiple=True,
+    help='Get logs of a specific step.')
 @check_connection
 @click.pass_context
-def workflow_logs(ctx, workflow, access_token, json_format):  # noqa: D301
+def workflow_logs(ctx, workflow, access_token, json_format,
+                  steps=None):  # noqa: D301
     """Get  workflow logs.
 
     The `logs` command allows to retrieve logs of running workflow. Note that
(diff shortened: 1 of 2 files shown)


reana_client/cli/workflow.py
```diff
@@ -638,15 +645,15 @@ def workflow_logs(ctx, workflow, access_token, json_format):  # noqa: D301
...
     """
     from reana_client.api.client import get_workflow_logs
     logging.debug('command: {}'.format(ctx.command_path.replace(" ", ".")))
     for p in ctx.params:
         logging.debug('{param}: {value}'.format(param=p, value=ctx.params[p]))
-
     if workflow:
         try:
-            response = get_workflow_logs(workflow, access_token)
+            response = get_workflow_logs(workflow, access_token, steps or None)
```
mvidalgarcia: Do you need `or None`? You already have `steps=None` in the method signature.

reana_client/cli/workflow.py
```diff
@@ -657,6 +666,15 @@ def workflow_logs(ctx, workflow, access_token, json_format):  # noqa: D301
...
             first = True
+            returned_step_names = \
+                set(workflow_logs['job_logs'][item]['job_name']
+                    for item in workflow_logs['job_logs'].keys())
+            if len(returned_step_names) != len(steps):
+                for item in steps:
+                    if item not in returned_step_names:
+                        click.echo(
+                            click.style('Logs of step {} were not found.'
+                                        .format(item), fg='red'))
```
diegodelemos: I think this could be more clear if we directly use `set` operations:
```python
>> server_response = {'step_1': 'I am the step one', 'step_2': 'I am the step two', 'step_3': 'I am the step three'}
>> user_requested_steps = ['step_1', 'step_2', 'step_3']
>> missing_steps = set(user_requested_steps).difference(server_response.keys())
>> print(missing_steps)
{'step_3'}
```
What do you think? Then we could say `'The logs for {} could not be retrieved, check for spelling mistakes in the step names.'.format(', '.join(missing_steps))`

PR #378, opened 2020-02-14
cli: modifies error msg

### LINKED ISSUE #372: cli: "Workflow could not"
**Current behaviour**

We have several client operiations such as `delete` that output message:

```
Workflow could not be deleted:
Expecting value: line 1 column 1 (char 0)
```

In these messages, it would be reassuring for the user if the workflow name appears explicitly.

**Expected behaviour**

Seeing 

```
Workflow Htt.15 could not be deleted
```

would be more advantageous for the user in case the user has many workflow runs.

**Note**

This applies not only to `delete` but to other commands:

```console
$ git grep 'Workflow could not'
reana_client/cli/workflow.py:258:            click.style('Workflow could not be created: \n{}'
reana_client/cli/workflow.py:370:                click.style('Workflow could not be started: \n{}'
reana_client/cli/workflow.py:478:                click.style('Workflow could not be started: \n{}'
reana_client/cli/workflow.py:776:            click.secho('Workflow could not be stopped: \n{}'.format(str(e)),
reana_client/cli/workflow.py:917:                click.style('Workflow could not be deleted: \n{}'
```

**Cosmetics** 

We could reformulate the passive tense ("Workflow foo.bar cannot be...") to active tense ("Cannot delete workflow foo.bar.") in case the workflow names are long.

roksys:
* Closes #372

### DIFF
--- a/reana_client/cli/workflow.py
+++ b/reana_client/cli/workflow.py
@@ -266,8 +266,8 @@ def workflow_create(ctx, file, name,
         logging.debug(traceback.format_exc())
         logging.debug(str(e))
         click.echo(
-            click.style('Workflow could not be created: \n{}'
-                        .format(str(e)), fg='red'),
+            click.style('Cannot create workflow {}: \n{}'
+                        .format(name, str(e)), fg='red'),
             err=True)
         if 'invoked_by_subcommand' in ctx.parent.__dict__:
             sys.exit(1)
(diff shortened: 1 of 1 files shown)


reana_client/cli/workflow.py
```diff
@@ -266,8 +266,8 @@ def workflow_create(ctx, file, name,
         logging.debug(traceback.format_exc())
         logging.debug(str(e))
         click.echo(
-            click.style('Workflow could not be created: \n{}'
-                        .format(str(e)), fg='red'),
+            click.style('Workflow {} could not be created: \n{}'
+                        .format(name, str(e)), fg='red'),
```
diegodelemos: While re-reading https://github.com/reanahub/reana-client/issues/372:
> We could reformulate the passive tense ("Workflow foo.bar cannot be...") to active tense ("Cannot delete workflow foo.bar.") in case the workflow names are long.

I think this would read better and faster, if we keep it like it is right now the most important part of the phrase will be pushed back to the end when the name is very big, what do you think?

diegodelemos:
Needs rebase.

PR #392, opened 2020-02-25
cli: accept reana.yml spec filename

### LINKED ISSUE #385: reana.yaml: accept also reana.yml
Some users created file called `reana.yml`.

The preferred extension for YAML files is `yaml`, see https://yaml.org/faq.html

However, we should accept `reana.yml` too if `reana.yaml` is not present.

And we should warn if *two of them* are both present.

So:
- look for presence of `reana.yaml` and `reana.yml`, if both are found, exit with an error
- look for `reana.yaml` , if it exists, use it
- look for `reana.yml`, if it exists, use it
- if no one was found, print error message "You need to create `reana.yaml` blah blah blah..."

mvidalgarcia:
closes #385

### DIFF
--- a/reana_client/utils.py
+++ b/reana_client/utils.py
@@ -373,8 +373,9 @@ def get_reana_yaml_file_path():
     if len(matches) > 1:
         click.echo(
             click.style(
-                '==> ERROR: Found two REANA specification files ({0}). '
-                'Please use only one. Exiting.'.format(', '.join(matches)),
+                '==> ERROR: Found {0} REANA specification files ({1}). '
+                'Please use only one. Exiting.'.format(len(matches),
+                                                       ', '.join(matches)),
                 fg='red'))
         sys.exit(1)
     for path in reana_yaml_valid_file_names:
--- a/reana_client/utils.py
+++ b/reana_client/utils.py
@@ -22,7 +22,8 @@ from jsonschema import ValidationError, validate
 from reana_commons.serial import serial_load
 from reana_commons.utils import get_workflow_status_change_verb
 
-from reana_client.config import reana_yaml_schema_file_path
+from reana_client.config import (reana_yaml_schema_file_path,
+                                 reana_yaml_valid_file_names)
 
 
 def workflow_uuid_or_name(ctx, param, value):
@@ -220,9 +221,8 @@ def get_workflow_name_and_run_number(workflow_name):
 
 def get_workflow_root():
     """Return the current workflow root directory."""
-    reana_yaml = 'reana.yaml'
+    reana_yaml = get_reana_yaml_file_path()
     workflow_root = os.getcwd()
-
     while True:
         file_list = os.listdir(workflow_root)
         parent_dir = os.path.dirname(workflow_root)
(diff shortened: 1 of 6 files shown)


reana_client/utils.py
```diff
@@ -358,3 +358,20 @@ def parse_secret_from_path(path):
...
+
+
+def get_reana_yaml_file_path():
+    """REANA specification file location."""
+    matches = [path for path in reana_yaml_valid_file_names
+               if os.path.exists(path)]
+    if len(matches) > 1:
+        click.echo(
+            click.style(
+                'You must have only one specification file. '
```
tiborsimko: Better to make the message less personal, and better to indicate clearly that we are exiting in this case... Something like:

```console
$ reana-client run -w test
==> ERROR: Found two REANA specification files (reana.yaml, reana.yml). Please use only one. Exiting.
```

We could use the same technique when no spec file is found:

```console
$ reana-client run -w test
==> ERROR: No REANA specification file (reana.yaml) found. Exiting.
```

Currently we print:

```console
$ reana-client run -w test
Usage: reana-client run [OPTIONS]
Try "reana-client run --help" for help.

Error: Invalid value for "-f" / "--file": Path "reana.yaml" does not exist.
```

which is perhaps a bit less friendly to the user... 

Otherwise the code looks good and the functionality works well :+1: 

P.S. The output styling  (==>) used above is just for illustration, but we can indeed start generalising it here and there in CLI commands, as we started to introduce it in `logs` command... see e.g. [yay screenshot](https://camo.githubusercontent.com/90b22c7fc8c60528a78f055c93cd33e05a991dd9/68747470733a2f2f72617763646e2e6769746861636b2e636f6d2f4a677565722f6a677565722e6769746875622e696f2f373736343766333936636237313536666433326533303937306462656166366436646337663938332f7961792f7961792d732e706e67).

mvidalgarcia: > Better to make the message less personal, and better to indicate clearly that we are exiting in this case... Something like:
> 
> ```
> $ reana-client run -w test
> ==> ERROR: Found two REANA specification files (reana.yaml, reana.yml). Please use only one. Exiting.
> ```
Agreed.

> 
> We could use the same technique when no spec file is found:
> 
> ```
> $ reana-client run -w test
> ==> ERROR: No REANA specification file (reana.yaml) found. Exiting.
> ```
> 
> Currently we print:
> 
> ```
> $ reana-client run -w test
> Usage: reana-client run [OPTIONS]
> Try "reana-client run --help" for help.
> 
> Error: Invalid value for "-f" / "--file": Path "reana.yaml" does not exist.
> ```
> 
> which is perhaps a bit less friendly to the user...

I thought about this at the beginning, but I think then we'd be losing valuable information for the user explaining how to pass a `reana.yaml` manually, WDYT? cc/ @diegodelemos

tiborsimko: I guess users will mostly use the default file name anyway, as they see it in the docs and other examples. It does not seem probable that someone would not have any `reana.yaml` but would rather insist on using say `.reana.yaml` and passing the `-f` option all the time.  It is more probable that people would have both `reana.yaml` and some other file, such as `reana-quicktest.yaml`. Hence I think it is OK to alter the message when there is no `reana.yaml`; these people would be mostly novices who would like to have `reana.yaml` anyway.  So the shorter message should be OK.

BTW, speaking of `-f', we should upgrade one day `reana-client` help messages for the `-f` flag anyway, since we are not listing there that `reana.yaml` is the default:

```
 -f, --file PATH                 REANA specifications file describing the
                                  workflow and context which REANA should
                                  execute.
```

This would be better simplified into _"REANA specification file describing the workflow to execute. [default=reana.yaml]"_ in every command.

We can leave some of these improvements and cosmetic formatting styles for later, my important comment was about the output message verbiage; all the other remarks about formatting etc were just for bonus points :wink:

diegodelemos:
Needs rebase.

reana_client/utils.py
mvidalgarcia: OK, makes sense to me. Updated and help messages upgraded.

PR #400, opened 2020-03-23
global: allow operational options in reana.yaml

### LINKED ISSUE #356: yadage: make toplevel configurable
yadage load has a configurable kwarg `toplevel` but currently I don't see any way to configure this from `reana.yml`. From the code it seems like this could be easy to add?

https://github.com/reanahub/reana-client/blob/fa025d0e57fbdde5c9ff86595e3c991cb2a2483e/reana_client/utils.py#L40

mvidalgarcia:
closes #356

`reana.yaml` could be like:
```yaml
version: 0.6.0
[...]
workflow:
  type: yadage
  file: workflow.yaml
  options: 
    toplevel: './example'
[...]
```

###### TODO:
- [x] Validation of `reana.yaml` operational options on creation.
- [ ] Execute a CWL workflow with `-o TARGET ...` successfully.

### DIFF
--- a/reana_client/cli/utils.py
+++ b/reana_client/cli/utils.py
@@ -136,7 +136,7 @@ def validate_workflow_name(ctx, _, workflow_name):
     return workflow_name
 
 
-def params_tuple_to_dict(ctx, param, value):
+def key_value_to_dict(ctx, param, value):
     """Convert tuple params to dictionary. e.g `(foo=bar)` to `{'foo': 'bar'}`.
 
     :param options: A tuple with CLI operational options.
--- a/reana_client/cli/utils.py
+++ b/reana_client/cli/utils.py
@@ -136,6 +136,21 @@ def validate_workflow_name(ctx, _, workflow_name):
     return workflow_name
 
 
+def params_tuple_to_dict(ctx, param, value):
+    """Convert tuple params to dictionary. e.g `(foo=bar)` to `{'foo': 'bar'}`.
+
+    :param options: A tuple with CLI operational options.
+    :returns: A dictionary representation of the given options.
+    """
+    try:
+        return dict(op.split('=') for op in value)
+    except ValueError as err:
+        click.secho('==> ERROR: Input parameter "{0}" is not valid. '
+                    'It must follow format "param=value".'
+                    .format(' '.join(value)), err=True, fg='red'),
+        sys.exit(1)
+
+
 class NotRequiredIf(click.Option):
     """Allow only one of two arguments to be missing."""
 
--- a/reana_client/utils.py
+++ b/reana_client/utils.py
@@ -83,10 +83,9 @@ def cwl_load(workflow_file, **kwargs):
     :returns: A dictionary which represents the valid `cwl` workflow.
     """
     result = \
-        subprocess.run(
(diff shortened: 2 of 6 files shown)


reana_client/utils.py
```diff
@@ -119,20 +119,24 @@ def load_reana_spec(filepath, skip_validation=False):
...
-        if reana_yaml['workflow']['type'] == 'serial':
+        workflow_type = reana_yaml['workflow']['type']
+        if workflow_type == 'serial':
             kwargs['specification'] = reana_yaml['workflow'].\
                 get('specification')
             kwargs['parameters'] = \
                 reana_yaml.get('inputs', {}).get('parameters', {})
             kwargs['original'] = True
 
+        if 'options' in reana_yaml['workflow']:
```
diegodelemos: We should add this new field to the [schema](https://github.com/reanahub/reana-client/blob/master/reana_client/schemas/reana_analysis_schema.json).

diegodelemos:
Needs rebase ⏭

reana_client/cli/utils.py
```diff
@@ -136,6 +136,21 @@ def validate_workflow_name(ctx, _, workflow_name):
     return workflow_name
 
 
+def params_tuple_to_dict(ctx, param, value):
```
diegodelemos, line 139: Could we make the name more explanatory? `params_tuple_to_dict` seems too low level, what about more high level, `key_value_to_dict` because it takes`-o KEY=VALUE` and gives `{'KEY': 'VALUE'}` or something simpler.

reana_client/utils.py
```diff
@@ -38,7 +40,7 @@ def workflow_uuid_or_name(ctx, param, value):
         return value
 
 
-def yadage_load(workflow_file, toplevel='.'):
+def yadage_load(workflow_file, toplevel='.', **kwargs):
```
diegodelemos, line 43: Who is consuming these `**kwargs`? We already unpack `toplevel` which is the only one we are interested in.

reana_client/utils.py
```diff
@@ -65,15 +67,15 @@ def yadage_load(workflow_file, toplevel='.'):
...
+        yadageschemas.load(spec=workflow_file, specopts=specopts,
+                           validopts=validopts, validate=True)
 
     except ValidationError as e:
         e.message = str(e)
         raise e
 
 
-def cwl_load(workflow_file):
+def cwl_load(workflow_file, **kwargs):
```
diegodelemos, line 78: Also here.

reana_client/utils.py
mvidalgarcia: If we set other `inputs.options` such as `initdir`, [its value will get here](https://github.com/reanahub/reana-client/pull/400/files#diff-670175eb24e7fca435fba316a6b7066fR137) and this function would fail. With `**kwargs` we can _filter out_ these options that are not needed on validation time. I agree it's a bit hacky but it's cleanest and safest way I could come up with. If you have any other idea please let me know 🙂

reana_client/utils.py
mvidalgarcia: Same here. For CWL workflows, when passing `TARGET: foo` for example, it will end up here.

PR #405, opened 2020-03-26
cli: upload & download based on saved spec

### LINKED ISSUE #403: cli: reana-client run looks for input files in reana.yaml
Regardless what reana file name you pass in using `-f` it always tries to get the input files from a `reana.yaml` file.

**Current result**

```console
$ reana-client run -f my-reana.yaml -w test         
[INFO] Creating a workflow...
test.1
[INFO] Uploading files...
==> ERROR: No REANA specification file (reana.yaml) found. Exiting.
```

**Expected result**
```console
$ reana-client run -f my-reana.yaml -w test         
[INFO] Creating a workflow...
test.1
[INFO] Uploading files...
File code/gendata.C was successfully uploaded.                                                                                
File code/fitdata.C was successfully uploaded.  
[INFO] Starting workflow...
test.1 has been queued
```

mvidalgarcia:
Instead of go looking for `reana.yaml` in the current (or parents)
directory to determine the input and output files, take into account
the previously created workflow and retrive its specification to get
this information.

closes #403

### DIFF
--- a/reana_client/api/client.py
+++ b/reana_client/api/client.py
@@ -17,8 +17,7 @@ import requests
 from bravado.exception import HTTPError
 from reana_client.config import ERROR_MESSAGES, WORKFLOW_ENGINES
 from reana_client.errors import FileDeletionError, FileUploadError
-from reana_client.utils import (_validate_reana_yaml, get_workflow_root,
-                                is_uuid_v4)
+from reana_client.utils import _validate_reana_yaml, is_uuid_v4
 from reana_commons.api_client import get_current_api_client
 from reana_commons.errors import (REANASecretAlreadyExists,
                                   REANASecretDoesNotExist)
@@ -344,9 +343,10 @@ def download_file(workflow_id, file_name, access_token):
             return http_response.content
         else:
             raise Exception(
-                "Expected status code 200 but replied with "
-                "{status_code}".format(
-                    status_code=http_response.status_code))
+                "Error {status_code} {reason} {message}".format(
+                    status_code=http_response.status_code,
+                    reason=http_response.reason,
+                    message=http_response.json().get('message')))
 
     except HTTPError as e:
         logging.debug(
@@ -486,15 +486,10 @@ def upload_to_server(workflow, paths, access_token):
                 symlink = True
             with open(path, 'rb') as f:
                 fname = os.path.basename(f.name)
-                workflow_root = get_workflow_root()
-                if not path.startswith(workflow_root):
-                    raise FileUploadError(
-                        'Files and directories to be uploaded '
-                        'must be under the workflow root directory.')
                 # Calculate the path that will store the file
                 # in the workflow controller, by subtracting
                 # the workflow root path from the file path
-                save_path = path.replace(workflow_root, '')
+                save_path = path.replace(os.getcwd(), '')
                 # Remove prepending dirs named "." or as the upload type
                 while len(save_path.split('/')) > 1 and \
                         save_path.split('/')[0] == '.':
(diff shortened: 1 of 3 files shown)


reana_client/api/client.py
```diff
@@ -486,15 +485,10 @@ def upload_to_server(workflow, paths, access_token):
...
-                workflow_root = get_workflow_root()
-                if not path.startswith(workflow_root):
-                    raise FileUploadError(
-                        'Files and directories to be uploaded '
-                        'must be under the workflow root directory.')
                 # Calculate the path that will store the file
                 # in the workflow controller, by subtracting
                 # the workflow root path from the file path
-                save_path = path.replace(workflow_root, '')
+                save_path = path.replace(os.getcwd(), '')
```
tiborsimko, line 491: Works well :+1:

**(1)** If we stop searching for workflow root, this will mean that people cannot be located in some nested directory when launching commands? IIRC they cannot submit easily from subdirs today either, so we can simply document this... so that people are not surprised.  Example:

```console
$ reana-client create -w myanalysis
$ reana-client upload mycode
$ # oh, I need to fix something
$ cd mycode/firstpart
$ vim myfirstpart.py
$ reana-client upload . -w myfirstpart
$ # oops, where this file did go? into root?
```

Ideally, we can behave analogously with how `git` behaves when being located in a subdirectory, so that users are not surprised... but this can wait. We can document current behaviour and amend later.


**(2)** Reading the specs only from the server will interfere with development stages of the workflow. Consider the following usage scenario:

```console
$ reana-client create -w myanalysis
$ vim reana.yaml # using mydata.csv and mycode.py
$ vim mycode.py
$ # ... working on mycode.py
$ # ... oops, I forgot to specify mybackground.csv too! let's fix it:
$ cp /mynfs/somewhere/mybackground.csv .
$ vim reana.yaml # adding mybackground.csv
$ reana-client upload
... oops, only mydata.csv was uploaded; why not background?
```

It is also OK for now if we document it; however for people using Serial will be surprised, as they will be frequently editing both the code and the workflow definition while working on an analysis. We have to document to always use `reana-client upload mynewfile.py` if need be manually. So, ideally, we should still parse the current reana.yaml file too to see about any user-visible changes... It can be done later, but the "get_workflow_root()" function might be back.

mvidalgarcia: **(1)** Exactly, I removed that function because I think it was adding too much complexity. I think it feels more natural to be able to upload files from any directory than _magically_ try to find the `reana.yaml` directory. In addition, we don't know the name of the reana specification file at this operation, so we wouldn't really know what to look for. I don't know what you think.

**(2)** IMHO it doesn't make sense to look for a _random_ `reana.yaml` file, it could be whatever. Another option would be to habilitate the `-f` option in the upload command so it can parse the inputs again. So, when doing `reana-client upload` do:
- If `-f` is present, parse inputs in provided `reana.yaml` file and upload them.
- Else, do the behavior of this PR, fetch specification from the server and use the inputs from there.

tiborsimko: (1) Agreed. Once we have `.reana` folder, we can do like `.git` does.

(2) It wouldn't be random; it would be the one that the user used during `create -f ...` step. (One could know it from the server-side or from `.reana` once we go for it.)

We can ticketise some of these for later...

reana_client/api/client.py
```diff
@@ -344,9 +343,10 @@ def download_file(workflow_id, file_name, access_token):
             return http_response.content
         else:
             raise Exception(
-                "Expected status code 200 but replied with "
-                "{status_code}".format(
-                    status_code=http_response.status_code))
+                "Error {status_code} {reason} {message}".format(
+                    status_code=http_response.status_code,
+                    reason=http_response.reason,
+                    message=http_response.json().get('message')))
```
tiborsimko, line 349: Thanks, it looks better, but sometimes I see the message gets long:

A. this is nice:

```console
$ reana-client download wrong.txt -w root6-roofit-cwl
File wrong.txt could not be downloaded: Error 404 Not Found wrong.txt does not exist.
```

B. this is less nice:

```console
$ reana-client download wrong.txt -w notexisting
File wrong.txt could not be downloaded: Error 404 Not Found REANA_WORKON is set to notexisting, but that workflow does not exist. Please set your REANA_WORKON environment variable appropriately.
```

I can ticketise for later...

PR #409, opened 2020-04-15
docs: add script to generate cli api docs

mvidalgarcia:
closes reanahub/docs.reana.io#22

### DIFF
--- a/run-tests.sh
+++ b/run-tests.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 #
 # This file is part of REANA.
 # Copyright (C) 2017, 2018 CERN.
@@ -6,11 +6,22 @@
 # REANA is free software; you can redistribute it and/or modify it
 # under the terms of the MIT License; see LICENSE file for more details.
 
+
+cli_docs_url=https://raw.githubusercontent.com/reanahub/docs.reana.io/master/docs/reference/reana-client-cli-api/index.md
+docs_differ_error_msg='Current reana-client differs with the documentation. Please update http://docs.reana.io/reference/reana-client-cli-api/.'
+python_version=$(python -c 'import sys;  print(sys.version_info.major)')
+
 pydocstyle reana_client && \
 isort -rc -c -df **/*.py && \
 reana-client --help > cmd_list.txt && \
 diff -q -w docs/cmd_list.txt cmd_list.txt  && \
 rm cmd_list.txt && \
+if [ "$python_version" -eq 3 ]
+then
+    python scripts/generate_cli_api.py > cli_api.md && \
+    (diff -q -w  cli_api.md <(curl -s $cli_docs_url) || (echo $docs_differ_error_msg && exit 1)) && \
+    rm cli_api.md
+fi
 check-manifest --ignore ".travis-*" && \
 sphinx-build -qnNW docs docs/_build/html && \
 python setup.py test && \
--- a/scripts/generate_cli_api.py
+++ b/scripts/generate_cli_api.py
@@ -1,0 +1,35 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of REANA.
+# Copyright (C) 2020 CERN.
+#
+# REANA is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+"""REANA client CLI API docs generation."""
+
+import click
+
+from reana_client.cli import cli
+
+
+def _print_code_block(content, lang=''):
+    print('```{}'.format(lang))
+    print(content)
+    print('```')
+
+
+def generate_cli_docs():
+    """Generate Markdown friendly CLI API documentation."""
+    print('# reana-client CLI API\n')
+    with click.Context(cli) as ctx:
+        _print_code_block(cli.get_help(ctx), lang='console')
+
+    for cmd_group in cli.cmd_groups:
+        print('\n## {}'.format(cmd_group.help))
+        for cmd_obj in cmd_group.commands.values():
+            print('\n### {}\n'.format(cmd_obj.name))
+            print(cmd_obj.help)
+
+
+if __name__ == "__main__":
+    generate_cli_docs()
--- a/.gitignore
+++ b/.gitignore
@@ -71,3 +71,7 @@ target/
 
 # Visual Studio Code files
 .vscode/*
+
+# Docs autogeneration
+cmd_list.txt
+cli_api.md
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -21,3 +21,4 @@ recursive-include docs *.py
 recursive-include docs *.png
 recursive-include docs *.rst
 recursive-include tests *.py
+recursive-include scripts *.py


scripts/generate_cli_api.py
```diff
@@ -0,0 +1,27 @@
...
+    print('```')
+
+
+print('# reana-client CLI API')
+with click.Context(cli) as ctx:
+    _print_code_block(cli.get_help(ctx))
+print('## Commands')
+for cmd_obj in cli.commands.values():
+    print('## {}'.format(cmd_obj.name))
+    print(cmd_obj.help)
```
diegodelemos, line 27: What do you think about encapsulating all of this inside a function which we can name, and where we can write a docstring explaining what it does? then we can call it right after.

run-tests.sh
```diff
@@ -1,16 +1,22 @@
...
+
+cli_docs_url=https://raw.githubusercontent.com/reanahub/docs.reana.io/master/docs/reference/reana-client-cli-api/index.md
+
 pydocstyle reana_client && \
 isort -rc -c -df **/*.py && \
 reana-client --help > cmd_list.txt && \
 diff -q -w docs/cmd_list.txt cmd_list.txt  && \
 rm cmd_list.txt && \
+python scripts/generate_cli_api.py > cli_api.md && \
+diff -q -w  cli_api.md <(curl -s $cli_docs_url) && \
```
diegodelemos, line 18: What do you think about improving the message so when we read the Travis failure (for example [this one](https://travis-ci.org/github/reanahub/reana-client/jobs/675244135#L626)) we don't get something like `Files cli_api.md and /dev/fd/63 differ` but something more descriptive?
```bash
... && \
$docs_differ_error_msg='Current reana-client differs with the documentation. Please update http://docs.reana.io/reference/reana-client-cli-api/.'
(diff -q -w  cli_api.md <(curl -s $cli_docs_url) || (echo $docs_differ_error_msg && exit 1)) && \
...
```

scripts/generate_cli_api.py
```diff
@@ -0,0 +1,27 @@
...
+    print(content)
+    print('```')
+
+
+print('# reana-client CLI API')
+with click.Context(cli) as ctx:
+    _print_code_block(cli.get_help(ctx))
+print('## Commands')
+for cmd_obj in cli.commands.values():
+    print('## {}'.format(cmd_obj.name))
```
diegodelemos, line 26:
```suggestion
    print('### {}'.format(cmd_obj.name))
```

Otherwise, the commands do not appear nested.
image

diegodelemos: More on nesting, what do you think of keeping the same logic separation we have in the help command inside the docs?
```
Usage:  [OPTIONS] COMMAND [ARGS]...

  REANA client for interacting with REANA server.

Options:
  -l, --loglevel [DEBUG|INFO|WARNING]
                                  Sets log level
  --help                          Show this message and exit.

Configuration commands:
  ping     Check connection to REANA server.
  version  Show version.

Workflow management commands:
  create  Create a new workflow.
  delete  Delete a workflow.
  diff    Show diff between two workflows.
  list    List all workflows and sessions.

Workflow execution commands:
  logs      Get workflow logs.
  restart   Restart previously run workflow.
  run       Shortcut to create, upload, start a new workflow.
  start     Start previously created workflow.
  status    Get status of a workflow.
  stop      Stop a running workflow.
  validate  Validate workflow specification file.

Workspace interactive commands:
  close  Close an interactive session.
  open   Open an interactive session inside the workspace.

Workspace file management commands:
  download  Download workspace files.
  du        Get workspace disk usage.
  ls        List workspace files.
  mv        Move files within workspace.
  rm        Delete files from workspace.
  upload    Upload files and directories to workspace.

Secret management commands:
  secrets-add     Add secrets from literal string or from file.
  secrets-delete  Delete user secrets by name.
  secrets-list    List user secrets.
```

So we would yield the following in markdown:
```markdown
# reana-client CLI API
## Configuration commands
### ping
...
### version
....
## Workflow management commands
### create
...
#### delete
....
```



We already have the logic to go through the click groups [here](https://github.com/reanahub/reana-client/blob/e2eb481775e8a0ae194ca2b66fa601c808246761/reana_client/cli/__init__.py#L58-L71).

scripts/generate_cli_api.py
```diff
@@ -0,0 +1,27 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of REANA.
+# Copyright (C) 2020 CERN.
+#
+# REANA is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+"""REANA client CLI API docs generation."""
```
diegodelemos, line 8: We don't need to do this now, but if you agree we can create an issue to add something similar to what we have for the Openapi specs, with a parameter you can instruct the script to copy the file to the correct location (`docs.reana.io/docs/reference/reana-client-cli-api/index.md`), see [here](https://github.com/reanahub/reana-workflow-controller/blob/8ba38e9b44621f58c63da1fd4403f5b4065910fb/scripts/generate_openapi_spec.py#L99-L100) how we do it in RWC.

scripts/generate_cli_api.py
mvidalgarcia: 1st comment: Sure, actually I have it changed locally but I didn't push it.
2nd comment: OK, I see that's easy to navigate inside the groups:
```py
In [1]: from reana_client.cli import cli                                                                                                                                                                                                          

In [2]: cli.cmd_groups                                                                                                                                                                                                                            
Out[2]: 
[<Group configuration-group>,
 <Group workflow-management-group>,
 <Group workflow-execution-group>,
 <Group interactive-group>,
 <Group files-group>,
 <Group secrets-group>]

In [3]: cli.cmd_groups[0]                                                                                                                                                                                                                         
Out[3]: <Group configuration-group>

In [4]: cli.cmd_groups[0].commands                                                                                                                                                                                                                
Out[4]: {'ping': <Command ping>, 'version': <Command version>}
```

scripts/generate_cli_api.py
mvidalgarcia: https://github.com/reanahub/reana-client/issues/410

run-tests.sh
```diff
@@ -1,16 +1,23 @@
...
+docs_differ_error_msg='Current reana-client differs with the documentation. Please update http://docs.reana.io/reference/reana-client-cli-api/.'
+
 pydocstyle reana_client && \
 isort -rc -c -df **/*.py && \
 reana-client --help > cmd_list.txt && \
 diff -q -w docs/cmd_list.txt cmd_list.txt  && \
 rm cmd_list.txt && \
+python scripts/generate_cli_api.py > cli_api.md && \
+(diff -q -w  cli_api.md <(curl -s $cli_docs_url) || (echo $docs_differ_error_msg && exit 1)) && \
+rm cli_api.md && \
```
tiborsimko, line 20: Put `cli_api.md` into `.gitignore` in case something fails before it is removed?

Alternatively use `/tmp/cli_api.md` place for this "temporary" file?

PR #414, opened 2020-06-09
release: v0.6.1

tiborsimko:


diegodelemos:
I think we are missing branch protection on `maint-0.6` here and the other repos.

PR #416, opened 2020-06-30
cli: return only failed logs

### LINKED ISSUE #415: cli: option to show failed logs
Add an option to only show failed job logs.

Example:

```
reana-client logs -w workflow --failed
```

Motivation: 

If a user runs a big workflow with hundreds of jobs it can be cumbersome to filter through all the job logs to find the subset that failed.

alintulu:
Added the option to return only failed logs.

Example

```
reana-client logs --failed
```

closes #415

### DIFF
--- a/reana_client/cli/workflow.py
+++ b/reana_client/cli/workflow.py
@@ -714,9 +714,12 @@ def workflow_status(
 @click.option(
     "-s", "--step", "steps", multiple=True, help="Get logs of a specific step."
 )
+@click.option("--failed", "failed", is_flag=True, help="Get logs of failed steps.")
 @check_connection
 @click.pass_context
-def workflow_logs(ctx, workflow, access_token, json_format, steps=None):  # noqa: D301
+def workflow_logs(
+    ctx, workflow, access_token, json_format, failed, steps=None
+):  # noqa: D301
     """Get  workflow logs.
 
     The `logs` command allows to retrieve logs of running workflow. Note that
@@ -738,6 +741,17 @@ def workflow_logs(ctx, workflow, access_token, json_format, steps=None):  # noqa
                 steps = list(set(steps))
             response = get_workflow_logs(workflow, access_token, steps)
             workflow_logs = json.loads(response["logs"])
+
+            if failed:
+                finished_steps = [
+                    key
+                    for key, value in workflow_logs["job_logs"].items()
+                    if value["status"] != "failed"
+                ]
+
+                for job_id in finished_steps:
+                    del workflow_logs["job_logs"][job_id]
+
             if json_format:
                 click.echo(json.dumps(workflow_logs, indent=2))
                 sys.exit(0)


alintulu:
Not sure what is the better approach

1. To have it this way, collect the names of the steps which failed and call

```
response = get_workflow_logs(workflow, access_token, steps)
workflow_logs = json.loads(response["logs"])
```

again with the updated steps.

2. Collect the names of the steps that didn't fail and delete them from the workflow_logs with 

```
del workflow_logs["job_logs"][item]
```

Opinions?

diegodelemos:
I think it would be better to go for (2.), that way we won't do a REST call two times (`get_workflow_logs` will call the logs endpoint each time we call it). Also, I don't see a big problem in modifying the object as the command line will "die" at the end of its execution and it won't affect anything but itself.

reana_client/cli/workflow.py
```diff
@@ -738,6 +741,18 @@ def workflow_logs(ctx, workflow, access_token, json_format, steps=None):  # noqa
                 steps = list(set(steps))
             response = get_workflow_logs(workflow, access_token, steps)
             workflow_logs = json.loads(response["logs"])
+
+            if failed:
+                steps = list(
+                    workflow_logs["job_logs"][item]["job_name"]
```
diegodelemos, line 747: I think you can simplify to `[key for key, value in workflow_logs["job_logs"].items() if value["status"] == "failed"]`

alintulu: I will try approach two :)

alintulu:
> I think it would be better to go for (2.), that way we won't do a REST call two times (`get_workflow_logs` will call the logs endpoint each time we call it). Also, I don't see a big problem in modifying the object as the command line will "die" at the end of its execution and it won't affect anything but itself.

I will try approach two :)

alintulu:
You can use this reana.yaml to test

```
version: 0.6.0
inputs:
  parameters:
    events: 20000
workflow:
  type: serial
  specification:
    steps:
      - name: success
        environment: 'python:2.7-slim'
        commands:
        - echo success!
      - name: fails
        environment: 'python:2.7-slim'
        commands:
        - ls /eos/user/a/adlintul
```

```
$ reana-client logs

==> Job logs
==> Step: success
==> Workflow ID: 18a8129f-08d8-4419-830f-c7321bf18309
==> Compute backend: Kubernetes
==> Job ID: reana-run-job-c609856d-a659-4fe3-8c91-63fa5f0d2d01
==> Docker image: python:2.7-slim
==> Command: echo success!
==> Status: finished
==> Logs:
job: :
 success!


==> Step: fails
==> Workflow ID: 18a8129f-08d8-4419-830f-c7321bf18309
==> Compute backend: Kubernetes
==> Job ID: reana-run-job-b8164edc-c023-45e9-8ad4-c3a84ad815ee
==> Docker image: python:2.7-slim
==> Command: ls /eos/user/a/adlintul
==> Status: failed
==> Logs:
job: :
 ls: cannot access '/eos/user/a/adlintul': No such file or directory
```

```
$ reana-client logs --failed

==> Job logs
==> Step: fails
==> Workflow ID: 18a8129f-08d8-4419-830f-c7321bf18309
==> Compute backend: Kubernetes
==> Job ID: reana-run-job-b8164edc-c023-45e9-8ad4-c3a84ad815ee
==> Docker image: python:2.7-slim
==> Command: ls /eos/user/a/adlintul
==> Status: failed
==> Logs:
job: :
 ls: cannot access '/eos/user/a/adlintul': No such file or directory
```

alintulu:
The way it is currenly implemented you can use `--step` in combination with `--failed`. I thought that could be useful, or is it confusing/unecessary?

reana_client/cli/workflow.py
```diff
@@ -738,6 +741,17 @@ def workflow_logs(ctx, workflow, access_token, json_format, steps=None):  # noqa
...
+
+            if failed:
+                finished_steps = [
+                    key
+                    for key, value in workflow_logs["job_logs"].items()
+                    if value["status"] != "failed"
+                ]
+
+                for job_id in finished_steps:
+                    del workflow_logs["job_logs"][job_id]
```
diegodelemos, line 753: All works fine except this message, which can be misleading:
```console
$ reana-client logs -w workflow.1 --step success --failed 
The logs of step(s) success were not found, check for spelling mistakes in the step names.
```
This happens when none of the steps meets the "failed" condition. I think it would make sense to check if the `workflow_logs["job_logs"]` is empty here and return a more accurate message, what do you think?

PR #419, opened 2020-07-13
cli: filter logs

### LINKED ISSUE #415: cli: option to show failed logs

alintulu:
closes #415

Generic way to filter the logs according to `step`, `compute_backend`, `docker_img` and `status`. Replaces the current `-s` `--step` option that returns the logs of desired steps.

Use example

```
$ reana-client logs --filter status=failed --filter compute_backend=htcondor
$ reana-client logs --filter step=gendata
```

### DIFF
--- a/reana_client/cli/workflow.py
+++ b/reana_client/cli/workflow.py
@@ -712,11 +712,14 @@ def workflow_status(
 @click.option("--json", "json_format", count=True, help="Get output in JSON format.")
 @add_access_token_options
 @click.option(
-    "-s", "--step", "steps", multiple=True, help="Get logs of a specific step."
+    "--filter",
+    "filters",
+    multiple=True,
+    help="Filter job logs to include only those steps that match certain filtering criteria. Use --filter name=value pairs. Available filters are compute_backend, docker_img, status and step.",
 )
 @check_connection
 @click.pass_context
-def workflow_logs(ctx, workflow, access_token, json_format, steps=None):  # noqa: D301
+def workflow_logs(ctx, workflow, access_token, json_format, filters=None):  # noqa: D301
     """Get  workflow logs.
 
     The `logs` command allows to retrieve logs of running workflow. Note that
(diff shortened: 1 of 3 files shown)


alintulu:
Some examples of how it handles different inputs

#### Bad formation

```
$ reana-client logs -w workflow --filter status.finished
Cannot read filter from input 'status.finished'. Please give it in the form key=value, e.g. status=failed
```

#### Non-existing filter

```
$ reana-client logs -w workflow --filter statuss=finished
Filter 'statuss' is not applicable. Applicable filters are ['step', 'compute_backend', 'docker_img', 'status']
```

#### Finding no matches

```
$ reana-client logs -w workflow --filter compute_backend=Slurm
No matches found for ['compute_backend=Slurm']
```

reana_client/cli/workflow.py
```diff
@@ -712,11 +712,14 @@ def add_verbose_data_from_response(response, verbose_headers, headers, data):
 @click.option("--json", "json_format", count=True, help="Get output in JSON format.")
 @add_access_token_options
 @click.option(
-    "-s", "--step", "steps", multiple=True, help="Get logs of a specific step."
+    "--filter",
+    "filters",
+    multiple=True,
+    help="Filter the logs with key=value pair. Applicable filters are step, compute_backend, docker_img and status.",
```
tiborsimko, line 718: Please present available filters sorted everywhere in the help output: compute_backend, docker_img, status, step.

alintulu: Do you think "Available filters are.." is better than "Applicable filters are.."?

reana_client/cli/workflow.py
```diff
@@ -729,22 +732,99 @@ def workflow_logs(ctx, workflow, access_token, json_format, steps=None):  # noqa
...
+                        for k, v in workflow_logs["job_logs"].items()
+                        if v[applicable_filters[key]] != value
+                    ]
+                    for job_id in unwanted_steps:
+                        del workflow_logs["job_logs"][job_id]
+
+                if not workflow_logs["job_logs"]:
+                    click.echo(
+                        click.style(
+                            "No matches found for {}\nApplicable filters are {}".format(
```
tiborsimko, line 810: (1) If the filter names are good, and if nothing is found, I think we can treat it as "filter passed successfully, but nothing came up after filter. 

```console
$ reana-client logs -w helloworld-yadage --filter status=failed
No matches found for ['step=failed'].
```

=> I'd say let's not print anything on the terminal in this case.  (Kind of like `docker ps` does not show anything when no container is running.)  This will allow people to use the commands in their own scripts (e.g.if they use `--json` output).

(2) If user provides wrong filter name, we should print error message and then sys.exit(1) so that exit status would indicate failure.

(3) The help messages could be prettified, for example:

```console
$ reana-client logs -w helloworld-yadage --filter steppp
Cannot read filter from input 'steppp'. Please give it in the form key=value, e.g. status=failed
```

=> Better say something like: "Error: please provide complete `--filter name=value` pairs, for example `--filter status=running`".

(4) You can also prettify messages like:

```console
$ reana-client logs -w helloworld-yadage --filter steppp=foo
Filter 'steppp' is not applicable. Applicable filters are ['step', 'compute_backend', 'docker_img', 'status']
```

=> Better said something like: "Error: filter `steppp` is not valid. Available filters are: 'compute_backend', 'docker_img', 'status', 'step'.

reana_client/cli/workflow.py
tiborsimko: Yeah, I think it better expresses that these filters are all *available*, and it is the user who at the end decides which ones will be *applied*.

(In the meantime I also suggested some wording for failure messages; see below.)

reana_client/cli/workflow.py
alintulu: I agree sounds better. Code updated!

reana_client/cli/workflow.py
```diff
@@ -712,11 +712,14 @@ def add_verbose_data_from_response(response, verbose_headers, headers, data):
 @click.option("--json", "json_format", count=True, help="Get output in JSON format.")
 @add_access_token_options
 @click.option(
-    "-s", "--step", "steps", multiple=True, help="Get logs of a specific step."
+    "--filter",
+    "filters",
+    multiple=True,
+    help="Filter the job logs with name=value pairs. Available filters are step, compute_backend, docker_img and status.",
```
tiborsimko, line 718: Please sort alphabetically: c_d, d_i, ste, sta.

reana_client/cli/workflow.py
```diff
@@ -729,22 +732,87 @@ def workflow_logs(ctx, workflow, access_token, json_format, steps=None):  # noqa
...
         logging.debug("{param}: {value}".format(param=p, value=ctx.params[p]))
     if workflow:
+        if filters:
+            try:
+                for f in filters:
+                    key, value = f.split("=")
+                    if key not in available_filters:
+                        click.echo(
+                            click.style(
+                                "Error: filter '{}' is not valid\nAvailable filters are {}".format(
```
tiborsimko, line 760: Still some room for prettification:

```
Available filters are ['step', 'compute_backend', 'docker_img', 'status']
```

could be said:

```
Available filters are 'compute_backend', 'docker_img', 'status', 'step'.
```

Just join the list items into a string instead of printing "naked" Python list.

reana_client/cli/workflow.py
tiborsimko: Hmm, the message "Filter the job logs with name=value pairs" may be a bit cryptic to the end users.  What about making it even more explicit, such as:

```
Filter job logs to include only those steps that match certain filtering criteria. Use --filter name=value pairs. Available...
```

tiborsimko:
Also:

- Can you please add yourself to the `AUTHORS.rst` file?

- Can you please leave a one-line note in `CHANGES.rst` about this change?

reana_client/cli/workflow.py
alintulu: Fixed!

reana_client/cli/workflow.py
alintulu: Fixed!

PR #420, opened 2020-07-13
cli: add pagination fields to workflow list cmd

mvidalgarcia:
closes reanahub/reana-server#180


mvidalgarcia:
Tests not passing due to reana-commons version.

PR #422, opened 2020-08-03
docs: revert cli api docs autogeneration

mvidalgarcia:
- adds link pointing to full docs (readthedocs)

addresses reanahub/reana#335

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -71,3 +71,7 @@ target/
 
 # Visual Studio Code files
 .vscode/*
+
+# Docs autogeneration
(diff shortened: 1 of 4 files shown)


diegodelemos:
Travis is not passing, looks like there is a difference in the produced documentation when it is generated with Python3.

mvidalgarcia:
> Travis is not passing, looks like there is a difference in the produced documentation when it is generated with Python3.

It's because Travis wasn't restarted after merging docs.reana.io PR, now it's passing 🙂

PR #428, opened 2020-10-05
cli: implement filtering by status and search by name

### LINKED ISSUE #427: cli: implement filtering by status and search by name
We already have the [filtering/search logic server-side](https://github.com/reanahub/reana-workflow-controller/blob/master/reana_workflow_controller/rest/workflows.py#L207-L215). The idea will be adding options to `reana-client list` that allows filtering by status and searching by workflow name.
Sorting by creation date is already possible locally via `reana-client list --sort CREATED`. However, if pagination is used, this won't work properly as it only will sort the page itself.

ParthS007:
closes #427

### DIFF
--- a/reana_client/cli/utils.py
+++ b/reana_client/cli/utils.py
@@ -16,7 +16,7 @@ import sys
 import click
 
 from reana_client.utils import workflow_uuid_or_name
-from reana_client.config import ERROR_MESSAGES
+from reana_client.config import ERROR_MESSAGES, RUN_STATUSES
 from reana_commons.errors import MissingAPIClientConfiguration
 
 
@@ -105,11 +105,11 @@ def add_pagination_options(func):
     return wrapper
 
 
-def parse_parameters(_filter):
+def parse_format_parameters(_format):
     """Return parsed filter parameters."""
     try:
         parsed_filters = []
-        filters = " ".join(_filter).replace(",", " ")
+        filters = " ".join(_format).replace(",", " ")
         for item in shlex.split(filters):
             if "=" in item:
                 filter_item = {
@@ -127,7 +127,48 @@ def parse_parameters(_filter):
         )
 
 
-def filter_data(parsed_filters, headers, tablib_data):
+def parse_filter_parameters(filters):
+    """Return parsed filter parameters."""
+    try:
+        status_filters = []
+        search_filters = None
+        filters = " ".join(filters).split(",")
+        for item in filters:
+            if "=" in item:
+                filter_name = item.split("=")[0].lower()
+                filter_value = item.split("=")[1]
+                if filter_name == "status":
+                    if filter_value in RUN_STATUSES:
+                        status_filters.append(filter_value)
+                    else:
+                        click.secho(
+                            "==> ERROR: Input status value {} is not valid. ".format(
+                                filter_value
+                            ),
+                            err=True,
+                            fg="red",
+                        ),
+                        sys.exit(1)
+                elif filter_name == "name":
+                    search_filters = filter_value
+                else:
+                    click.secho(
+                        "==> ERROR: Filter {} is not valid.".format(filter_name),
+                        err=True,
+                        fg="red",
+                    ),
+                    sys.exit(1)
+            else:
+                raise click.BadParameter("Wrong input format")
+        return status_filters, search_filters
+    except ValueError as e:
+        click.echo(
+            click.style("Wrong filter format \n{0}".format(e.message), fg="red"),
+            err=True,
+        )
+
+
+def format_data(parsed_filters, headers, tablib_data):
     """Return filtered data."""
     parsed_filters = [i for i in parsed_filters if i["column_name"] in headers]
     column_headers = [i["column_name"] for i in parsed_filters] or None
--- a/reana_client/cli/workflow.py
+++ b/reana_client/cli/workflow.py
@@ -22,13 +22,14 @@ from reana_client.cli.utils import (
     add_pagination_options,
     add_workflow_option,
     check_connection,
-    filter_data,
+    format_data,
     format_session_uri,
     key_value_to_dict,
-    parse_parameters,
+    parse_format_parameters,
+    parse_filter_parameters,
     validate_workflow_name,
 )
-from reana_client.config import ERROR_MESSAGES, TIMECHECK
+from reana_client.config import ERROR_MESSAGES, TIMECHECK, RUN_STATUSES
 from reana_client.utils import (
     get_reana_yaml_file_path,
     get_workflow_name_and_run_number,
@@ -64,10 +65,10 @@ def workflow_execution_group(ctx):
 )
 @click.option(
     "--format",
-    "_filter",
+    "_format",
     multiple=True,
     help="Format output according to column titles or column values. "
-    "Use `<columm_name>=<column_value>` format. For "
+    "Use `<columm_name>=<column_value>` format. "
     "E.g. display workflow with failed status and named test_workflow "
     "`--format status=failed,name=test_workflow`.",
 )
@@ -107,6 +108,14 @@ def workflow_execution_group(ctx):
     default="CREATED",
     help="Sort the output by specified column",
 )
+@click.option(
+    "--filter",
+    "filters",
+    multiple=True,
+    help="Filter workflow that contains certain filtering criteria. "
+    "Use --filter `<columm_name>=<column_value>` pairs. "
+    "Available filters are `name` and `status`.",
+)
 @add_access_token_options
 @add_pagination_options
 @check_connection
@@ -114,7 +123,7 @@ def workflow_execution_group(ctx):
 def workflow_workflows(
     ctx,
     sessions,
-    _filter,
+    _format,
     output_format,
     access_token,
     show_all,
@@ -123,6 +132,7 @@ def workflow_workflows(
     sort_columm_name,
     page,
     size,
+    filters,
 ):  # noqa: D301
     """List all workflows and sessions.
 
@@ -143,8 +153,13 @@ def workflow_workflows(
     for p in ctx.params:
         logging.debug("{param}: {value}".format(param=p, value=ctx.params[p]))
     type = "interactive" if sessions else "batch"
-    if _filter:
-        parsed_filters = parse_parameters(_filter)
+
+    status_filter = None
+    search_filter = None
+    if filters:
+        status_filter, search_filter = parse_filter_parameters(filters)
+    if _format:
+        parsed_format_filters = parse_format_parameters(_format)
     try:
         if not verbose:
             block_size = None
@@ -155,6 +170,8 @@ def workflow_workflows(
             block_size=block_size,
             page=page,
             size=size,
+            status=status_filter,
+            search=search_filter,
         )
         verbose_headers = ["id", "user", "size"]
         headers = {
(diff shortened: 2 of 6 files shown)


mvidalgarcia (commented):
I think there was a misunderstanding with the issue description and I'm sorry about that. The idea behind the issue was making use of the [recently added args in reana-server](https://github.com/reanahub/reana-workflow-controller/blob/master/reana_workflow_controller/rest/workflows.py#L207-L215). In other words, passing these args (`search` and `status`) directly in [this call](https://github.com/reanahub/reana-client/blob/577e5dd52f14a75bed32863ccbd419e6c2ebac0d/reana_client/api/client.py#L70-L77) so the server does the heavy lifting. @tiborsimko Did you understand the same? Probably the issue wasn't well enough explained...

tiborsimko:
Yes, I understand it the same, but there were two things, actually: (1) adding new filters and (2) amending old format/filter convention to standardise across various `r-client` commands... So both parts need to be addressed...

mvidalgarcia (commented):
I get this:

```console
$ reana-client list
Workflow list could not be retrieved: 
local variable 'status_filter' referenced before assignment
```

Do you get the same?

PS. It seems you need to rebase on top of the latest master.

reana_client/cli/workflow.py
```diff
@@ -143,8 +153,13 @@ def workflow_workflows(
...
     type = "interactive" if sessions else "batch"
-    if _filter:
-        parsed_filters = parse_parameters(_filter)
+
+    status_filter = None
+    search_filter = None
+    if filters:
+        status_filter, search_filter = parse_filter_parameters(filters)
+    if format:
+        parsed_formatting_filters = parse_parameters(format)
```
mvidalgarcia, line 162: What do you think about renaming this function to `parse_format_parameters`? So we have both `parse_filter_parameters` and `parse_format_parameters`, each one corresponding to a different option logic.

reana_client/cli/utils.py
```diff
@@ -127,7 +127,34 @@ def parse_parameters(_filter):
...
+        for item in filters:
+            if "=" in item:
+                filter_name = item.split("=")[0].lower()
+                filter_value = item.split("=")[1]
+                if filter_name == "status":
+                    status_filters.append(filter_value)
+                elif filter_name == "name":
+                    search_filters = filter_value
+                else:
+                    click.echo("Filter {} is not supported yet.".format(filter_name))
```
mvidalgarcia, line 145:
```suggestion
                    click.echo("Filter {} is not supported.".format(filter_name))
```

If someone types `--filter foo=bar`, let's not give them fake hope that we're gonna implement filter `foo` 😉

mvidalgarcia (commented):
**A)**
Can you reproduce this?

```console
$ reana-client list
NAME           RUN_NUMBER   CREATED               STARTED               ENDED                 STATUS  
roofit-fail    1            2020-10-14T10:42:20   2020-10-14T10:42:21   2020-10-14T10:42:38   failed  
roo            1            2020-10-14T10:23:06   2020-10-14T10:23:06   2020-10-14T10:23:30   finished
root6-roofit   1            2020-10-14T10:22:48   2020-10-14T10:22:49   2020-10-14T10:23:13   finished
$ reana-client list --format status=failed,name=test_workflow  
Workflow list could not be retrieved: 
list.remove(x): x not in list
$ reana-client list --format status=failed,name=roofit-fail  
Workflow list could not be retrieved: 
list.remove(x): x not in list
```

I might only happen when there are `failed` workflows (?)

**B)**
I see that we have [`--format` for listing workflow files](https://github.com/reanahub/reana-client/blob/577e5dd52f14a75bed32863ccbd419e6c2ebac0d/reana_client/cli/files.py#L48), could we amend the variable name to be consistent, i.e. name it `_format`?

reana_client/cli/workflow.py
```diff
@@ -143,8 +153,13 @@ def workflow_workflows(
...
         logging.debug("{param}: {value}".format(param=p, value=ctx.params[p]))
     type = "interactive" if sessions else "batch"
-    if _filter:
-        parsed_filters = parse_parameters(_filter)
+
+    status_filter = None
+    search_filter = None
+    if filters:
+        status_filter, search_filter = parse_filter_parameters(filters)
+    if format:
```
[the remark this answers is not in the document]
ParthS007: Done 👍

reana_client/cli/workflow.py
```diff
@@ -64,7 +65,7 @@ def workflow_execution_group(ctx):
 )
 @click.option(
     "--format",
-    "_filter",
+    "format",
     multiple=True,
     help="Format output according to column titles or column values. "
     "Use `<columm_name>=<column_value>` format. For "
```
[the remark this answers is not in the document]
ParthS007: Done 👍

reana_client/cli/workflow.py
```diff
@@ -107,14 +108,22 @@ def workflow_execution_group(ctx):
     default="CREATED",
     help="Sort the output by specified column",
 )
+@click.option(
+    "--filter",
+    "filters",
+    multiple=True,
+    help="Filter workflow that contains certain filtering criteria. "
+    "Use --filter `<columm_name>=<column_value>` pairs. "
+    "Available filters are name and status.",
```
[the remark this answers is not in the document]
ParthS007: Done 👍

reana_client/cli/workflow.py
ParthS007: Done 👍

reana_client/cli/utils.py
ParthS007: Done 👍

reana_client/cli/utils.py
```diff
@@ -127,7 +127,34 @@ def parse_parameters(_filter):
...
+    try:
+        status_filters = []
+        search_filters = None
+        filters = " ".join(filters).split(",")
+        for item in filters:
+            if "=" in item:
+                filter_name = item.split("=")[0].lower()
+                filter_value = item.split("=")[1]
+                if filter_name == "status":
+                    status_filters.append(filter_value)
```
mvidalgarcia, line 141: Maybe we can check here that the provided status is a valid one? Otherwise:

```console
$ reana-client list --filter status=foo
Workflow list could not be retrieved: 
Malformed request.
```

Do you know if we have somewhere in r-commons all the workflow statuses to do this check? cc/ @diegodelemos @tiborsimko

Another option would be returning a more descriptive exception from r-server when the provided status is invalid.

This doesn't necessarily need to be addressed now. Maybe we can create an issue for it.

reana_client/cli/utils.py
```diff
@@ -127,7 +127,34 @@ def parse_parameters(_filter):
...
+                elif filter_name == "name":
+                    search_filters = filter_value
+                else:
+                    click.echo("Filter {} is not supported.".format(filter_name))
```
mvidalgarcia, line 145: Maybe we can add `Error: ` prefix as we have in other places and print it in red, WDYT?

mvidalgarcia (commented):
A couple of more comments, it's looking great 🚀

reana_client/cli/utils.py
ParthS007: @mvidalgarcia  What are the different status possible, I can create a mapping for them?

reana_client/cli/utils.py
ParthS007: I can add but this is not an error, just info for a wrong input. Do we have some color for `info` as well, that would be more apt here?

reana_client/cli/utils.py
tiborsimko: Validation would be very good to do :+1: The list of statuses is living in `reana-db` enum, see `class  WorkflowStatus` there. But we cannot use `r-db` here in `r-client`... So we'd either have to pull it out into to `r-commons`, or simply hard-code here for the MVP... That should do, since they are not changing much as a function of time. So we could simply enrich `r-client/config.py` with the status list for now...

reana_client/cli/utils.py
mvidalgarcia: I don't think so.. we just use white for info messages. Nevermind, it's not very important as long as we inform the user.

reana_client/cli/utils.py
tiborsimko: We can rather say that the filter is not valid.  We are doing it already for the `logs` command:

```console
$ reana-client logs -w helloworld-yadage --filter staaatus=foo
Error: filter 'staaatus' is not valid.
Available filters are 'compute_backend' 'docker_img' 'status' 'step'.
```

So we can reuse something identical here regarding validation. Aiming for the same user experience.

reana_client/cli/utils.py
tiborsimko: Yes, as we do for the `logs` command already:

```console
$ reana-client logs -w helloworld-yadage --filter staaatus=foo
Error: filter 'staaatus' is not valid.
Available filters are 'compute_backend' 'docker_img' 'status' 'step'.
```

We can unify the formatting at least for the various `--filter` commands everywhere.

BTW note that the output message formatting cosmetics will remain elsewhere to be amend as well, for example in `ping` we don't report it either:

```console
$ reana-client ping
REANA client is not connected to any REANA cluster.
```

Something for another issue. We have quite a potential to improve cosmetics regarding `==> ERROR: ...` and friends.

reana_client/cli/utils.py
ParthS007: Done 👍

reana_client/cli/utils.py
ParthS007: Done 👍

reana_client/cli/utils.py
ParthS007: Created a new issue #431

ParthS007:
> **A)**
> Can you reproduce this?
> 
> ```
> $ reana-client list
> NAME           RUN_NUMBER   CREATED               STARTED               ENDED                 STATUS  
> roofit-fail    1            2020-10-14T10:42:20   2020-10-14T10:42:21   2020-10-14T10:42:38   failed  
> roo            1            2020-10-14T10:23:06   2020-10-14T10:23:06   2020-10-14T10:23:30   finished
> root6-roofit   1            2020-10-14T10:22:48   2020-10-14T10:22:49   2020-10-14T10:23:13   finished
> $ reana-client list --format status=failed,name=test_workflow  
> Workflow list could not be retrieved: 
> list.remove(x): x not in list
> $ reana-client list --format status=failed,name=roofit-fail  
> Workflow list could not be retrieved: 
> list.remove(x): x not in list
> ```
> 
> I might only happen when there are `failed` workflows (?)
> 
> **B)**
> I see that we have [`--format` for listing workflow files](https://github.com/reanahub/reana-client/blob/577e5dd52f14a75bed32863ccbd419e6c2ebac0d/reana_client/cli/files.py#L48), could we amend the variable name to be consistent, i.e. name it `_format`?

@mvidalgarcia 

A ) Yes, It is happening for me as well, I can work on it to fix this in a different issue.

B ) Done 👍

reana_client/config.py
```diff
@@ -39,3 +39,14 @@
 
 WORKFLOW_ENGINES = ["serial", "cwl", "yadage"]
 """Supported workflow engines."""
+
+AVAILABLE_STATUS = [
```
audrium, line 43: Minor detail: in this [pr](https://github.com/reanahub/reana-db/pull/90) we have also changed `WorkflowStatus` enum name to `RunStatus`. Maybe it's worth to name it here accordingly just for consistency?

ParthS007: Going for RUN_STATUS , thanks 👍

ParthS007: Done 👍

reana_client/config.py
```diff
@@ -39,3 +39,14 @@
 
 WORKFLOW_ENGINES = ["serial", "cwl", "yadage"]
 """Supported workflow engines."""
+
+RUN_STATUS = [
```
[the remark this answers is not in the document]
ParthS007: Done 👍

PR #437, opened 2020-10-23
CI: Add support for GH Actions

ParthS007:
Latest CI Run: https://github.com/ParthS007/reana-client/actions/runs/333439790


run-tests.sh
```diff
@@ -13,22 +13,67 @@ set -o errexit
...
+    case $arg in
+        --check-shellscript) check_script;;
+        --check-black) check_black;;
+        --check-pydocstyle) check_pydocstyle;;
+        --check-manifest) check_manifest;;
+        --check-cli-cmds) check_cli_cmds;;
+        --check-cli-api) check_cli_api;;
+        --check-sphinx) check_sphinx;;
+        --check-pytest) check_pytest;;
+        --check-all)
```
mvidalgarcia, line 67: Would it be possible to keep a no-args option to run all the tests?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,186 @@
+name: CI
+
+on:
+    push:
+        branches: [master]
+    pull_request:
+        branches: [master]
+
+jobs:
+    Lint-Shellcheck:
```
mvidalgarcia, line 10: Any reason why we call the jobs capitalizing? Is there a convention or something? Also note "Lint-Shellcheck" vs "Lint-black"

codecov.yml
```diff
@@ -0,0 +1,20 @@
+codecov:
```
audrium, line 1: This file could be moved to `.github` folder in order not to pollute the main directory with too many config files

.github/workflows/ci.yml
ParthS007: I don't think there is a convention, just maintaining the sentence case. will do `Lint-shellcheck` 👍

audrium: IMHO these jobs are called in such a way because they appear in the workflow job list like here in the left: https://github.com/reanahub/reana/actions/runs/329122827

Also, it seems there are no strict naming convention for `yaml` files: https://stackoverflow.com/questions/22771226/what-is-the-canonical-yaml-naming-style so I guess we need to select one and use it across the projects

run-tests.sh
ParthS007: yes, adding it. 👍

codecov.yml
ParthS007: Done 👍

.github/workflows/ci.yml
ParthS007: Yeah, we can just group similar kinds of steps together and follow sentence case for the naming. wdyt?

tiborsimko: I would vote for using all lowercase, for simplicity.

P.S. BTW what about making PRs against `maint-0.7`? We shall need to provide support for this branch, so GA there would be handy for longer time.  This seems to be a good example of "make PR to the lowest reasonable `maint-X.Y` branch where the fix may be applicable".)

run-tests.sh
ParthS007: Done 👍

.github/workflows/ci.yml
ParthS007: okay, going with lowercase 👍 .

I am still curious why GH actions don't trigger for my PR 🤔. @audrium Do you see anything that I might be missing?

ParthS007: I can change the base of the PR to `maint-0.7`

audrium: I think it can only be triggered once the workflow file is already in the master branch. I hope so :)

run-tests.sh
diegodelemos: I guess we can remove `--check-all` now that providing no options triggers all checks.

ParthS007: Okay 👍

ParthS007: Done 👍

run-tests.sh
```diff
@@ -13,22 +13,69 @@ set -o errexit
...
+    sphinx-build -qnNW -b doctest docs docs/_build/doctest
+}
+
+check_pytest () {
+    python setup.py test
+}
+
+
+if [ $# -eq 0 ]; then
+    check_script
```
audrium, line 57: Just a thought: maybe it would be cleaner to have another function and call it in case of no arguments?
```bash
check_all() {
    check_script
    check_black
    check_pydocstyle
    check_manifest
    check_cli_cmds
    check_cli_api
    check_sphinx
    check_pytest
}
```
what do you think?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,186 @@
...
+
+        - name: Run Sphinx documentation with doctests
+          run: ./run-tests.sh --check-sphinx
+
+    python-tests:
+        runs-on: ubuntu-latest
+        strategy:
+          matrix:
+              include:
+                - testenv: lowest
```
diegodelemos, line 124: Is there any specific reason for using this approach over `python-version: [...]` as in [Audrius PR](https://github.com/reanahub/reana/pull/433/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR104)? I find it more compact and readable.

ParthS007: Yes, here we have 2 different environments to test in `lowest` and `release` and these envs have to be in each python version.

I tried `python-version` but then it gives priority to it and not `testenv`.

run-tests.sh
ParthS007: Yes, I can do that 👍

ParthS007: Done 👍

audrium:
LGTM :rocket:

.github/workflows/ci.yml
```diff
@@ -0,0 +1,186 @@
+name: CI
+
+on:
```
audrium, line 3:
```suggestion
on: [push, pull_request]
```

Probably it's a good idea not to limit running ci workflows only on a master branch since we have maint-0.* branches as well

.github/workflows/ci.yml
```diff
@@ -0,0 +1,182 @@
...
+          - uses: actions/checkout@v2
+
+          - name: Setup Python
+            uses: actions/setup-python@v2
+            with:
+              python-version: 3.8
+
+          - name: Check Python code formatting
+            run: |
+              pip install black==19.10b0
```
mvidalgarcia, line 28: Crazy idea: Would it be possible to install `black` from [pytest-reana](https://github.com/reanahub/pytest-reana/blob/master/setup.py#L37)? So we can avoid amending all the GitHub action workflow files in case we decide to switch version.

Worst case scenario we can install `pytest-reana` altogether although it seems overkill

tiborsimko: For MVP, since we are blocked by the current CI situation, I'd say that anything will do, whatever is easier to implement :wink:  If we decide to say upgrade `black` version in `r-commons`, we ran run an automated multi-`sed` job to upgrade all the workflow definitions. (However, installing `pytest-reana` is naturally also OK, provided it does not slow things down too much or won't bring other f-string or DB circular import issues and friends.)

diegodelemos: FYI this exact discussion was initially started at https://github.com/reanahub/reana/pull/433#discussion_r512544718.

mvidalgarcia: I had missed that, thanks. Good to see we have similar opinions 😀

.github/workflows/ci.yml
ParthS007: Done 👍

.github/workflows/ci.yml
```diff
@@ -0,0 +1,185 @@
...
+        - name: Setup Python
+          uses: actions/setup-python@v2
+          with:
+            python-version: 3.8
+
+        - name: Install Python dependencies
+          run: pip install check-manifest
+
+        - name: Check Python manifest completeness
+          run: ./run-tests.sh --check-manifest
```
diegodelemos, line 62: I see we are approaching this a bit different than in e.g. [RWC GH Actions PR](https://github.com/reanahub/reana-workflow-controller/pull/347/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR57-R72). Maybe this has been discussed before and I didn't see the outcome of the conversation?

ParthS007: It's the same, I just separated the `installation` in a different step.

diegodelemos (approved):

PR #438, opened 2020-10-25
Merge branch 'maint-0.7'

tiborsimko:


reana_client/version.py
```diff
@@ -14,4 +14,4 @@
 
 from __future__ import absolute_import, print_function
 
-__version__ = "0.7.0a1"
+__version__ = "0.8.0a0"
```
tiborsimko, line 17: Tests for py27 fail because `pytest-reana` brings `reana-db` which uses f-strings which are incompatible with py27. Since we shall be supporting python2.7 for longer for client, what about *not* using f-strings in common modules (r-db, p-reana, r-commons)? 

```console
$ rg 'f"'
reana_db/models.py
161:                        f"resources of resource type {resource_type} use "
524:                    f"{bytes_}B" if block_size == "b" else f"{round(bytes_/1024, 2)}K"
772:                        title=f"Default {type_} resource.",

reana_db/utils.py
226:        raise Exception(f"Default resource of type {resource_type} does not exist.")

reana_db/cli.py
216:            f"Created resources: {[r.name for r in created_resources]}", fg="green"
233:            f"[ERROR]: An error occurred when updating users disk quota usage: {repr(e)}",

tests/conftest.py
56:            name=f"test_{id_}",
```

I guess we can merge this PR and remove f-strings in `r-db` later...

diegodelemos (approved):

run-tests.sh
```diff
@@ -25,7 +25,7 @@ if [ "$python_version" -eq 3 ]
 then
     black --check .
     python scripts/generate_cli_api.py > cli_api.md
-    (diff -q -w  cli_api.md <(curl -s $cli_docs_url) || (echo $docs_differ_error_msg && exit 1))
+    #(diff -q -w  cli_api.md <(curl -s $cli_docs_url) || (echo $docs_differ_error_msg && exit 1))
```
diegodelemos, line 28: Should we create an issue to introduce documentation versioning in docs.reana.io?

PR #442, opened 2020-11-03
cli: enrich ping command with version information

### LINKED ISSUE #441: cli: enrich `ping` command with version information
Currently, the `ping` command returns:

```console
$ reana-client ping
REANA server: https://reana-qa.cern.ch
Authenticated as: John Doe <john.doe@example.org>
Status: Connected
```

Some users were experiencing troubles in the past mixing old REANA client version with new REANA server deployments.

It would be therefore good to output also the client/server version information in the `ping` command, similarly to `kubectl`.

Example:

```console
$ reana-client ping
REANA server: https://reana-qa.cern.ch
REANA server version: 0.8.0a1
REANA client version: 0.7.0
Authenticated as: John Doe <john.doe@example.org>
Status: Connected
```

We should also emit a big warning in case the client and the server versions do not match and are expected to be incompatible, e.g. client/0.6.0 and server/0.7.0.

For now, we may require the same 0.X.* minor version.

If we detect incompatibilities, we could also display `pip` instructions how to upgrade the client, or simply link to the docs.

ParthS007:
Screenshot 2020-11-03 at 17 26 10
closes #441

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,11 @@
 Changes
 =======
 
+Version 0.7.1 (2020-11-10)
+--------------------------
+
+- Changes ``ping`` command output to include REANA client and server version information.
+
 Version 0.7.0 (2020-10-20)
 --------------------------
 
--- a/reana_client/cli/ping.py
+++ b/reana_client/cli/ping.py
@@ -43,10 +43,14 @@ def ping(ctx, access_token):  # noqa: D301
         click.echo(
             click.style(
                 "REANA server: {0}\n"
-                "Authenticated as: {1} <{2}>\n"
-                "Status: {3}".format(
+                "REANA server version: {1}\n"
+                "REANA client version: {2}\n"
+                "Authenticated as: {3} <{4}>\n"
+                "Status: {5}".format(
                     get_api_url(),
-                    response.get("full_name") or "",
+                    response.get("reana_server_version", ""),
+                    __version__,
+                    response.get("full_name", ""),
                     response.get("email"),
                     response.get("status"),
                 ),
--- a/reana_client/version.py
+++ b/reana_client/version.py
@@ -14,4 +14,4 @@ This file is imported by ``reana_client.__init__`` and parsed by
 
 from __future__ import absolute_import, print_function
 
-__version__ = "0.7.0"
+__version__ = "0.7.1"
--- a/setup.py
+++ b/setup.py
@@ -45,7 +45,7 @@ install_requires = [
     "cwltool==1.0.20191022103248",
     "pyOpenSSL>=19.0.0",  # FIXME remove once yadage-schemas solves deps.
     "jsonpointer>=2.0",
-    "reana-commons>=0.7.0,<0.8.0",
+    "reana-commons>=0.7.1,<0.8.0",
     "rfc3987>=1.3.8",  # FIXME remove once yadage-schemas solves deps.
     "six==1.12.0",
     "strict-rfc3339>=0.7",  # FIXME remove once yadage-schemas solves deps.


reana_client/cli/ping.py
```diff
@@ -43,9 +43,13 @@ def ping(ctx, access_token):  # noqa: D301
...
             click.style(
                 "REANA server: {0}\n"
-                "Authenticated as: {1} <{2}>\n"
-                "Status: {3}".format(
+                "REANA server version: {1}\n"
+                "REANA client version: {2}\n"
+                "Authenticated as: {3} <{4}>\n"
+                "Status: {5}".format(
                     get_api_url(),
+                    response.get("reana_server_version"),
```
tiborsimko, line 51: What if server does not contain `reana_server_version`?  Note that this PR must make sure that newly released reana-client 0.7.1 would work with previously released REANA cluster 0.7.0 which may not contain the new field. So we have to be very liberal in what server may send.  Kind of:

```
if response contains reana_server_version:
then
   print reana_server_version
else
  do nothing
endif
```

P.S. Please also rebase against latest `maint-0.7`.

ParthS007: Done 👍

reana_client/cli/ping.py
```diff
@@ -43,9 +43,13 @@ def ping(ctx, access_token):  # noqa: D301
...
             click.style(
                 "REANA server: {0}\n"
-                "Authenticated as: {1} <{2}>\n"
-                "Status: {3}".format(
+                "REANA server version: {1}\n"
+                "REANA client version: {2}\n"
+                "Authenticated as: {3} <{4}>\n"
+                "Status: {5}".format(
                     get_api_url(),
+                    response.get("reana_server_version") or "",
```
mvidalgarcia, line 51:
```suggestion
                    response.get("reana_server_version", ""),
```

ParthS007: @mvidalgarcia 

I added `""` after this comment https://github.com/reanahub/reana-client/pull/442#discussion_r518647376

If this doesn't address the comment, I can remove it.

mvidalgarcia: It's not removing the `""` but making it the default value if `"reana_server_version"` doesn't exist.
My point is that you don't need the `or`, both expressions are equivalent.

ParthS007: Yes, got it. Thanks for pointing it out, I corrected one more similar thing.

diegodelemos (approved):

CHANGES.rst
```diff
@@ -1,6 +1,13 @@
 Changes
 =======
 
+Version 0.7.1 (2020-11-09)
+--------------------------
+
+- Adds support for ``htcondor_max_runtime`` and ``htcondor_accounting_group`` in HTC jobs.
+- Adds support for restarting yadage workflows (through ``accept_metadir`` operational option).
+- Adds REANA server version to ``reana-client ping`` output.
```
tiborsimko, line 9: WRT first point: HTCondor instead of HTC, see the usual phrasing:

- Adds support for specifying ``htcondor_max_runtime`` and ``htcondor_accounting_group`` for HTCondor compute backend jobs.

WRT second point, Yadage with capital Y.  But shouldn't we say CWL and Yadage, i.e. both?  Also, I would say `Fixes restarting..." because this is more of a bug fix, which is why we put it to 0.7.1. So:

- Fixes restarting of CWL and Yadage workflows.

WRT third point, perhaps we should say: "Changes..." following the same feature/improvement/bugfix logic, so:

- Changes `ping` command output to include REANA client and server version information.

However, the first two points are not really thanks to the changes in the `r-client` itself, more thanks to the changes in the cluster. So we should probably leave them out from here and put them into the main repository `reana`'s release notes. Which would leave us with the only point about ``ping`` command.

PR #443, opened 2020-11-04
cli: check htcondor_max_runtime

alintulu:
People submitting e.g. `htcondor_max_runtime = epreso` (or any other typo) will see that their workflow runs but it will eventually specify no `JobFlavour` (it will use the default `MaxRunTime`) which is not what they meant.

If we move this check before in time and we make it more strict (e.g. if string check if it is in the list otherwise fail), we can be more correct and avoid unnecessary use of resources:
- If we do it in REANA-Client:
  - We avoid doing a call to the REANA cluster before the check passes
  - We give faster feedback to users (instead of waiting for minutes to realise that what they wanted didn't happen)

_Originally posted by @diegodelemos in https://github.com/reanahub/reana-job-controller/pull/278#discussion_r517156957_

closes reanahub/reana-job-controller#32

### DIFF
--- a/reana_client/cli/workflow.py
+++ b/reana_client/cli/workflow.py
@@ -34,6 +34,7 @@ from reana_client.cli.utils import (
     parse_filter_parameters,
     parse_format_parameters,
     validate_workflow_name,
+    check_htcondor_max_runtime,
 )
 from reana_client.config import ERROR_MESSAGES, RUN_STATUSES, TIMECHECK
 from reana_client.utils import (
@@ -298,6 +299,9 @@ def workflow_create(ctx, file, name, skip_validation, access_token):  # noqa: D3
         reana_specification = load_reana_spec(
             click.format_filename(file), skip_validation
         )
+        click.echo(reana_specification)
+        if not check_htcondor_max_runtime(reana_specification):
+            return Exception("Invalid input in htcondor_max_runtime.")
         logging.info("Connecting to {0}".format(get_api_url()))
         response = create_workflow(reana_specification, name, access_token)
         click.echo(click.style(response["workflow_name"], fg="green"))
--- a/reana_client/cli/utils.py
+++ b/reana_client/cli/utils.py
@@ -18,6 +18,7 @@ import click
 from reana_client.utils import workflow_uuid_or_name
 from reana_client.config import ERROR_MESSAGES, RUN_STATUSES
 from reana_commons.errors import MissingAPIClientConfiguration
+from reana_commons.config import HTCONDOR_JOB_FLAVOURS
 
 
 def add_access_token_options(func):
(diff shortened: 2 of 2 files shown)


alintulu:
@diegodelemos Since `compute_backend` is set for every step independently, the `htcondor_max_runtime` has to be checked for every step, for every workflow. Every time a user submits a workflow it goes through all steps, collects the ones that has  `compute_backend: htcondor` and `htcondor_max_runtime: something`. And further checks them. I am afraid this is too costly to do on the client? At the same time if I do it on r-server the `reana-client create` command would anyway have to send a request to the server to check it, and wait to get the result back. Not sure what is better.

alintulu:
Example output:

```
$ reana-client create -w htcondor-workflow
In step gendata:
'espr' is not a valid input for htcondor_max_runtime. Inputs must be a digit in the form of a string, or one of the following job flavours: 'espresso' 'microcentury' 'longlunch' 'workday' 'tomorrow' 'testmatch' 'nextweek'.
In step fitdata:
'micro' is not a valid input for htcondor_max_runtime. Inputs must be a digit in the form of a string, or one of the following job flavours: 'espresso' 'microcentury' 'longlunch' 'workday' 'tomorrow' 'testmatch' 'nextweek'.
Cannot create workflow htcondor-workflow: 
Invalid input for htcondor_max_runtime
```

reana_client/cli/workflow.py
```diff
@@ -298,6 +299,9 @@ def workflow_create(ctx, file, name, skip_validation, access_token):  # noqa: D3
         reana_specification = load_reana_spec(
             click.format_filename(file), skip_validation
         )
+        click.echo(reana_specification)
+        if not check_htcondor_max_runtime(reana_specification):
```
diegodelemos, line 303: Since this implementation is Serial specific, what do you think about moving the check to the [`load_serial` function](https://github.com/reanahub/reana-commons/blob/34ddc1ef866b0f54eba1789153f8001d107d234b/reana_commons/serial.py#L82)?

REANA-Client will eventually execute it through [`load_reana_spec` function](https://github.com/reanahub/reana-client/blob/2436dde2df70ca835afcc49ad81f1f91ce79e030/reana_client/utils.py#L112) which will perform a validation of the [`reana.yaml`](https://github.com/reanahub/reana-client/blob/2436dde2df70ca835afcc49ad81f1f91ce79e030/reana_client/utils.py#L149) in general and [the workflow specifications in particular depending on its type](https://github.com/reanahub/reana-client/blob/2436dde2df70ca835afcc49ad81f1f91ce79e030/reana_client/utils.py#L152-L156).

diegodelemos:
> Is this too costly to do in the client? At the same time if I do it on r-server the reana-client create command would anyway have to send a request to the server to check it, and wait to get the result back. Not sure what is better.

Looks great in my opinion. I think the loop will be much faster than making the request.

reana_client/cli/workflow.py
diegodelemos: Also, this means that we would be missing this check for Yadage and CWL, for that we can create a couple of issues to not overload this PR group?

alintulu: Ooh of course this will only work for Serial! These values are not set in the reana specifications for CWL and Yadage.

alintulu: > Also, this means that we would be missing this check for Yadage and CWL, for that we can create a couple of issues to not overload this PR group?

Yes I agree.

alintulu: Closed this PR as the function has been moved to r-commons with this commit reanahub/reana-commons@985f71f9a57fa79fe9ec4a538d79630bf7b8eae9.

PR #445, opened 2020-11-12
cli: fix exception handling on file upload

mvidalgarcia:
For compatibility reasons between cluster 0.8.x and client 0.7.x
When quota is already exceeded:

**Before:**
```console
$ reana-client upload -w helloworld LICENSE                                                                                                                                                                              
File /LICENSE was successfully uploaded.       
```
_(even though the file is not uploaded)_

**After:**
```console
$ reana-client upload -w helloworld 800Kfile.dat
Something went wrong while uploading /Users/marco/code/reanahub/reana-demo-helloworld/800Kfile.dat: 
User quota exceeded.
```

### DIFF
--- a/reana_client/api/client.py
+++ b/reana_client/api/client.py
@@ -289,7 +289,9 @@ def upload_file(workflow_id, file_, file_name, access_token):
             headers={"Content-Type": "application/octet-stream"},
             verify=False,
         )
-        return http_response.json()
+        if http_response.ok:
+            return http_response.json()
+        raise Exception(http_response.json().get("message"))
     except requests.exceptions.ConnectionError:
         logging.debug("File could not be uploaded.", exc_info=True)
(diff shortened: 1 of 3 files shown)


reana_client/api/client.py
```diff
@@ -289,7 +289,9 @@ def upload_file(workflow_id, file_, file_name, access_token):
             headers={"Content-Type": "application/octet-stream"},
             verify=False,
         )
-        return http_response.json()
+        if http_response.ok:
+            return http_response.json()
+        raise Exception(http_response.json().get("message"))
```
diegodelemos, line 294: Tested with https://github.com/reanahub/reana-workflow-controller/issues/339 all working fine. Shall we add an entry in changelog?

mvidalgarcia: Added.

diegodelemos (approved):

PR #248, opened 2019-02-11
cli: show link to protected interactive session

diegodelemos:
Connects reanahub/reana-workflow-controller#184.

PR #459, opened 2021-02-23
validate: warn about using non-tagged latest images

ParthS007:
addresses reanahub/reana#248

### DIFF
--- a/reana_client/cli/workflow.py
+++ b/reana_client/cli/workflow.py
@@ -861,8 +861,15 @@ def workflow_logs(
     help="REANA specification file describing the workflow to "
     "execute. [default=reana.yaml]",
 )
+@click.option(
+    "--environments",
+    is_flag=True,
+    default=False,
+    help="If set, check all runtime environments specified in REANA "
+    "specification file. [default=False]",
+)
 @click.pass_context
-def workflow_validate(ctx, file):  # noqa: D301
+def workflow_validate(ctx, file, environments):  # noqa: D301
     """Validate workflow specification file.
 
     The `validate` command allows to check syntax and validate the reana.yaml
@@ -875,7 +882,9 @@ def workflow_validate(ctx, file):  # noqa: D301
     for p in ctx.params:
         logging.debug("{param}: {value}".format(param=p, value=ctx.params[p]))
     try:
-        load_reana_spec(click.format_filename(file))
+        load_reana_spec(
+            click.format_filename(file), skip_validate_environments=not environments
+        )
         click.echo(
             click.style(
                 "File {filename} is a valid REANA specification file.".format(
--- a/reana_client/config.py
+++ b/reana_client/config.py
@@ -40,5 +40,5 @@ URL = "url"
 WORKFLOW_ENGINES = ["serial", "cwl", "yadage"]
 """Supported workflow engines."""
 
-ENVIRONMENT_IMAGE_SUSPECTED_TAGS_VALIDATOR = ["latest", "master"]
+ENVIRONMENT_IMAGE_SUSPECTED_TAGS_VALIDATOR = ["latest", "master", ""]
 """Warns user if above environment image tags are used."""
--- a/reana_client/config.py
+++ b/reana_client/config.py
@@ -39,3 +39,6 @@ URL = "url"
 
 WORKFLOW_ENGINES = ["serial", "cwl", "yadage"]
 """Supported workflow engines."""
+
+ENVIRONMENT_IMAGE_SUSPECTED_TAGS_VALIDATOR = ["latest", "master"]
+"""Warns user if above environment image tags are used."""
--- a/reana_client/utils.py
+++ b/reana_client/utils.py
@@ -207,38 +207,33 @@ def _validate_serial_workflow_environment(workflow_steps):
     :raises Warning: Warns user if the workflow environment is invalid in serial workflow steps.
     """
     for step in workflow_steps:
-        if step.get("environment", None):
-            if ":" in step["environment"]:
-                environment = step["environment"].split(":", 1)
-                environment_image, environment_image_tag = (
-                    environment[0],
-                    environment[-1],
-                )
-                if ":" in environment_image_tag:
-                    logging.warning(
-                        "Image {} has invalid tag {}".format(
-                            environment_image, environment_image_tag
-                        )
+        if ":" in step["environment"]:
+            environment = step["environment"].split(":", 1)
+            environment_image, environment_image_tag = environment[0], environment[-1]
+            if ":" in environment_image_tag:
+                logging.warning(
+                    "Image {} has invalid tag {}".format(
+                        environment_image, environment_image_tag
                     )
-                    sys.exit(1)
-                elif (
-                    environment_image_tag in ENVIRONMENT_IMAGE_SUSPECTED_TAGS_VALIDATOR
-                ):
-                    logging.warning(
-                        "Using {} is not recommended in {} image.".format(
-                            environment_image_tag, environment_image
-                        )
-                    )
-                    sys.exit(1)
-            click.echo(
-                click.style(
-                    "Image {} has correct format.".format(step["environment"]),
-                    fg="green",
                 )
-            )
+                sys.exit(1)
+            elif environment_image_tag in ENVIRONMENT_IMAGE_SUSPECTED_TAGS_VALIDATOR:
+                logging.warning(
+                    "Using '{}' tag is not recommended in {} image.".format(
+                        environment_image_tag, environment_image
+                    )
+                )
+                sys.exit(1)
         else:
-            logging.warning("Workflow step does not have explicit enviornment.")
+            logging.warning(
+                "Image {} does not have an explicit tag.".format(step["environment"])
+            )
             sys.exit(1)
+        click.echo(
+            click.style(
+                "Image {} has correct format.".format(step["environment"]), fg="green",
+            )
+        )
 
 
 def _validate_reana_yaml(reana_yaml):
--- a/reana_client/utils.py
+++ b/reana_client/utils.py
@@ -19,6 +19,7 @@ import click
 import yadageschemas
 import yaml
 from jsonschema import ValidationError, validate
+from reana_client.config import ENVIRONMENT_IMAGE_SUSPECTED_TAGS_VALIDATOR
 from reana_commons.errors import REANAValidationError
 from reana_commons.operational_options import validate_operational_options
 from reana_commons.serial import serial_load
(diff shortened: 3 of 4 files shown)


ParthS007:
I will be adding `cwl` and `yadage` validation soon. :)

mvidalgarcia:
> closes #248

The "Closes" clause is pointing to a PR. Should be reanahub/reana#248.

reana_client/utils.py
```diff
@@ -173,6 +180,49 @@ def _prepare_kwargs(reana_yaml):
...
+
+    :param workflow_steps: List of dictionary which represents different steps involved in workflow.
+    :raises Warning: Warns user if valid tags are not used in serial workflow steps.
+    """
+    for step in workflow_steps:
+        environment = step["environment"].split(":")
+        environment_tag = environment[-1]
+        if len(environment) == 1:
+            logging.warning(
+                "Include a tag in the enviornment {}".format(environment[0])
```
mvidalgarcia, line 213:
```suggestion
                "Include a tag in the environment {}".format(environment[0])
```

reana_client/utils.py
```diff
@@ -173,6 +180,49 @@ def _prepare_kwargs(reana_yaml):
...
+            )
+            sys.exit(1)
+        elif len(environment) == 2 and environment_tag in WARNING_TAGS:
```
mvidalgarcia, line 216: I think that if you pass something like `foo:bar:baz` it returns `True` as its length is not 1 nor 2 (untested).

reana_client/utils.py
```diff
@@ -173,6 +180,49 @@ def _prepare_kwargs(reana_yaml):
...
+            logging.warning(
+                "Using {} in {} is not recommended.".format(
+                    environment_tag, environment[0]
+                )
+            )
+            sys.exit(1)
+    return True
```
mvidalgarcia, line 223: This value is not being used, should we return a boolean? If so, maybe we shouldn't stop the execution (`sys.exit(1)`) and return `False` when the validation fails? cc/ @tiborsimko

reana_client/utils.py
```diff
@@ -173,6 +180,49 @@ def _prepare_kwargs(reana_yaml):
...
+    elif workflow_type == "cwl":
+        logging.info("Not Implemented yet.")
+
+    return True
+
+
+def _validate_serial_workflow_environment(workflow_steps):
+    """Validate environments in REANA serial workflow.
+
+    :param workflow_steps: List of dictionary which represents different steps involved in workflow.
```
mvidalgarcia, line 205:
```suggestion
    :param workflow_steps: List of dictionaries which represents different steps involved in workflow.
```

reana_client/cli/workflow.py
```diff
@@ -874,8 +874,16 @@ def workflow_logs(
     help="REANA specification file describing the workflow to "
     "execute. [default=reana.yaml]",
 )
+@click.option(
+    "--environments",
+    is_flag=True,
+    default=False,
+    help="Check all runtime environments specified in REANA "
+    "specification file describing the workflow to "
+    "execute.",
```
mvidalgarcia, line 883: Perhaps we don't need the "describing the workflow to execute" part? It seems a bit redundant.

mvidalgarcia (commented):
The commit message mentions `--environment` while the option is `--environments`.

reana_client/utils.py
ParthS007: I can do like this:

```
if len()>2 or len()==0
  Warning: Invalid environment format. Use x:y
elif len() == 1:
  Warning: Include a tag
elif len() == 2 and environment_tag in WARNING_TAGS:
  Warning: Invalid tag used
```

wdyt?

reana_client/utils.py
ParthS007: Boolean is not required, I can omit return. 👍

reana_client/utils.py
tiborsimko: The format without `:`, for example `myorg/myimage`, is perfectly valid, so please beware about wording.

Regarding implementation for `foo:bar:baz:fuux:quux`, you can do upfront something like `split(":", 1)` to have only two elements, the image name and the image tag, and then check whether the detected tag contains more `:` inside. That might yield simpler code logic.

reana_client/cli/workflow.py
```diff
@@ -888,7 +896,7 @@ def workflow_validate(ctx, file):  # noqa: D301
     for p in ctx.params:
         logging.debug("{param}: {value}".format(param=p, value=ctx.params[p]))
     try:
-        load_reana_spec(click.format_filename(file))
+        load_reana_spec(click.format_filename(file), environments)
```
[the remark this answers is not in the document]
ParthS007: The validation is working inside this `load_reana_spec()`. 

I can avoid it but I will duplicating the code for reading the `yaml` file and loading the specs if it is okay?

reana_client/config.py
```diff
@@ -50,3 +50,6 @@
     "queued",
 ]
 """Available run statuses."""
+
+WARNING_TAGS = ["latest", "master"]
+"""Warns user if above tags are used."""
```
tiborsimko, line 55: Please scope this variable more, since the "tags" concept could in theory refer to Git tags. For example `VALIDATOR_ENVIRONMENT_IMAGE_SUSPECTED_TAGS` or some such.

reana_client/utils.py
```diff
@@ -173,6 +180,49 @@ def _prepare_kwargs(reana_yaml):
...
+    :param reana_yaml: Dictionary which represents REANA specifications file.
+    :return: True if environment tags are valid.
+    """
+    workflow_type = reana_yaml["workflow"]["type"]
+
+    if workflow_type == "serial":
+        workflow_steps = reana_yaml["workflow"]["specification"]["steps"]
+        _validate_serial_workflow_environment(workflow_steps)
+    elif workflow_type == "yadage":
+        logging.info("Not Implemented yet.")
```
tiborsimko, line 195: The logging info would not be visible to the user unless they raise the loglevel... which is not probable.

You can simply use `warning`, as it it were a finding, just to emit the message saying "The environment validation is not implemented for Yadage workflows yet."

This would be better, otherwise people may be mislead into a false security that the images were checked and not found problematic.

reana_client/utils.py
tiborsimko, line 223: Also, when everything passed, shall we print out a message at the end that things are fine?

Otherwise currently people see:

```console
$ reana-client validate --environments                       
File .../reana-demo-root6-roofit/reana.yaml is a valid REANA specification file.
```

so they won't have any explicit output about whether image checking passed.

I guess we can emit a final message that environment images were found to be OK.  "Explicit is better than implicit", as per the Zen of Python :wink: Since we print one validation message already, we could print more for each additional image check.

P.S. This consideration may lead to a configurable "verbosity" level option, for those people that would like to go quiet, in a kind of linter style. But we could address this in a later PR...

tiborsimko: We shall probably refactor the exit-or-not strategy soon, after we shall introduce those additional tests for environment images.

reana_client/cli/workflow.py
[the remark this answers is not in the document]
tiborsimko: ... or just make the new argument optional, so that we don't have to amend existing `load_reana_spec()` callers? Kind of like `skip_validaton_of_environments=True`? A bit like `skip_validation=False` is used... So if you add a new option at the end, and it will be optional, the branch could still go into `maint-0.7` I think.  We can expand this later more nicely to cover all the various validation scenarios that are to be skipped...

reana_client/cli/workflow.py
mvidalgarcia: We can add "[default=False]", as the `--file` option.

reana_client/utils.py
```diff
@@ -147,6 +148,12 @@ def _prepare_kwargs(reana_yaml):
                 )
             )
             _validate_reana_yaml(reana_yaml)
+        if environments:
+            logging.info(
+                "Validating environments in REANA specification file: "
+                "{filepath}".format(filepath=filepath)
+            )
+            _validate_environment_tags(reana_yaml)
```
mvidalgarcia, line 156: This function can be simply called `_validate_environments` as it will include image existence and uid/gid validation.

reana_client/utils.py
```diff
@@ -173,6 +180,49 @@ def _prepare_kwargs(reana_yaml):
         raise e
 
 
+def _validate_environment_tags(reana_yaml):
+    """Validate environments in REANA specification file according to workflow type.
+
+    :param reana_yaml: Dictionary which represents REANA specifications file.
+    :return: True if environment tags are valid.
```
mvidalgarcia, line 187:
```suggestion
    :return: True if environment is valid.
```

ParthS007: I removed the `return`, we don't need this now. Removed it.

reana_client/utils.py
ParthS007: Added the message after check of every image tag.

reana_client/utils.py
```diff
@@ -173,6 +180,57 @@ def _prepare_kwargs(reana_yaml):
...
+        logging.warning(
+            "The environment validation is not implemented for CWL workflows yet."
+        )
+
+
+def _validate_serial_workflow_environment(workflow_steps):
+    """Validate environments in REANA serial workflow.
+
+    :param workflow_steps: List of dictionaries which represents different steps involved in workflow.
+    :raises Warning: Warns user if valid tags are not used in serial workflow steps.
```
mvidalgarcia, line 207: This function will perform more validations so we can just have: "Warns user if the workflow environment is invalid in serial workflow steps."

reana_client/utils.py
```diff
@@ -109,7 +110,7 @@ def load_workflow_spec(workflow_type, workflow_file, **kwargs):
     return workflow_load[workflow_type](workflow_file, **kwargs)
 
 
-def load_reana_spec(filepath, skip_validation=False):
+def load_reana_spec(filepath, skip_validation=False, skip_validate_environments=False):
```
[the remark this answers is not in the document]
mvidalgarcia: Not addressed, `skip_validate_environments` should be True by default.

reana_client/utils.py
```diff
@@ -173,6 +180,67 @@ def _prepare_kwargs(reana_yaml):
...
+
+
+def _validate_serial_workflow_environment(workflow_steps):
+    """Validate environments in REANA serial workflow.
+
+    :param workflow_steps: List of dictionaries which represents different steps involved in workflow.
+    :raises Warning: Warns user if the workflow environment is invalid in serial workflow steps.
+    """
+    for step in workflow_steps:
+        if step.get("environment", None):
```
mvidalgarcia, line 210: It should check if it has no tag, instead of no environment. For instance, check `reana-demo-lhcb-d2pimumu` example that has the environment [`reanahub/reana-env-root6`](https://github.com/reanahub/reana-demo-lhcb-d2pimumu/blob/7c008deb1974f25080a663711f530a0a515dda4f/reana.yaml#L16) (no tag included). In this case it should raise the warning.

mvidalgarcia: Fixed in https://github.com/reanahub/reana-client/pull/459/commits/0b466c29bd6aadefba8a8afdd98a931a76b7cbaf

mvidalgarcia (approved):

PR #461, opened 2021-02-23
validate: uid/gid environment validation for serial

### LINKED ISSUE #382: spec: validate specified docker images
It has happened to have user workflows stuck with docker image related errors in production (see https://github.com/reanahub/reana-job-controller/issues/229 and https://github.com/reanahub/reana-job-controller/issues/3).

We could fix this by parsing and validating that all specified images have a correct name and other verifications such as if they can be pulled. This will avoid unnecessary waste of resources on the server-side.

### LINKED ISSUE #457: validate: verify UID/GID for workflow environment images
**Current behaviour**

Some workflows are forcing certain Kubernetes user IDs, for example in `reana-demo-atlas-recast`:

```
$ cat workflow/steps.yml
...
  environment:
    environment_type: 'docker-encapsulated'
    image: reanahub/reana-demo-atlas-recast-eventselection
    imagetag: '1.0'
    resources:
      - kubernetes_uid: 500
```

This can become troublesome if the image is not using the same UID by default, or if GID is not 0.  If either of these is not matched, then the job may not have write rights, and it fails.

If this happens during late in the execution, this is basically waisting time.

It would be good to catch these situations faster, i.e. at the `reana-client validate --environment` time.

Currently, we verify this manually as follows:

```console
$ docker pull reanahub/reana-demo-atlas-recast-eventselection:1.0
$ docker run -i -t --rm reanahub/reana-demo-atlas-recast-eventselection:1.0 /usr/bin/id
uid=500(atlas) gid=500(atlas) groups=500(atlas),0(root)
```

This is good for two reasons: UID matches to what is declared in steps yaml; and even if it would not, then GID is 0, ensuring write permissions.

**Expected behaviour**

After we introduce `validate --environment` flags, see e.g. https://github.com/reanahub/reana-client/issues/382#issuecomment-781976172, we could enrich it to catch these situations.

Ideally, verify UID/GID without downloading image, but I think that is not easily possible... So this should be an extra option.

Perhaps using `singularity` or `podman` could help... The same considerations as in the base issue.

Basically, this issue would take the validation even further -- not only validating image existence, but also its UID and GID setup.

mvidalgarcia:
closes #457 
closes #382

#### To test:
- To run: `reana-client validate --environments`
- Choose a `reana-demo-*` and play removing environment tags, adding invalid ones, etc.
- Add `kubernetes_uid` to remove warnings for e.g. ATLAS images.

### DIFF
--- a/reana_client/utils.py
+++ b/reana_client/utils.py
@@ -207,31 +207,46 @@ def _validate_serial_workflow_environment(workflow_steps):
     :raises Warning: Warns user if the workflow environment is invalid in serial workflow steps.
     """
     for step in workflow_steps:
-        if ":" in step["environment"]:
-            environment = step["environment"].split(":", 1)
-            environment_image, environment_image_tag = environment[0], environment[-1]
-            if ":" in environment_image_tag:
-                logging.warning(
-                    "Image {} has invalid tag {}".format(
-                        environment_image, environment_image_tag
-                    )
-                )
-                sys.exit(1)
-            elif environment_image_tag in ENVIRONMENT_IMAGE_SUSPECTED_TAGS_VALIDATOR:
-                logging.warning(
-                    "Using '{}' tag is not recommended in {} image.".format(
-                        environment_image_tag, environment_image
-                    )
-                )
-                sys.exit(1)
-        else:
-            logging.warning(
-                "Image {} does not have an explicit tag.".format(step["environment"])
+        _validate_image_tag(step["environment"])
+
+
+def _validate_image_tag(image):
+    """Validate if image tag is valid."""
+
+    has_warnings = False
+    if ":" in image:
+        environment = image.split(":", 1)
+        image_name, image_tag = environment[0], environment[-1]
+        if ":" in image_tag:
+            click.secho(
+                "==> ERROR: Environment image {} has invalid tag '{}'".format(
+                    image_name, image_tag
+                ),
+                err=True,
+                fg="red",
             )
             sys.exit(1)
+        elif image_tag in ENVIRONMENT_IMAGE_SUSPECTED_TAGS_VALIDATOR:
+            click.secho(
+                "==> WARNING: Using '{}' tag is not recommended in {} environment image.".format(
+                    image_tag, image_name
+                ),
+                fg="yellow",
+            )
+            has_warnings = True
+    else:
+        click.secho(
+            "==> WARNING: Environment image {} does not have an explicit tag.".format(
+                image
+            ),
+            fg="yellow",
+        )
+        has_warnings = True
+    if not has_warnings:
         click.echo(
             click.style(
-                "Image {} has correct format.".format(step["environment"]), fg="green",
+                "==> Environment image {} has correct format.".format(image),
+                fg="green",
             )
         )
 
--- a/reana_client/utils.py
+++ b/reana_client/utils.py
@@ -7,6 +7,7 @@
 # under the terms of the MIT License; see LICENSE file for more details.
 """REANA client utils."""
 import base64
+import datetime
 import json
 import logging
 import os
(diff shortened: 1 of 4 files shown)


tests/test_validate_environment.py
```diff
@@ -0,0 +1,37 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of REANA.
+# Copyright (C) 2021 CERN.
+#
+# REANA is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
```
tiborsimko, line 7: Let us rename to `test_validate_environments` (plural) for analogy with `validate --environments` CLI option.

reana_client/utils.py
```diff
@@ -207,32 +207,42 @@ def _validate_serial_workflow_environment(workflow_steps):
...
+def _validate_image_tag(image):
+    """Validate if image tag is valid."""
+
+    has_warnings = False
+    if ":" in image:
+        environment = image.split(":", 1)
+        image_name, image_tag = environment[0], environment[-1]
+        if ":" in image_tag:
+            click.secho(
+                "==> ERROR: Image {} has invalid tag '{}'".format(
```
tiborsimko, line 222: Shall we use everywhere in the output strings the expression "environment image", or even simply "environment", for better analogy with the `--environment` option name?

reana_client/utils.py
```diff
@@ -207,31 +207,46 @@ def _validate_serial_workflow_environment(workflow_steps):
...
-                    )
-                )
-                sys.exit(1)
-        else:
-            logging.warning(
-                "Image {} does not have an explicit tag.".format(step["environment"])
+        _validate_image_tag(step["environment"])
+
+
+def _validate_image_tag(image):
```
ParthS007, line 213: As far as I understand and according to this [comment](https://github.com/reanahub/reana/issues/248#issuecomment-784004662), all the workflows have image tag in different places. 

Naming can go like this `_validate_serial_image_tag`.

mvidalgarcia: This function works regardless of the workflow engine. You pass an image in the format `foo:bar` and it should do the proper validation, not matter if it's serial, Yadage or CWL.

The difference among workflow engines resides in where to find that image string.

ParthS007: Okay, Sounds good. Nice approach.

In `yadage` the `image_name` and `image_tag` are not together. To use this function, I will combine them and then pass it to this function 👍

mvidalgarcia: Good point, so maybe we can create yet another function that extracts the image name and tag, and depending on the workflow engine it searches for it in one place or another. Then we can pass these two variables to the different validation functions.

ParthS007: Since we have `image_name` and `image_tag` inn different places in workflows engines. I am assuming that the new function will also be also be checking the workflow type and then extracting the image name and tag from the dict.

If we will not use them anywhere else extracting them just before we call `_validate_image_tag` would be enough else creating a function that will return image(for example: `x:y`) will be useful.

ParthS007 (commented):
Thanks for the PR. 
Tested locally, works great. Left some comments.

reana_client/utils.py
```diff
@@ -207,32 +213,133 @@ def _validate_serial_workflow_environment(workflow_steps):
...
+
+
+def _image_exists(image, tag):
+    """Verify if image exists."""
+    DOCKER_REGISTRY = "https://index.docker.io/v1/repositories/{image}/tags/{tag}"
+    docker_registry_url = DOCKER_REGISTRY.format(image=image, tag=tag)
+    # Remove traling slash if no tag was specified
+    if not tag:
+        docker_registry_url = docker_registry_url[:-1]
+    response = requests.get(docker_registry_url)
```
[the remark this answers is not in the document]
mvidalgarcia: Exceptions will be caught already by `load_reana_spec`, is it worth adding another try-except block?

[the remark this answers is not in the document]
ParthS007: From the user point of view, It would be nice to see if API is failing to respond since it will be an external error and not on reana side but at the same time it's okay for me to just use common exception from `load_reana_spec`

ParthS007 (approved):
LGTM 👍

Issue #396, opened 2020-03-01
yadage: make initdir configurable


danikam:
For workflows running over local (rather than downloadable) input, such as the [current monojet workflow](https://gitlab.cern.ch/damacdon/workflow_monojet) (see [run instructions](reana_workflow_engine_yadage/cli.py#L104)), the yadage `-d initdir=` option is currently hardcoded in [reana_workflow_engine_yadage/cli.py#L104](https://github.com/reanahub/reana-workflow-engine-yadage/blob/f5e88cb186428b4c885f0717b6739758a1d2c436/reana_workflow_engine_yadage/cli.py#L104) (thanks to @tiborsimko for identifying this). Could this become configurable in reana.yaml?

mvidalgarcia:
Hi @danikam I don't have access to the links you posted:
- https://gitlab.cern.ch/atlas-phys/exot/jdm/ANA-EXOT-2018-06/workflow
- https://github.com/reanahub/reana-client/issues/reana_workflow_engine_yadage/cli.py#L104

Could you update them or paste an example?

danikam:
Hi @mvidalgarcia, sorry about that and thanks for letting me know! 

I copied the monojet gitlab repo over to a public one, here's the link to it: https://gitlab.cern.ch/damacdon/workflow_monojet

And here's the correct link to the line in the cli.py: https://github.com/reanahub/reana-workflow-engine-yadage/blob/master/reana_workflow_engine_yadage/cli.py#L104

Issue #386, opened 2020-02-24
validate: yadage parameters


tiborsimko:
Take `reana-demo-root6-roofit` as an example.

The workflow needs a parameter called `events`. 

If a user makes a mistake and calls this parameter `xevents`, the workflow still validates:

```console
reana-client validate -f ./reana-yadage.yaml
File reana-demo-root6-roofit/reana-yadage.yaml is a valid REANA specification file.
```

but the execution understandably fails:

```console
$ reana-client logs -w roofit-yadage
...
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/adage/__init__.py", line 51, in run_polling_workflow
    for stepnum, controller in enumerate(coroutine):
  File "/usr/local/lib/python3.6/site-packages/adage/pollingexec.py", line 92, in adage_coroutine
    update_dag(controller, extend_decider,recursive_updates)
  File "/usr/local/lib/python3.6/site-packages/adage/pollingexec.py", line 48, in update_dag
    update_dag(controller, decider, recurse)
  File "/usr/local/lib/python3.6/site-packages/adage/pollingexec.py", line 43, in update_dag
    update_loop.send(command)
  File "/usr/local/lib/python3.6/site-packages/adage/pollingexec.py", line 21, in update_coroutine
...
  File "/usr/local/lib/python3.6/site-packages/yadage/handlers/scheduler_handlers.py", line 200, in <dictcomp>
    for k, v in get_parameters(spec["parameters"]).items()
  File "/usr/local/lib/python3.6/site-packages/yadage/handlers/scheduler_handlers.py", line 49, in select_parameter
    value = handler(wflowview, parameter)
  File "/usr/local/lib/python3.6/site-packages/yadage/handlers/expression_handlers.py", line 161, in stage_output_selector
    step, selection.get("output"), selection_type == "scopes"
  File "/usr/local/lib/python3.6/site-packages/yadage/handlers/expression_handlers.py", line 55, in select_reference
    % (selection, step["result"])
RuntimeError: no matches found in reference selection. selection events | result <TypedLeafs: {'fitdata': '/var/reana/users/00000000-0000-0000-0000-000000000000/workflows/4616756d-2a10-4787-9237-3e243f83c782/code/fitdata.C', 'gendata': '/var/reana/users/00000000-0000-0000-0000-000000000000/workflows/4616756d-2a10-4787-9237-3e243f83c782/code/gendata.C', 'xevents': 20000}>
...
[...]

PR #462, opened 2021-02-24
validate: warn about using non-tagged images for yadage workflows

ParthS007:
addresses reanahub/reana#248

### DIFF
--- a/reana_client/utils.py
+++ b/reana_client/utils.py
@@ -7,7 +7,6 @@
 # under the terms of the MIT License; see LICENSE file for more details.
 """REANA client utils."""
 import base64
-import datetime
 import json
 import logging
 import os
@@ -17,11 +16,9 @@ import traceback
 from uuid import UUID
 
 import click
-import requests
 import yadageschemas
 import yaml
 from jsonschema import ValidationError, validate
-from reana_commons.config import WORKFLOW_RUNTIME_USER_GID, WORKFLOW_RUNTIME_USER_UID
 from reana_commons.errors import REANAValidationError
 from reana_commons.operational_options import validate_operational_options
 from reana_commons.serial import serial_load
@@ -33,7 +30,7 @@ from reana_client.config import (
     reana_yaml_schema_file_path,
     reana_yaml_valid_file_names,
 )
-from reana_client.validation import validate_parameters
+from reana_client.validation import validate_environment, validate_parameters
 
 
 def workflow_uuid_or_name(ctx, param, value):
@@ -156,6 +153,7 @@ def load_reana_spec(filepath, skip_validation=False, skip_validate_environments=
                 )
             )
             _validate_reana_yaml(reana_yaml)
+            validate_parameters(workflow_type, reana_yaml)
 
         reana_yaml["workflow"]["specification"] = load_workflow_spec(
             workflow_type,
@@ -174,14 +172,13 @@ def load_reana_spec(filepath, skip_validation=False, skip_validate_environments=
                 "Validating environments in REANA specification file: "
                 "{filepath}".format(filepath=filepath)
             )
-            _validate_environment(reana_yaml)
+            validate_environment(reana_yaml)
 
         if workflow_type == "yadage":
-            """
-            We don't want to expose the yadage workflow spec in the UI to
-            avoid the inconsistency between reana-server and reana-ui.
-            More Info: https://github.com/reanahub/reana-client/pull/462#discussion_r585794297
-            """
+            # We don't send the loaded Yadage workflow spec to the cluster as
+            # it may result in inconsistencies between what's displayed in the
+            # UI and the actual spec loaded at the workflow engine level.
+            # More info: https://github.com/reanahub/reana-client/pull/462#discussion_r585794297
             reana_yaml["workflow"]["specification"] = None
 
         return reana_yaml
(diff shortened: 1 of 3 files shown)


reana_client/utils.py
```diff
@@ -191,15 +191,27 @@ def _validate_environment(reana_yaml):
         workflow_steps = reana_yaml["workflow"]["specification"]["steps"]
         _validate_serial_workflow_environment(workflow_steps)
     elif workflow_type == "yadage":
-        logging.warning(
-            "The environment validation is not implemented for Yadage workflows yet."
-        )
+        with open(reana_yaml["workflow"].get("file")) as f:
```
ParthS007, line 194: @mvidalgarcia What do you think about loading the yadage steps in `load_reana_spec` somewhere or here it is fine?

reana_client/utils.py
```diff
@@ -191,15 +191,27 @@ def _validate_environment(reana_yaml):
...
 
+def _validate_yadage_workflow_environment(workflow_steps):
+    """Validate environments in REANA yadage workflow.
+
+    :param workflow_steps: List of dictionaries which represents different steps involved in workflow.
+    :raises Warning: Warns user if the workflow environment is invalid in yadage workflow steps.
+    """
+    for stage in workflow_steps:
+        environment = stage["scheduler"]["step"]["environment"]
+        image = "{}:{}".format(environment["image"], environment["imagetag"])
```
ParthS007, line 211: @mvidalgarcia I combined the image name and image tag here to send it to `_validate_image_tag`.

Should we go ahead with creating a separate function for that?

reana_client/utils.py
mvidalgarcia: To prevent this maybe we can move [the environment validation](https://github.com/reanahub/reana-client/blob/d36de00b2b2f748de0bcc19cbb163c9ed2ec00f2/reana_client/utils.py#L151-L156) after [loading the workflow spec](https://github.com/reanahub/reana-client/blob/d36de00b2b2f748de0bcc19cbb163c9ed2ec00f2/reana_client/utils.py#L164), so we have access to the Yadage steps already. Would that make sense?

reana_client/utils.py
mvidalgarcia: Something like what you have should already be fine. We can look later to avoid some repetition if possible.

reana_client/utils.py
ParthS007: @mvidalgarcia If we do that, we will have more options to explore the params in future by just passing `reana_yaml` in different methods.

mvidalgarcia:
Needs rebase

mvidalgarcia (commented):
Testing with `reana-demo-atlas-recast` I get:

```console
$ reana-client validate --environments                     
Something went wrong when trying to validate /Users/marco/code/reanahub/reana-demo-atlas-recast/reana.yaml
```

ParthS007:
@mvidalgarcia  

It is giving error because `environment` is not present in `stages` and it ends up in `KeyError`. I can add the check for `environment` as well.

Is it possible to not have `environment` at all? If yes, then this check should also be there in `serial` as well. What do you think?

mvidalgarcia:
> Is it possible to not have `environment` at all? If yes, then this check should also be there in `serial` as well. What do you think?

AFAIK it's not possible, all the jobs need an environment to run.

ParthS007:
Ahh, I see. Then we have to implement a new function to recursively iterate through the file to discover paths and load the specification. Looking into it. 👍

mvidalgarcia:
> Ahh, I see. Then we have to implement a new function to recursively iterate through the file to discover paths and load the specification. Looking into it. 👍

This has to be implemented somewhere in the code as we need this info to run the workflow. You can perhaps debug the steps when running the workflow to see when this information is attached to the reana spec and how.

reana_client/utils.py
```diff
@@ -198,15 +205,27 @@ def _validate_environment(reana_yaml):
...
+        image_name, image_tag = _validate_image_tag(image)
+        _image_exists(image_name, image_tag)
```
mvidalgarcia, line 226: We also need to validate UID and GIDs

```python
uid, gids = _get_image_uid_gids(image_name, image_tag)
_validate_uid_gids(uid, gids)
```

ParthS007: @mvidalgarcia  I have to add `kubernetes_uid` or skip it?

` _validate_uid_gids(uid, gids, kubernetes_uid=stage.get("kubernetes_uid"))`

mvidalgarcia: You have to add it yes, as it's possible for Yadage workflows to be explicit on it.

ParthS007: I have added it, PR should be good to review

reana_client/utils.py
```diff
@@ -198,15 +199,29 @@ def _validate_environment(reana_yaml):
...
+        uid, gids = _get_image_uid_gids(image_name, image_tag)
+        _validate_uid_gids(uid, gids, kubernetes_uid=stage.get("kubernetes_uid"))
```
mvidalgarcia, line 222: `kubernetes_uid` is not placed there, check for instance`reana-demo-atlas-recast` how it's not taking it into account. It should be something like this:

```python
k8s_uid = (
    environment.get("resources")[0].get("kubernetes_uid")
    if environment.get("resources")
    else None
)
```

*(untested)*

reana_client/utils.py
```diff
@@ -78,7 +78,7 @@ def yadage_load(workflow_file, toplevel=".", **kwargs):
     }
 
     try:
-        yadageschemas.load(
+        return yadageschemas.load(
```
mvidalgarcia, line 81: I'm not sure if this can have some repercussions, as this operation is done in the workflow-engine level [here](https://github.com/reanahub/reana-workflow-engine-yadage/blob/0cb641bd8b7e233dae6504ebbf8cf8e7091b7176/reana_workflow_engine_yadage/cli.py#L125-L130).

I think here it isn't returning anything because is run uniquely for validation purposes.

Maybe @tiborsimko @diegodelemos you know better?

mvidalgarcia: Checking [this PR](https://github.com/reanahub/reana-workflow-engine-yadage/pull/150/files), I think I understand the reason behind moving the `yadadeschemas.load()` call to the workflow-engine.

Yadage has some operation options called `toplevel` and `initdir` that are useful to point out the path to where to look for the workflow file to resolve it. This can admit for instance GitHub URLs (which is pretty cool!). In the past, these options were hardcoded on the workflow-engine level to point always to the current directory. See previous [`toplevel`](https://github.com/reanahub/reana-workflow-engine-yadage/pull/150/files#diff-0e0b04c271e320cd6f4f031cb2acf70a67f1ad43bd8ca6e4b9a8d247e83c0dd1L71) and [`initdir`](https://github.com/reanahub/reana-workflow-engine-yadage/pull/150/files#diff-0e0b04c271e320cd6f4f031cb2acf70a67f1ad43bd8ca6e4b9a8d247e83c0dd1L104) treatment.

As a consequence of [this issue](https://github.com/reanahub/reana-client/issues/396) we were asked to make these options configurable from `reana-client` side, so one could e.g. pass a GitHub URL as a `toplevel` parameter and get the workflow file remotely before executing the workflow.

But you might be wondering... why can't we just do this on the client-side? That's a very good question that I even asked myself half an hour ago.

The reason is that these options are relative paths which [we need to prepend the workflow workspace to](https://github.com/reanahub/reana-workflow-engine-yadage/pull/150/files#diff-0e0b04c271e320cd6f4f031cb2acf70a67f1ad43bd8ca6e4b9a8d247e83c0dd1R47), and the workflow workspace is only available on the cluster once the workflow has been created. All this has to be done before [calling `yadageschemas.load()`](https://github.com/reanahub/reana-workflow-engine-yadage/pull/150/files#diff-0e0b04c271e320cd6f4f031cb2acf70a67f1ad43bd8ca6e4b9a8d247e83c0dd1R109) to have the correct paths.

Said that, I think that after these changes we never pass the Yadage workflow specification as JSON and we always point to the file, to do all the logic on the engine level. Actually, as you can see [here](https://github.com/reanahub/reana-workflow-controller/blob/d1b14ee1aa33c6c7b376b1ce99e49909d3a4db26/reana_workflow_controller/workflow_run_manager.py#L102),  we removed the `--workflow-json` option altogether, in comparison with the other engines.

So I think this means that it might be safe to resolve the workflow spec on the client-side to perform validations, as we will resolve it again in the engine anyway, building the proper operational options.

mvidalgarcia: Another update on this, it seems that runtime parameters are respected with this PR checked out. Plus, I had completely forgotten that we show these separately in the Spec tab.

image

_Note the "...gendata.C(**1000**,"/var/reana/..." in the logs._

image

mvidalgarcia: The main drawback I see is what @diegodelemos mentioned. If you run a workflow with this PR, on will get the Yadage spec visible in the UI. e.g. one will be able to see the Yadage environment image and tag. But if the user re-uploads the yadage/workflow.yaml spec with some changes, those won't be visible in the Spec tab.

Whereas with the previous approach, the yadage workflow spec wasn't visible, instead its path was shown, which could be accessed via the Workspace tab with the latest changes.

reana_client/utils.py
ParthS007: Tested on `reana-demo-atlas-recast` and `reana-demo-root6-roofit`. Works fine.

reana_client/utils.py
ParthS007: Thanks @mvidalgarcia for the detailed explanation. 💯 

I will try to conclude this here and please correct me here if I am deviating somewhere.

- If we `return` the `spec` from `yadage_load`, the validation will work fine without any issues.
- The problem will be that It will not be in sync with `r-ui` and `r-server` while running workflows.
- Instead of paths, we will have the `spec` because of `return` and if user do the changes later, it will not be visible or applied in running workflows.

If I am understanding  it correct so far, then we can make a separate method for returning/loading the schema and use it for validation purposes and inform user about (what/which workflow spec) is being the validated.

What do you think?

audrium: If we decide that we don't want to expose the yadage workflow spec in the UI, one possible solution would be to still load and return yadage spec from [load_workflow_spec()](https://github.com/reanahub/reana-client/blob/03460b791804e287063a7b4e56cb6bf7c7d78e19/reana_client/utils.py#L94) (same interface for all engines will be kept) and once the spec is inside [load_reana_spec()](https://github.com/reanahub/reana-client/blob/03460b791804e287063a7b4e56cb6bf7c7d78e19/reana_client/utils.py#L94) and all the parameter validation is done it can be set to `None` with a comment and a link to this discussion..
Because we still have to return the spec for additional validation. It is also needed for this [PR](https://github.com/reanahub/reana-client/pull/470/files)

mvidalgarcia: Use case described, based on `reana-demo-root6-roofit`:

- Checkout this PR
- Run Yadage workflow: `reana-client run -w roofit-yad -f reana-yadage.yaml`
- Check on the UI the spec tab of the workflow, look for `image` and `imagetag`:
  ```console
    "image": "reanahub/reana-env-root6",
    "imagetag": "6.18.04",
  ```
- Modify `workflow/yadage/workflow.yaml` and change the image to an invalid one, e.g. `reanahub/reana-env-root666`.
- Upload the new Yadage spec: `reana-client upload -w roofit-yad workflow/yadage/workflow.yaml`
- Restart the workflow: `reana-client restart -w roofit-yad`
- Workflow fails due to invalid image but in the Spec the old image remains, leaving us in an inconsistent state (meaning that it's also inconsistent in the DB).

Conclusion: the `return` statement has an impact on the consistency of the Spec stored on DB and displayed on UI.


Possible solution: Would it be possible to **only** return the spec for validation purposes and not always?

ParthS007: I think it is possible to pass an optional flag (`validation=True`) to `load_reana_spec()` and return on the basis of that flag.

What do you think @mvidalgarcia @audrium ?

mvidalgarcia: `load_reana_spec()` already has a `skip_validation` flag, so it wouldn't be clear its meaning. Maybe `yadage_load` can get an optional `kwarg` called `return` and return on basis of that.

audrium: The problem is that parameters validation is not optional (I'm referring to [this](https://github.com/reanahub/reana-client/issues/386) issue), so while the flag would kind of work in this issue context, it won't for upcoming issues

ParthS007: Even if it will be not optional, having a flag to return or not the spec will be helpful in case of sync with `r-ui` and `r-server`.

reana_client/utils.py
```diff
@@ -134,6 +136,9 @@ def _prepare_kwargs(reana_yaml):
             kwargs["parameters"] = reana_yaml.get("inputs", {}).get("parameters", {})
             kwargs["original"] = True
 
+        if workflow_type == "yadage":
+            kwargs["return_workflow_spec"] = True
+
```
mvidalgarcia, line 141: `load_reana_spec` is called upon:

- [Workflow creation](https://github.com/reanahub/reana-client/blob/732664851a6a504896e53f6346eaad8074e60c83/reana_client/cli/workflow.py#L293)
- [Workflow restart](https://github.com/reanahub/reana-client/blob/732664851a6a504896e53f6346eaad8074e60c83/reana_client/cli/workflow.py#L513) (when `--file` is passed)

Apart from validation.

So we would still get the Yadage spec returned in those cases and [sent to the server](https://github.com/reanahub/reana-client/blob/732664851a6a504896e53f6346eaad8074e60c83/reana_client/api/client.py#L133).

mvidalgarcia (commented):
Needs rebase.

reana_client/utils.py
ParthS007: @mvidalgarcia  What about this then?

`if not skip_validate_environments and workflow_type == "yadage"`?

mvidalgarcia: @audrium is working on Yadage parameters validation (no `skip_validate_env` involved), and he needs the spec loaded, so I guess it wouldn't work for his case. I see two options, both ugly, unfortunately..

- `return_yadage_spec` flag in `load_reana_spec`, `False` by default.
- What @audrium mentioned [here](https://github.com/reanahub/reana-client/pull/462#discussion_r586296218)
- Any other idea?

ParthS007: Another option which is ugly as well could be that since we have created a separate `validation.py`, we can duplicate the code there and use that function to load the yadage spec just for validation.

I can't think of anything else which is not ugly.

We have to choose the least worst option.

audrium: Just to add an example, [this](https://github.com/reanahub/reana-client/pull/470/files#diff-4f20d18bf32c181d09b016a6a363171402d113bb01347de15da4768afcbe6998R181) is what I had in mind. Still ugly solution, but nothing better comes into my mind for now

reana_client/utils.py
```diff
@@ -200,15 +205,32 @@ def _validate_environment(reana_yaml):
...
+    """
+    for stage in workflow_steps:
+        environment = stage["scheduler"]["step"]["environment"]
+        image = "{}:{}".format(environment["image"], environment["imagetag"])
+        image_name, image_tag = _validate_image_tag(image)
+        _image_exists(image_name, image_tag)
+        uid, gids = _get_image_uid_gids(image_name, image_tag)
+        k8s_uid = None
+        if environment.get("resources"):
+            k8s_uid = environment.get("resources")[0].get("kubernetes_uid")
```
diegodelemos, line 230: This won't work if the `kubernetes_uid` is not in the first position, [which is perfectly possible](https://github.com/yadage/yadage-schemas/blob/260fbc40ec5b8afa7d59c0dd00591e02e13774e4/yadageschemas/packtivity/environment/docker-enc-schema.json#L22):
```python
In [2]: environment
Out[2]: 
{'environment_type': 'docker-encapsulated',
 'image': 'reanahub/reana-env-root6',
 'imagetag': '6.18.04',
 'resources': [{'compute_backend': 'htcondorcern'},
  {'htcondor_max_runtime': 'espresso'}]}

In [3]: environment['resources']
Out[3]: [{'compute_backend': 'htcondorcern'}, {'htcondor_max_runtime': 'espresso'}]

In [4]: environment['resources'].append({'kubernetes_uid': '500'})

In [5]: environment['resources']
Out[5]: 
[{'compute_backend': 'htcondorcern'},
 {'htcondor_max_runtime': 'espresso'},
 {'kubernetes_uid': '500'}]

In [6]: environment['resources'][0].get('kubernetes_uid')

In [7]: environment['resources'][2].get('kubernetes_uid')
Out[7]: '500'
```

mvidalgarcia: Good catch! Any idea why they're using a list when it could be a simple dictionary?

I guess something like this should do...
```python
next((resource['kubernetes_uid'] for resource in resources if 'kubernetes_uid' in resource), None)
```

reana_client/utils.py
mvidalgarcia: > Just to add an example, [this](https://github.com/reanahub/reana-client/pull/470/files#diff-4f20d18bf32c181d09b016a6a363171402d113bb01347de15da4768afcbe6998R181) is what I had in mind. Still ugly solution, but nothing better comes into my mind for now

IMHO, this plus an explanatory comment would be the cleanest solution.

reana_client/utils.py
ParthS007: Tested on `r-d-atlas-recast`.

reana_client/utils.py
```diff
@@ -200,15 +203,40 @@ def _validate_environment(reana_yaml):
...
+        image_name, image_tag = _validate_image_tag(image)
+        _image_exists(image_name, image_tag)
+        uid, gids = _get_image_uid_gids(image_name, image_tag)
+        k8s_uid = None
+        if environment.get("resources"):
+            resources = environment.get("resources")
+            k8s_uid = next(
+                (
+                    resource["kubernetes_uid"]
+                    for resource in resources
```
mvidalgarcia, line 232: If you have `environment.get("resources", [])` I think you can save the `if` statement and the `resources = environment.get("resources")` assignment.

reana_client/utils.py
```diff
@@ -200,15 +203,40 @@ def _validate_environment(reana_yaml):
...
 
 
+def _validate_yadage_workflow_environment(workflow_steps):
+    """Validate environments in REANA yadage workflow.
+
+    :param workflow_steps: List of dictionaries which represents different steps involved in workflow.
+    :raises Warning: Warns user if the workflow environment is invalid in yadage workflow steps.
+    """
+    for stage in workflow_steps:
+        environment = stage["scheduler"]["step"]["environment"]
```
audrium, line 221: I haven't tested, but this probably wouldn't work with `reana-demo-bsm-search` example which have nested structure I showed this morning. Not all `scheduler` objects have `step`, and not all `step` objects have `environment` value. So perhaps you could use `.get("step", {}).get("environment", {})` technique here for safety.

Also, since there could be nested workflow structures, steps inside steps like in this [example](https://github.com/reanahub/reana-demo-bsm-search/blob/master/workflow/databkgmc.yml#L20), we should probably parse it recursively. I'm currently working on this for parameters parsing, so perhaps this PR could be merged as is, and separate issue can be created to later address it once the implementation for parameters would be there?

reana_client/utils.py
mvidalgarcia: Have you tested it by adding more resources as in @diegodelemos's example?

reana_client/utils.py
mvidalgarcia: Isn't it a way to load all the nested structured and have a _final_ dict with everything loaded? Instead of parsing it ourselves recursively. In the end, Yadage has to load these nested structures sooner or later so there might some utils for this?

audrium: That would be great to have, but I'm not familiar with such a way. Currently `yadageschemas` returns only the nested structure for us

reana_client/utils.py
```diff
@@ -177,6 +169,17 @@ def _prepare_kwargs(reana_yaml):
...
 
+        if not skip_validate_environments:
+            logging.info(
+                "Validating environments in REANA specification file: "
+                "{filepath}".format(filepath=filepath)
+            )
+            _validate_environment(reana_yaml)
+
+        if workflow_type == "yadage":
+            # Info: https://github.com/reanahub/reana-client/pull/462#discussion_r585794297
```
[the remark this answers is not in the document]
ParthS007: I have added couple of lines and link to GH discussion.

mvidalgarcia (approved):

PR #468, opened 2021-03-01
validate: add cwl environment validation 

mvidalgarcia:
Depends on https://github.com/reanahub/reana-client/pull/462
Closes reanahub/reana#248

### DIFF
--- a/reana_client/validation.py
+++ b/reana_client/validation.py
@@ -42,9 +42,8 @@ def validate_environment(reana_yaml):
         workflow_steps = reana_yaml["workflow"]["specification"]["stages"]
         _validate_yadage_workflow_environment(workflow_steps)
     elif workflow_type == "cwl":
-        logging.warning(
-            "The environment validation is not implemented for CWL workflows yet."
-        )
+        workflow_file = reana_yaml["workflow"].get("file")
+        _validate_cwl_workflow_environment(workflow_file)
 
 
 def _validate_yadage_workflow_environment(workflow_steps):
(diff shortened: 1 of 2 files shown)


reana_client/utils.py
```diff
@@ -208,9 +208,34 @@ def _validate_environment(reana_yaml):
...
+
+    :param workflow_file: Path to CWL workflow specification.
+    :raises Warning: Warns user if the workflow environment is invalid in CWL workflow steps.
+    """
+    try:
+        import cwl_utils.parser_v1_0 as cwl_parser
+        from cwl_utils.docker_extract import traverse
+    except ImportError as e:
+        click.secho(
+            "==> ERROR: Environment image validation using Python 2.7 is not supported for CWL workflows.",
```
diegodelemos, line 226: Minor: I think it would not be clear what action to take if you see this message, I would rather specifically say what to do. Something along these lines:
`==> ERROR: Cannot validate environment. Please install reana-client on Python 3+ to enable environment validation for CWL workflows.`

reana_client/utils.py
```diff
@@ -208,9 +208,34 @@ def _validate_environment(reana_yaml):
...
+            fg="red",
         )
+        raise e
+
+    top = cwl_parser.load_document(workflow_file)
+
+    for image in traverse(top):
+        image_name, image_tag = _validate_image_tag(image)
+        _image_exists(image_name, image_tag)
+        uid, gids = _get_image_uid_gids(image_name, image_tag)
```
diegodelemos, line 237: Observed while trying different scenarios:
```
~/reana-demo-cms-h4l $ reana-client validate --environments -f reana.yaml
==> WARNING: Using 'latest' tag is not recommended in cmsopendata/cmssw_5_3_32 environment image.
==> Environment image cmsopendata/cmssw_5_3_32:latest exists.                                                                              
Something went wrong when trying to validate /Users/rodrigdi/project/reana/src/reana-demo-cms-h4l/reana.yaml
```

This is due to additional output written to stdout by the image's initialisation script (`~/.bashrc`):
```console
$ docker run -i -t --rm cmsopendata/cmssw_5_3_32:latest bash -c "/usr/bin/id -u && /usr/bin/id -G"
Setting up CMSSW_5_3_32
CMSSW should now be available.
1000
1000 0
```
We can quickly fix it by taking into account only the last two lines of the output in https://github.com/reanahub/reana-client/blob/732664851a6a504896e53f6346eaad8074e60c83/reana_client/utils.py#L335.

mvidalgarcia: Wow good catch, I'd never thought of this possibility. Pushed a new commit to fix this.

diegodelemos (approved):
Solid 💪🏽

Issue #458, opened 2021-02-19
validate: verify possibly dangerous operations


tiborsimko:
Building up on previous tasks of verifying environment images, see #457 for example, we can take validation further to perform some sanity checks on the workflow commands.

For example, it happened that a workflow was running `sudo /some/command/...` all the while asking to run the job as some kubernetes_uid. We should warn the user about these cases.

For example, the `validate --steps` could run a quick `grep` looking for `sudo`, and report findings.

Another example, operations like `cd /some/path` may be dangerous, if there is a future cp/wget/xrdcp` populating some internal directory hard-coded inside the container image, and not populating the usual workspace directory, with input files. We could do a quick grep helping to spot these situations.

IOW, we are not looking for the perfect solution, but a good first approach to catching these early on the reana-client side.

(E.g. the writing-to-local-ephemeral-storage situation will be best caught by the REANA platform on the server side, but that will become another issue.)

===

**Current status:**

- [x] Serial - https://github.com/reanahub/reana-client/pull/471
- [x] Yadage - https://github.com/reanahub/reana-client/pull/470/commits/9c886a1c11cc3524a2c9525e8a8850f2736e5aaf
- [x] CWL - https://github.com/reanahub/reana-client/pull/481

mvidalgarcia:
Closed by https://github.com/reanahub/reana-client/pull/481

PR #469, opened 2021-03-02
validate: add serial parameters validation

mvidalgarcia:
closes #465

PR #471, opened 2021-03-03
validate: add serial workflow parameters test and check for dangerous operations

mvidalgarcia:
addresses #458

### DIFF
--- a/tests/test_validate_parameters.py
+++ b/tests/test_validate_parameters.py
@@ -1,0 +1,69 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of REANA.
+# Copyright (C) 2021 CERN.
+#
+# REANA is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""REANA client validate parameters tests."""
+
+import yaml
+
+from reana_client.validation import validate_serial_parameters
+
+
+def test_validate_parameters_serial(create_yaml_workflow_schema, capsys):
+    """Validate parameters for Serial workflows."""
+    reana_yaml = yaml.load(create_yaml_workflow_schema, Loader=yaml.FullLoader)
+    validate_serial_parameters(reana_yaml)
+    captured = capsys.readouterr()
+    assert not captured.out
+
+    # Input parameter not being used
+    reana_yaml["inputs"]["parameters"]["foo"] = "foo"
+    validate_serial_parameters(reana_yaml)
+    captured = capsys.readouterr()
+    assert 'WARNING: Input parameter "foo" is not being used' in captured.out
+    del reana_yaml["inputs"]["parameters"]["foo"]
+
+    # Escaped env vars don't trigger validation warning
+    reana_yaml["workflow"]["specification"]["steps"].append(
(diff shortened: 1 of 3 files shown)


tests/test_validate_parameters.py
```diff
@@ -0,0 +1,53 @@
...
+    assert 'WARNING: Input parameter "foo" is not being used' in captured.out
+    del reana_yaml["inputs"]["parameters"]["foo"]
+
+    reana_yaml["workflow"]["specification"]["steps"].append(
+        {"commands": ["python ${foo} --bar"]}
+    )
+    validate_serial_parameters(reana_yaml)
+    captured = capsys.readouterr()
+    assert (
+        'WARNING: Parameter "foo" found on step "Unnamed" is not defined in inputs parameters'
```
diegodelemos, line 31: Perhaps, instead of "Unnamed" we can simply display the step number that has the error? It is not fully harmonised with the UI (there if the step doesn't have a name, we display the job UUID, see below) but still it could make more sense?
image

tests/test_validate_parameters.py
```diff
@@ -0,0 +1,53 @@
...
+
+    reana_yaml["workflow"]["specification"]["steps"].append(
+        {"commands": ["python ${bar} --foo"], "name": "qux"}
+    )
+    validate_serial_parameters(reana_yaml)
+    captured = capsys.readouterr()
+    assert (
+        'WARNING: Parameter "bar" found on steps "baz, qux" is not defined in inputs parameters'
+        in captured.out
+    )
```
diegodelemos, line 53: Perhaps we could add one last test to make sure that escaped environment variables are not cougth as false positives? (similar to second case listed [here](https://github.com/reanahub/reana-client/pull/469#pullrequestreview-602772503)).

tests/test_validate_parameters.py
mvidalgarcia: How can we get this on the client-side?

mvidalgarcia: Nevermind, I think I hadn't understood your message... Updated.

diegodelemos (approved):
Works nicely.
```
$ git diff
index b1e8a2b..784770c 100644
--- a/reana.yaml
+++ b/reana.yaml
@@ -18,7 +18,7 @@ workflow:
       - name: fitdata
         environment: 'reanahub/reana-env-root6:6.18.04'
         commands:
-        - root -b -q 'code/fitdata.C("${data}","${plot}")'
+        - cd / && sudo root -b -q 'code/fitdata.C("${data}","${plot}")'
 outputs:
   files:
     - results/plot.png
$ reana-client validate -f reana.yaml
==> WARNING: Operation "sudo" found in step "fitdata" might be dangerous.
==> WARNING: Operation "cd /" found in step "fitdata" might be dangerous.
File /Users/rodrigdi/project/reana/src/reana-demo-root6-roofit/reana.yaml is a valid REANA specification file.
```

mvidalgarcia:
> Edited: shouldn't we change `closes` to `addresses`, since we did not implement it for other workflow engines? cc @mvidalgarcia

True 🤦, _l'habitude_... thanks for the thorough review!

PR #477, opened 2021-03-10
validate: Add CWL parameters validation

### LINKED ISSUE #473: validate: CWL parameters validation
Already implemented for [Serial and Yadage workflows](https://github.com/reanahub/reana-client/blob/cc3df10d59ffa1d82a376d351d075afcde0a41a7/reana_client/validation.py#L288-L291), parameter validation should also be implemented for CWL workflows.

Check both the current implementation and previous issues created for other engines to get an idea of what should be implemented:
- https://github.com/reanahub/reana-client/issues/465
- https://github.com/reanahub/reana-client/issues/386
- https://github.com/reanahub/reana-client/issues/466

While doing this one can easily [plug the dangerous operations check](https://github.com/reanahub/reana-client/issues/458).

ParthS007:
closes #473

Tested  with:
- `reana-demo-root6-roofit`
- `reana-demo-cdci-crab-pulsar-integral-verification`
- `reana-demo-cms-h4l`

### DIFF
--- a/reana_client/validation.py
+++ b/reana_client/validation.py
@@ -15,6 +15,7 @@ import subprocess
 import sys
 from collections import defaultdict
 from itertools import chain
+import os
 
 import click
 import requests
@@ -431,7 +432,7 @@ def validate_parameters(workflow_type, reana_yaml):
     """
     validate = {
         "yadage": _validate_yadage_parameters,
-        "cwl": lambda *args: None,
+        "cwl": _validate_cwl_parameters,
         "serial": _validate_serial_parameters,
     }
     """Dictionary to extend with new workflow specification loaders."""
@@ -487,6 +488,28 @@ def _warn_not_defined_parameters(
             ),
             fg="yellow",
         )
+
+
+def _validate_cwl_parameters(reana_yaml):
+    """Validate input parameters for CWL workflows.
+
+    :param reana_yaml: REANA YAML specification.
+    """
+    cwl_main_spec_path = reana_yaml["workflow"].get("file")
+    if os.path.exists(cwl_main_spec_path):
+        run_command(
+            "cwltool --validate --strict {}".format(cwl_main_spec_path),
+            display=False,
+            return_output=True,
+            stderr_output=True,
+        )
+    else:
+        click.secho(
+            "==> ERROR: Workflow path {} is not valid.".format(cwl_main_spec_path),
+            err=True,
+            fg="red",
+        )
+        sys.exit(1)
 
 
 def _validate_serial_parameters(reana_yaml):
@@ -620,7 +643,7 @@ def _validate_dangerous_operations(command, step):
             )
 
 
-def run_command(cmd, display=True, return_output=False):
+def run_command(cmd, display=True, return_output=False, stderr_output=False):
     """Run given command on shell in the current directory.
 
     Exit in case of troubles.
@@ -638,7 +661,8 @@ def run_command(cmd, display=True, return_output=False):
         click.secho("{0}".format(cmd), bold=True)
     try:
         if return_output:
-            result = subprocess.check_output(cmd, shell=True)
+            stderr_flag_val = subprocess.STDOUT if stderr_output else None
+            result = subprocess.check_output(cmd, stderr=stderr_flag_val, shell=True)
             return result.decode().rstrip("\r\n")
         else:
             subprocess.check_call(cmd, shell=True)
@@ -646,4 +670,6 @@ def run_command(cmd, display=True, return_output=False):
         if display:
             click.secho("[{0}] ".format(now), bold=True, nl=False, fg="green")
             click.secho("{0}".format(err), bold=True, fg="red")
+        if stderr_output:
+            sys.exit(err.output.decode())
         sys.exit(err.returncode)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -43,6 +43,103 @@ def create_yaml_workflow_schema():
 
 
 @pytest.fixture()
+def create_cwl_yaml_workflow_schema():
+    """Return dummy cwl workflow schema."""
+    reana_cwl_yaml_schema = """
+        version: 0.7.2
+        workflow:
+          type: cwl
+          file: main.cwl
+        outputs:
+          files:
+            - foo/bar
+        """
+    return reana_cwl_yaml_schema
+
+
+@pytest.fixture()
+def cwl_workflow_spec_step():
+    """Return dummy CWL workflow loaded spec step."""
+    cwl_workflow_spec_step = """
+        #!/usr/bin/env cwl-runner
+
+        cwlVersion: v1.0
+        class: Workflow
+
+        inputs:
+          outputfile:
+            type: string
+            inputBinding:
+              prefix: --outputfile
+
+        outputs:
+          result:
+            type: File
+            outputSource: first/result
+
+        steps:
+          first:
+            run: test.tool
+            in:
+              outputfile: outputfile
+            out: [result]
+        """
+    return cwl_workflow_spec_step
+
+
+@pytest.fixture()
+def cwl_workflow_spec_correct_input_param():
+    """Return correct dummy CWL workflow loaded spec."""
+    cwl_workflow_spec = """
+        #!/usr/bin/env cwl-runner
+
+        cwlVersion: v1.0
+        class: CommandLineTool
+
+        baseCommand: python
+
+        inputs:
+          outputfile:
+            type: string
+            inputBinding:
+              prefix: --outputfile
+
+        outputs:
+          result:
+            type: File
+            outputBinding:
+              glob: $(inputs.outputfile)
+        """
+    return cwl_workflow_spec
+
+
+@pytest.fixture()
+def cwl_workflow_spec_wrong_input_param():
+    """Return wrong dummy CWL workflow loaded spec."""
+    cwl_workflow_spec = """
+        #!/usr/bin/env cwl-runner
+
+        cwlVersion: v1.0
+        class: CommandLineTool
+
+        baseCommand: python
+
+        inputs:
+          xoutputfile:  # wrong input param
+            type: string
+            inputBinding:
+              prefix: --outputfile
+
+        outputs:
+          result:
+            type: File
+            outputBinding:
+              glob: $(inputs.outputfile)
+        """
+    return cwl_workflow_spec
+
+
+@pytest.fixture()
 def yadage_workflow_spec_loaded():
     """Return dummy Yadage nested workflow spec loaded as dictionary."""
     return {
--- a/tests/test_validate_parameters.py
+++ b/tests/test_validate_parameters.py
@@ -10,12 +10,63 @@
 
 import pytest
 import yaml
+from click.testing import CliRunner
 
+from reana_client.utils import cwl_load
 from reana_client.validation import (
     _validate_dangerous_operations,
     _validate_serial_parameters,
     _validate_yadage_parameters,
+    _validate_cwl_parameters,
 )
+
+
+def test_validate_parameters_cwl(
+    create_cwl_yaml_workflow_schema,
+    cwl_workflow_spec_step,
+    cwl_workflow_spec_correct_input_param,
+    cwl_workflow_spec_wrong_input_param,
+    capsys,
+):
+    """Validate parameters for CWL workflows."""
+    runner = CliRunner()
+
+    def get_loaded_yaml(step_spec, input_spec):
+        with open("main.cwl", "w") as f:
+            f.write(step_spec)
+        with open("test.tool", "w") as g:
+            g.write(input_spec)
+        reana_yaml = yaml.load(create_cwl_yaml_workflow_schema, Loader=yaml.FullLoader)
+        reana_yaml["workflow"]["specification"] = cwl_load(
+            reana_yaml["workflow"].get("file")
+        )
+        return reana_yaml
+
+    with runner.isolated_filesystem():
+        reana_yaml = get_loaded_yaml(
+            cwl_workflow_spec_step, cwl_workflow_spec_correct_input_param
+        )
+        _validate_cwl_parameters(reana_yaml)
+        captured = capsys.readouterr()
+        assert not captured.out
+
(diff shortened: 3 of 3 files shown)


reana_client/validation.py
```diff
@@ -343,6 +344,28 @@ def _warn_not_defined_parameters(
...
+def _validate_cwl_parameters(reana_yaml):
+    """Validate input parameters for CWL workflows.
+
+    :param reana_yaml: REANA YAML specification.
+    """
+
+    # Check if cwltool is installed.
+    run_command("cwltool --version", display=False, return_output=True)
+
+    cwl_directory = reana_yaml["inputs"]["directories"][0]
```
mvidalgarcia, line 356: Moderate: We can't just assume that there is always a `directories`, nor that the first item it's gonna be always `workflow/cwl`. `directories` is just an optional input property to upload all that's inside to the cluster. Instead, we should better look at [`workflow.file`](https://docs.reana.io/reference/reana-yaml/#reanayaml-workflow) and from there maybe look at possible references.

Check for instance `reana-demo-cdci-crab-pulsar-integral-verification` that doesn't have `directories`.

reana_client/validation.py
```diff
@@ -343,6 +344,33 @@ def _warn_not_defined_parameters(
...
+    cwl_files = set()
+    cwl_main_spec_path = reana_yaml["workflow"]["file"]
+    cwl_graph = reana_yaml["workflow"]["specification"].get("$graph", None)
+
+    if cwl_graph:
+        cwl_directory = cwl_main_spec_path.rsplit("/", 1)[0]
+        for spec in cwl_graph:
+            spec_file = spec.get("id")[1:]
+            spec_file_path = cwl_directory + "/" + spec_file
+            cwl_files.add(spec_file_path)
```
mvidalgarcia, line 362: As we discussed yesterday, we don't need to check the files that are referenced inside the main workflow spec file as `cwltool` would take care of it for us. 

For instance, choosing `r-d-helloworld` as reference. If you modify `helloworld.tool`:
```diff
diff --git a/workflow/cwl/helloworld.tool b/workflow/cwl/helloworld.tool
index 4aa963c..60a8b6b 100644
--- a/workflow/cwl/helloworld.tool
+++ b/workflow/cwl/helloworld.tool
@@ -19,7 +19,7 @@ inputs:
     type: File
     inputBinding:
       prefix: --inputfile
-  sleeptime:
+  sleeptim:
     type: int
     inputBinding:
       prefix: --sleeptime
```

And the you run `cwltool` pointing **to the main workflow file**, it would already warn us that something is wrong:

```console
$ cwltool --validate --strict workflow/cwl/helloworld.cwl
INFO /Users/marco/.virtualenvs/r/bin/cwltool 1.0.20191022103248
INFO Resolved 'workflow/cwl/helloworld.cwl' to 'file:///Users/marco/code/reanahub/reana-demo-helloworld/workflow/cwl/helloworld.cwl'
ERROR Tool definition failed validation:
workflow/cwl/helloworld.cwl:34:5: Step is missing required parameter 'sleeptim'
```

So we don't need to iterate ourselves, it would be as simple as calling `cwltool` once with `workflow.file`.

mvidalgarcia (commented):
Are we adding unit tests in the end? We can take https://github.com/reanahub/reana-client/blob/maint-0.7/tests/test_validate_parameters.py as a reference.

reana_client/validation.py
ParthS007: Yes, Adding the test rn. Thanks for reminding 👍

reana_client/validation.py
```diff
@@ -343,6 +344,20 @@ def _warn_not_defined_parameters(
         )
 
 
+def _validate_cwl_parameters(reana_yaml):
+    """Validate input parameters for CWL workflows.
+
+    :param reana_yaml: REANA YAML specification.
+    """
+    cwl_main_spec_path = reana_yaml["workflow"]["file"]
+    if os.path.exists(cwl_main_spec_path):
```
tiborsimko, line 353: What if the file does not exist, since the user has made a typo? The parameter validator should fail in this case, since the validation cannot be done... I guess the workflow validation would fail before, but if someone calls this function just for the parameter validation purposes, it would be safer to report the error as well.

tests/conftest.py
```diff
@@ -42,6 +42,124 @@ def create_yaml_workflow_schema():
     return reana_yaml_schema
 
 
+@pytest.fixture()
+def create_cwl_yaml_workflow_schema():
+    """Return dummy cwl workflow schema."""
+    reana_cwl_yaml_schema = """
+        version: 0.3.0
```
mvidalgarcia, line 49:
```suggestion
        version: 0.7.2
```

tests/conftest.py
```diff
@@ -42,6 +42,124 @@ def create_yaml_workflow_schema():
...
+        - id: chi2_limit
+          inputBinding:
+            prefix: --inp-chi2_limit=
+            separate: false
+          type: float
+        - id: systematic_fraction
+          inputBinding:
+            prefix: --inp-systematic_fraction=
+            separate: false
+          type: float
```
mvidalgarcia, line 102: I would simplify as much as possible and keep generic params: foo, bar, etc. Check the other examples for reference. I wouldn't copy anything from our examples.

tests/conftest.py
```diff
@@ -42,6 +42,124 @@ def create_yaml_workflow_schema():
...
+          type: string
+        - doc: the best tested unacceptable fit folded model
+          id: next_good_fit_png
+          type: string
+        - doc: the best tested unacceptable fit folded model png b64
+          id: next_good_fit_png_content
+          type: string
+        requirements:
+          DockerRequirement:
+            dockerPull: reanahub/reana-demo-cdci-crab-pulsar-integral-verification
```
mvidalgarcia, line 127: This can be a dummy docker image. No need to reference a real one.

tests/conftest.py
```diff
@@ -42,6 +42,124 @@ def create_yaml_workflow_schema():
...
 
 
+@pytest.fixture()
+def create_cwl_yaml_workflow_schema():
+    """Return dummy cwl workflow schema."""
+    reana_cwl_yaml_schema = """
+        version: 0.3.0
+        inputs:
+          parameters:
+            input: inputs.yaml
```
mvidalgarcia, line 52: Is this being tested? If not I'd remove it and keep it as simple as possible.

tests/test_validate_parameters.py
```diff
@@ -10,14 +10,47 @@
...
+
+    # Wrong Input parameter used
+    with runner.isolated_filesystem():
+        with open("test.cwl", "w") as f:
+            f.write(cwl_workflow_spec_two)
+        reana_yaml = yaml.load(create_cwl_yaml_workflow_schema, Loader=yaml.FullLoader)
+        reana_yaml["workflow"]["specification"] = cwl_load(
+            reana_yaml["workflow"].get("file")
+        )
+        pytest.raises(SystemExit, _validate_cwl_parameters, reana_yaml)
```
mvidalgarcia, line 51: Could you add a check of the error output and verify which parameters are wrong?

E.g. with this output

```console
$ cwltool --validate --strict workflow/cwl/helloworld.cwl 
INFO /Users/marco/.virtualenvs/r/bin/cwltool 1.0.20191022103248
INFO Resolved 'workflow/cwl/helloworld.cwl' to 'file:///Users/marco/code/reanahub/reana-demo-helloworld/workflow/cwl/helloworld.cwl'
ERROR Tool definition failed validation:
workflow/cwl/helloworld.cwl:34:5: Step is missing required parameter 'sleeptim'
```

verify  `sleeptim`  is in the error output.

tests/conftest.py
```diff
@@ -42,6 +42,124 @@ def create_yaml_workflow_schema():
...
+          type: string
+        requirements:
+          DockerRequirement:
+            dockerPull: reanahub/reana-demo-cdci-crab-pulsar-integral-verification
+        """
+    return cwl_workflow_spec
+
+
+@pytest.fixture()
+def cwl_workflow_spec_two():
```
mvidalgarcia, line 133: At first sight, I don't see what's wrong with this workflow spec. Instead of calling them one and two, could we have more meaningful names?

reana_client/validation.py
```diff
@@ -343,6 +344,27 @@ def _warn_not_defined_parameters(
...
+    cwl_main_spec_path = reana_yaml["workflow"].get("file")
+    if os.path.exists(cwl_main_spec_path):
+        run_command(
+            "cwltool --validate --strict --debug {}".format(cwl_main_spec_path),
+            display=False,
+            return_output=True,
+        )
+    else:
+        click.secho(
+            "==> ERROR: Workflow path {} is not valid.".format(cwl_main_spec_path),
```
mvidalgarcia, line 361: Could you please add a test case for this case?

tests/conftest.py
```diff
@@ -42,6 +42,84 @@ def create_yaml_workflow_schema():
...
+        - -c
+        baseCommand: bash
+        class: CommandLineTool
+        cwlVersion: v1.0
+        id: foo
+        inputs:
+        - id: bar
+          inputBinding:
+            prefix: --inp-bar=
+            sepearate: false
```
mvidalgarcia, line 110: Is this a typo? I still don't understand what's wrong with this spec. Where are the steps?

tests/test_validate_parameters.py
```diff
@@ -10,14 +10,53 @@
...
+        reana_yaml["workflow"]["specification"] = cwl_load(
+            reana_yaml["workflow"].get("file")
+        )
+        with pytest.raises(SystemExit) as error:
+            _validate_cwl_parameters(reana_yaml)
+        assert error.value.code == 1
+
+    # Wrong file path used
+    reana_yaml = yaml.load(create_cwl_yaml_workflow_schema, Loader=yaml.FullLoader)
+    pytest.raises(SystemExit, _validate_cwl_parameters, reana_yaml)
```
mvidalgarcia, line 57: Could you check the output and verify that is what we expect? We do the same for other tests.

tests/test_validate_parameters.py
mvidalgarcia: This is not addressed.

tests/conftest.py
ParthS007: Yes, This is the intentional error to check.

tests/test_validate_parameters.py
ParthS007: The error message is not captured in `capsys`. I am still looking into that. Only the error code is there.

tests/conftest.py
mvidalgarcia: Hard to catch... I think this isn't validating what we want... It's validating syntax instead of parameters. What we want to catch is that you have e.g. a parameter called `foo` and you reference it with `fooo`.

tests/test_validate_parameters.py
ParthS007: @mvidalgarcia  I am just getting the error code that is `1` here not the error message that I get when I run  `reana-client validate --environments`.

ParthS007: @mvidalgarcia The reason why I am not getting the error message is that when `run_command` method runs the `cwltool` command, it throws an error and in case of error we are just returning the `err.returncode` from there.

I can do the `subprocess` call direct from `validate_cwl` instead of calling `run_command`. Wdyt?

mvidalgarcia: Maybe you can enrich `run_command` with a kwarg to return also the output and use it?

reana_client/validation.py
```diff
@@ -454,7 +477,7 @@ def _validate_dangerous_operations(command, step):
             )
 
 
-def run_command(cmd, display=True, return_output=False):
+def run_command(cmd, display=True, return_output=False, **kwargs):
```
mvidalgarcia, line 480: No need to call it `kwargs`, I meant that it could be a keyword argument. It can be called `stderr_output` and can be `False` by default.

tests/conftest.py
```diff
@@ -42,6 +42,103 @@ def create_yaml_workflow_schema():
...
+    cwl_workflow_spec = """
+        #!/usr/bin/env cwl-runner
+
+        cwlVersion: v1.0
+        class: CommandLineTool
+
+        baseCommand: python
+
+        inputs:
+          outputfiles:
```
mvidalgarcia, line 128: Can we make this more obvious? It took me a while to spot the problem 😅 Maybe we can call it `xoutputfile` and add a comment next to it.

```yaml
inputs:
  xoutputfile:  # wrong input param
  ...
```

tests/test_validate_parameters.py
```diff
@@ -10,14 +10,60 @@
...
+    runner = CliRunner()
+    with runner.isolated_filesystem():
+        with open("main.cwl", "w") as f:
+            f.write(cwl_workflow_spec_step)
+        with open("test.tool", "w") as g:
+            g.write(cwl_workflow_spec_correct_input_param)
+        reana_yaml = yaml.load(create_cwl_yaml_workflow_schema, Loader=yaml.FullLoader)
+        reana_yaml["workflow"]["specification"] = cwl_load(
+            reana_yaml["workflow"].get("file")
+        )
```
mvidalgarcia, line 41: This is kind of repeated between the tests, maybe we can wrap it into a function inside this test that gets the workflow spec as param?

tests/test_validate_parameters.py
```diff
@@ -10,14 +10,60 @@
...
+        with pytest.raises(SystemExit) as exc_info:
+            _validate_cwl_parameters(reana_yaml)
+        assert (
+            "Step is missing required parameter 'outputfiles'" in exc_info.value.args[0]
+        )
+
+    # Wrong file path used
+    reana_yaml = yaml.load(create_cwl_yaml_workflow_schema, Loader=yaml.FullLoader)
+    pytest.raises(SystemExit, _validate_cwl_parameters, reana_yaml)
```
mvidalgarcia, line 64: Could we verify here that the output message is actually `ERROR: Workflow path main.cwl is not valid.`?

mvidalgarcia (commented):
It's looking much much better 👍. Some last details.

mvidalgarcia (approved):

Issue #431, opened 2020-10-15
cli: Improve the cosmetics of output messages


ParthS007:
Yes, as we do for the `logs` command already:

```console
$ reana-client logs -w helloworld-yadage --filter staaatus=foo
Error: filter 'staaatus' is not valid.
Available filters are 'compute_backend' 'docker_img' 'status' 'step'.
```

We can unify the formatting at least for the various `--filter` commands everywhere.

BTW note that the output message formatting cosmetics will remain elsewhere to be amend as well, for example in `ping` we don't report it either:

```console
$ reana-client ping
REANA client is not connected to any REANA cluster.
```

Something for another issue. We have quite a potential to improve cosmetics regarding `==> ERROR: ...` and friends.

_Originally posted by @tiborsimko in https://github.com/reanahub/reana-client/pull/428#discussion_r504760350_

mvidalgarcia:
Closed by https://github.com/reanahub/reana-client/pull/553, thanks to @Sinclert 👍

PR #479, opened 2021-03-17
validate: enrich output messages

mvidalgarcia:
Paves the way to implement https://github.com/reanahub/reana-client/issues/431

### How it looks
image

image

### To test
Run `reana-client validate` or `reana-client validate --environments --pull` with any of our examples for any workflow engine.
- To force warnings: e.g. add dangerous operations (`sudo` , `cd /`) to workflow commands, add unused parameters, use undefined parameters..
- To force errors: e.g. modify env images to non-existing ones.

### DIFF
--- a/reana_client/printer.py
+++ b/reana_client/printer.py
@@ -1,0 +1,51 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of REANA.
+# Copyright (C) 2021 CERN.
+#
+# REANA is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""reana-client output print configuration."""
+
+import click
+
+from reana_client.config import (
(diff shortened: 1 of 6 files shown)


reana_client/printer.py
```diff
@@ -0,0 +1,51 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of REANA.
+# Copyright (C) 2021 CERN.
+#
+# REANA is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""cernopendata-client output print configuration."""
```
ParthS007, line 9:
```suggestion
"""reana-client output print configuration."""
```

😉

ParthS007 (approved):
LGTM! Tested with validation outputs.

Just the output from `cwltool` is different which is okay IMHO.

PR #481, opened 2021-03-17
validate: verify possibly dangerous operations in CWL workflows

### LINKED ISSUE #458: validate: verify possibly dangerous operations

ParthS007:
closes #458

### DIFF
--- a/reana_client/validation.py
+++ b/reana_client/validation.py
@@ -521,6 +521,33 @@ def _validate_cwl_parameters(reana_yaml):
         )
         sys.exit(1)
 
+    def _check_dangerous_operations(workflow):
+        """Check for "baseCommand" and "arguments" in workflow.
+
+        If these keys are found, validate if they have dangerous operations.
+        """
+        cmd_keys = ["baseCommand", "arguments"]
+        for cmd_key in cmd_keys:
+            if cmd_key in workflow:
+                cmd_values = (
+                    workflow[cmd_key]
+                    if isinstance(workflow[cmd_key], list)
+                    else [workflow[cmd_key]]
+                )
+                for cmd_value in cmd_values:
+                    _validate_dangerous_operations(
+                        str(cmd_value), step=workflow.get("id")
+                    )
+
+    workflow = reana_yaml["workflow"]["specification"].get(
+        "$graph", reana_yaml["workflow"]["specification"]
+    )
+    if isinstance(workflow, dict):
+        _check_dangerous_operations(workflow)
+    elif isinstance(workflow, list):
+        for wf in workflow:
+            _check_dangerous_operations(wf)
+
 
 def _validate_serial_parameters(reana_yaml):
     """Validate input parameters for Serial workflows.
@@ -540,7 +567,7 @@ def _validate_serial_parameters(reana_yaml):
     ):
         step_name = step.get("name", str(idx))
         for command in step["commands"]:
-            _validate_dangerous_operations(command, step_name)
+            _validate_dangerous_operations(command, step=step_name)
             cmd_params = parse_command(command)
             for cmd_param in cmd_params:
                 cmd_param_steps_mapping[cmd_param].add(step_name)
@@ -612,7 +639,7 @@ def _validate_yadage_parameters(reana_yaml):
                     cmd_param_steps_mapping.update(publisher_cmd_param_steps_mapping)
                     if step_key == "script":
                         command = step_val
-                        _validate_dangerous_operations(command, step_name)
+                        _validate_dangerous_operations(command, step=step_name)
                     step_commands_params = parse_command_params(step_val)
                     for command_param in step_commands_params:
                         cmd_param_steps_mapping[command_param].add(step_name)
@@ -637,7 +664,7 @@ def _validate_yadage_parameters(reana_yaml):
     _warn_not_defined_parameters(cmd_param_steps_mapping, workflow_params, "yadage")
 
 
-def _validate_dangerous_operations(command, step):
+def _validate_dangerous_operations(command, step=None):
     """Warn the user if a command has dangerous operations.
 
     :param command: A workflow step command to validate.
@@ -645,10 +672,11 @@ def _validate_dangerous_operations(command, step):
     """
     for operation in COMMAND_DANGEROUS_OPERATIONS:
         if operation in command:
+            msg = 'Operation "{}" found in step "{}" might be dangerous.'
+            if not step:
+                msg = 'Operation "{}" might be dangerous.'
             display_message(
-                'Operation "{}" found in step "{}" might be dangerous.'.format(
-                    operation.strip(), step
-                ),
+                msg.format(operation.strip(), step if step else None),
                 msg_type="warning",
                 indented=True,
             )
--- a/tests/test_validate_parameters.py
+++ b/tests/test_validate_parameters.py
@@ -238,10 +238,11 @@ def test_validate_parameters_yadage(yadage_workflow_spec_loaded, capsys):
             "installation",
             '"cd /" found in step "installation" might be dangerous.',
         ),
+        ("sudo npm install", None, '"sudo" might be dangerous.',),
     ],
 )
 def test_validate_dangerous_operations(command, step, warning, capsys):
     """Validate if dangerous operations in a command trigger a warning."""
-    _validate_dangerous_operations(command, step)
+    _validate_dangerous_operations(command, step=step)
     captured = capsys.readouterr()
     assert warning in captured.out


reana_client/validation.py
```diff
@@ -521,6 +521,39 @@ def _validate_cwl_parameters(reana_yaml):
         )
         sys.exit(1)
 
+    dang_params = _validate_cwl_dangerous_operations(reana_yaml)
```
mvidalgarcia, line 524: We already have [`_validate_dangerous_operations`](https://github.com/reanahub/reana-client/blob/20dc10301b74766b3f9c27daba6608d1e65a3840/reana_client/validation.py#L640-L654), please use it to have consistent warning messages for all the workflow engines.

I see that for CWL the potential dangerous operations can be both in `baseCommand` and `arguments`. Some examples:
- https://github.com/reanahub/reana-demo-cdci-crab-pulsar-integral-verification/blob/2674d484cd2cb1ac8c00068e6bc53804be31b999/workflow/cwl/crab.cwl#L3-L7
- https://github.com/reanahub/reana-demo-root6-roofit/blob/a43db4aad1c3e1e2c3f7eb8d352169c0eba1beb6/workflow/cwl/fitdata.cwl#L20-L26

I wonder if we should also check `inputs` as they may contain some bash statements inside.. 🤔, although this wouldn't be a good practice at all so we can pass, it's users' resposibility.

So the idea would be simply passing the value of those fields to `_validate_dangerous_operations`, which will take care of doing the actual checks.

tests/test_validate_parameters.py
```diff
@@ -61,6 +62,15 @@ def get_loaded_yaml(step_spec, input_spec):
...
 
+    # Dangerous operation used in parameter
+    with runner.isolated_filesystem():
+        reana_yaml = get_loaded_yaml(
+            cwl_workflow_spec_step, cwl_workflow_spec_dangerous_input_param
+        )
+        _validate_cwl_parameters(reana_yaml)
+        captured = capsys.readouterr()
+        assert 'Operation "sudo python" found in "baseCommand"' in captured.out
+
```
mvidalgarcia, line 73: We probably don't need tests for this. We're already testing the functionality [here](https://github.com/reanahub/reana-client/blob/20dc10301b74766b3f9c27daba6608d1e65a3840/tests/test_validate_parameters.py#L243).

reana_client/validation.py
ParthS007: I am now using the `_validate_dangerous_operations` and warning are shown like as below. 

Screenshot 2021-03-18 at 12 14 50 PM

mvidalgarcia: Judging from your screenshot. `arguments` and `baseCommand` are not steps, steps are e.g. `gendata` and `fitdata` from `r-d-root6-roofit` example.

ParthS007: okay, Should I name them as `param` or `Operation "cd /" found in "arguments" might be dangerous` is enough?

mvidalgarcia: No, it's not enough. The idea here is that if the user has a workflow with tens of steps, is able to quickly identify where this dangerous operation is. Please check the output when introducing a dangerous operation for Serial or Yadage workflows in `r-d-root6-roofit`:

```console
$ reana-client validate -f reana-yadage.yaml               
==> Verifing REANA specification file... /Users/marco/code/reanahub/reana-demo-root6-roofit/reana-yadage.yaml
==> Verifing workflow parameters and commands... 
  -> WARNING: Operation "cd /" found in step "gendata" might be dangerous.
  -> WARNING: Operation "sudo" found in step "fitdata" might be dangerous.
```

The output should be exactly the same for CWL.

Take into account that a step might have no name, (e.g. `r-d-worldpopulation`). In that case, we show the number.

```console
$ reana-client validate                     
==> Verifing REANA specification file... /Users/marco/code/reanahub/reana-demo-worldpopulation/reana.yaml
==> Verifing workflow parameters and commands... 
  -> WARNING: Operation "sudo" found in step "0" might be dangerous.
```

ParthS007: The difference in CWL workflows is these `arguments` and `baseCommand` may/may not be be included in the steps. In `reana-demo-cdci-crab-pulsar-integral-verification` it is present in `workflow spec` directly not in any step.

mvidalgarcia: I see.. it seems that the cdci example is a bit special as it doesn't have a `Worlkflow` file but only a `CommandLineTool` one, which by checking the docs:

```
A CommandLineTool is a process characterized by the execution of a standalone, non-interactive program which is invoked on some input, produces output, and then terminates.
```

So it looks like a basic execution, no steps involved, which seems a bit weird to me.

So what we can do is making `step` optional in `_validate_dangerous_operations`, `None` by default. If `None`, we can omit the "found in step {}" part. So we can check if the workflow has steps, and pass the proper values. Please a a new test case for this in [test_validate_dangerous_operations](https://github.com/reanahub/reana-client/blob/20dc10301b74766b3f9c27daba6608d1e65a3840/tests/test_validate_parameters.py#L243).

ParthS007: As we see that this behavior is limited to only `cwl` workflow. what about sending `step=None` for CWL and keep as it is for other?

ParthS007: @mvidalgarcia  
Screenshot 2021-03-18 at 3 26 08 PM
How does it look now?

reana_client/validation.py
```diff
@@ -637,21 +656,33 @@ def parse_params(stages):
...
-            display_message(
-                'Operation "{}" found in step "{}" might be dangerous.'.format(
-                    operation.strip(), step
-                ),
-                msg_type="warning",
-                indented=True,
-            )
+        if isinstance(command, str) and operation in command:
+            _display_message(operation, step, step_flag)
+        if isinstance(command, list):
```
mvidalgarcia, line 680: A command should never be a list, could we move this logic to the `_check_dangerous_operations` inside CWL and leave this function more generic?

reana_client/validation.py
mvidalgarcia: > As we see that this behavior is limited to only `cwl` workflow. what about sending `step_flag=None` for CWL and keep as it is for other?

I don't see the benefit of adding a flag. The step itself can serve as a flag. If it's passed it means we have to display it on the message, if it's not passed, we don't show it.

reana_client/validation.py
```diff
@@ -521,6 +521,23 @@ def _validate_cwl_parameters(reana_yaml):
         )
         sys.exit(1)
 
+    def _check_dangerous_operations(workflow):
+        """Check for dang_keys in workflow and validate operations."""
+        [
+            _validate_dangerous_operations(v, k, step_flag=None)
+            for k, v in workflow.items()
+            if k in dang_keys
+        ]
```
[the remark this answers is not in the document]
ParthS007: Yes we are not returning anything. Initially, I added just plain for loop and if condition then I changed like this. Now I think you are right, I am not using the list anywhere so better keep it plain with list comprehension.

reana_client/validation.py
```diff
@@ -521,6 +521,29 @@ def _validate_cwl_parameters(reana_yaml):
...
+    def _check_dangerous_operations(workflow):
+        """Check for "baseCommand" and "arguments" in workflow.
+
+        If these keys are found, validate if they have dangerous operations.
+        """
+        for k, v in workflow.items():
+            if k in dang_keys:
+                if isinstance(v, list):
+                    for val in v:
+                        _validate_dangerous_operations(val, None)
```
mvidalgarcia, line 533: The idea here is to pass the step if it exists. Maybe we can pass `workflow.get('id')`, at least we would have some useful info for the user.

ParthS007: What would be the name of the `id` in the msg. `step`/`id` or something else?

ParthS007: If we pass the `id` from here, then we need the flag in `_validate_dangerous_operations` to check if it's coming from cwl to show the output msg accordingly.

mvidalgarcia: We can just treat it as the step, in the end is kind of similar.

reana_client/validation.py
```diff
@@ -521,6 +521,31 @@ def _validate_cwl_parameters(reana_yaml):
         )
         sys.exit(1)
 
+    def _check_dangerous_operations(workflow):
+        """Check for "baseCommand" and "arguments" in workflow.
+
+        If these keys are found, validate if they have dangerous operations.
+        """
+        dang_keys = ["baseCommand", "arguments"]
+        for k in workflow.keys():
```
mvidalgarcia, line 530: I meant we don't need to iterate over all keys, as we're only interested in `baseCommand` and `arguments`. Something like this:

```python
cmd_keys = ["baseCommand", "arguments"]
for cmd_key in cmd_keys:
    if cmd_key in workflow:
        cmd_values = (
            workflow[cmd_key]
            if isinstance(workflow[cmd_key], list)
            else [workflow[cmd_key]]
        )
        for cmd_value in cmd_values:
            _validate_dangerous_operations(str(cmd_value), step=workflow.get("id"))
```

mvidalgarcia (approved):
Looking good 👍

PR #483, opened 2021-03-23
cli: support glob patterns when listing files

### LINKED ISSUE #225: cli: regex/glob or similar for `reana-client ls`
for large workspaces it would be nice to either be able to glob / regex filter the files

```
reana-client ls 'mydir'
reana-client ls 'mydir/mysubdir'
reana-client ls 'mydir/mysubdir/*.root'
```

mvidalgarcia:
closes #225

#### To test:

```console
$ export REANA_WORKON=helloworld
$ reana-client run
$ reana-client ls
$ reana-client ls '**/*.py'
$ reana-client ls 'code/*'
$ touch code/test.txt && reana-client upload code/test.txt
$ reana-client ls '**/*.txt'
```

### DIFF
--- a/reana_client/cli/files.py
+++ b/reana_client/cli/files.py
@@ -66,11 +66,12 @@ def files_group(ctx):
     default=None,
     help="Get URLs of output files.",
 )
+@click.argument("filename", metavar="SOURCE", nargs=1, required=False)
 @add_access_token_options
 @add_pagination_options
 @click.pass_context
 def get_files(
-    ctx, workflow, _filter, output_format, access_token, page, size
+    ctx, workflow, _filter, output_format, filename, access_token, page, size
 ):  # noqa: D301
     """List workspace files.
 
@@ -79,8 +80,9 @@ def get_files(
     `--workflow` or `-w`.
 
     Examples: \n
-    \t $ reana-client ls --workflow myanalysis.42
-    """
+    \t $ reana-client ls --workflow myanalysis.42 \n
+    \t $ reana-client ls --workflow myanalysis.42 'code/\*'
+    """  # noqa: W605
     import tablib
     from reana_client.api.client import current_rs_api_client, list_files
 
@@ -93,7 +95,7 @@ def get_files(
     if workflow:
         logging.info('Workflow "{}" selected'.format(workflow))
         try:
-            response = list_files(workflow, access_token, page, size)
+            response = list_files(workflow, access_token, filename, page, size)
             headers = ["name", "size", "last-modified"]
             data = []
             file_path = get_path_from_operation_id(
--- a/reana_client/api/client.py
+++ b/reana_client/api/client.py
@@ -433,11 +433,12 @@ def delete_file(workflow, file_name, access_token):
         raise e
 
 
-def list_files(workflow, access_token, page=None, size=None):
+def list_files(workflow, access_token, file_name=None, page=None, size=None):
     """Return the list of files for a given workflow workspace.
 
     :param workflow: name or id which identifies the workflow.
     :param access_token: access token of the current user.
+    :param file_name: file name(s) (glob) to list.
     :param page: page number of returned file list.
     :param size: page size of returned file list.
     :returns: a list of dictionaries composed by the `name`, `size` and
@@ -447,6 +448,7 @@ def list_files(workflow, access_token, page=None, size=None):
         (response, http_response) = current_rs_api_client.api.get_files(
             workflow_id_or_name=workflow,
             access_token=access_token,
+            file_name=file_name,
             page=page,
             size=size,
         ).result()


reana_client/cli/files.py
```diff
@@ -66,11 +66,12 @@ def files_group(ctx):
     default=None,
     help="Get URLs of output files.",
 )
+@click.argument("filename", metavar="SOURCE", nargs=1, required=False)
```
audrium, line 69: Minor: for consistency this could be called `filenames` as in `download` and `upload` commands

audrium (approved):
Works great! :rocket: 

Couple of small questions:
-  `reana-client ls '*'` now returns a list of folders. I as a user would expect the whole list of files as in `reana client ls`. Or is it just me?
- Should we care about case sensitivity? Right now if there is a file 'test.JPG', we wouldn't return it with `*.jpg`

reana_client/cli/files.py
```diff
@@ -79,8 +80,9 @@ def get_files(
     `--workflow` or `-w`.
 
     Examples: \n
-    \t $ reana-client ls --workflow myanalysis.42
-    """
+    \t $ reana-client ls --workflow myanalysis.42 \n
+    \t $ reana-client ls --workflow myanalysis.42 'code/\*'
```
audrium, line 84:
```
reana ❯ reana-client ls --help
....
  Examples:
       $ reana-client ls --workflow myanalysis.42 'code/\*'
```
`'code/\*'` is printed to user which is not valid example

mvidalgarcia: The wildcard needs to be escaped, otherwise the readthedocs build fails as it interprets `*` as the beginning of a word in bold. More info [here](https://stackoverflow.com/questions/30454549/a-literal-in-restructuredtext). This is very unfortunate as the example is not working then.. But I can't find another solution. OTOH, it's displayed properly in RTD sit. We have the same in `reana-client rm`.

reana_client/cli/files.py
mvidalgarcia: In `download` and `upload` you can pass multiple files, whereas for `ls` you can only pass a single expression for simplicity. Check `nargs` value that differs in the different cases.

Maybe it's something to be covered in the future, support use cases like `reana-client ls -w workflow 'data/*' 'results/*'`?

mvidalgarcia:
> `reana-client ls '*'` now returns a list of folders. I as a user would expect the whole list of files as in `reana client ls`. Or is it just me?

Good point. Unfortunately, this is how [`pathlib.glob()`](https://docs.python.org/3/library/pathlib.html#pathlib.Path.glob) behaves..

A simple example to illustrate:

Starting point:
```console
/tmp/reana
├── bar
└── foo
    └── test.txt
```

Pseudologic present in `r-w-c`:
```py
In [1]: from pathlib import Path

In [2]: list(Path('/tmp/reana').glob('*'))
Out[2]: [PosixPath('/tmp/reana/foo'), PosixPath('/tmp/reana/bar')]
```

If you have any idea of how to obtain the same behaviour as in `reana-client ls` please let me know.

mvidalgarcia:
> Should we care about case sensitivity? Right now if there is a file 'test.JPG', we wouldn't return it with `*.jpg`

After running a quick test in my box, I see that `ls` is case sensitive.

```console
$ touch test1.png test2.PNG
$ ls *.png
test1.png
$ ls *.PNG
test2.PNG
```

Altough[ I think it is the SHELL (`zsh` in my case) behaving like that](https://unix.stackexchange.com/questions/48770/how-to-match-case-insensitive-patterns-with-ls). How does it behave in your end?

audrium:
> > `reana-client ls '*'` now returns a list of folders. I as a user would expect the whole list of files as in `reana client ls`. Or is it just me?
> 
> Good point. Unfortunately, this is how [`pathlib.glob()`](https://docs.python.org/3/library/pathlib.html#pathlib.Path.glob) behaves..
> 
> A simple example to illustrate:
> 
> Starting point:
> 
> ```
> /tmp/reana
> ├── bar
> └── foo
>     └── test.txt
> ```
> 
> Pseudologic present in `r-w-c`:
> 
> ```python
> In [1]: from pathlib import Path
> 
> In [2]: list(Path('/tmp/reana').glob('*'))
> Out[2]: [PosixPath('/tmp/reana/foo'), PosixPath('/tmp/reana/bar')]
> ```
> 
> If you have any idea of how to obtain the same behaviour as in `reana-client ls` please let me know.

I see.. I guess nothing better could be done than explicitly checking for the `*` value, which is not very nice solution.
Out of curiosity I've tried to change this [line](https://github.com/reanahub/reana-workflow-controller/pull/366/files#diff-9c016b5e33561473c8ffe0c925acb336235987daa8bdb8265e3b5d1e79cdc677R416) to:
```python
if is_directory(directory_path, secure_path) or secure_path == '*':
```
and it indeed works.. but it's ugly solution which I don't propose

audrium:
> > Should we care about case sensitivity? Right now if there is a file 'test.JPG', we wouldn't return it with `*.jpg`
> 
> After running a quick test in my box, I see that `ls` is case sensitive.
> 
> ```
> $ touch test1.png test2.PNG
> $ ls *.png
> test1.png
> $ ls *.PNG
> test2.PNG
> ```
> 
> Altough[ I think it is the SHELL (`zsh` in my case) behaving like that](https://unix.stackexchange.com/questions/48770/how-to-match-case-insensitive-patterns-with-ls). How does it behave in your end?

Same, as I also use `zsh` :) I guess that's a valid argument that all the shell commands (e.g, `grep`) is not case sensitive by default and usually needs some additional parameter (e.g. `-i` for `grep`) for this functionality

mvidalgarcia:
Superseded by https://github.com/reanahub/reana-client/pull/490

Issue #482, opened 2021-03-22
cli: list command does not incude deleted runs


tiborsimko:
**Current behaviour**

Consider the following workflow run list:

```console
$ reana-client list -v                                                                     
NAME    RUN_NUMBER   CREATED               STARTED               ENDED                 STATUS     ID                                     USER                                   SIZE
hello   1            2021-03-18T23:57:21   2021-03-18T23:57:22   2021-03-18T23:59:25   finished   49315fd8-04f1-4e41-b8c3-200b51a862ca   00000000-0000-0000-0000-000000000000   100G 
```

The user deletes the run:

```console
$ reana-client delete -w hello                                                                     
hello has been deleted
```

Whoops! The user noticed 100G and would like to delete the workspace too, so the user lists workflows again:

```console

$ reana-client list           
NAME   RUN_NUMBER   CREATED   STARTED   ENDED   STATUS
```

The list is empty, so there is apparently nothing more to delete... and 100G leave hanging out.

Note that the user would still be able to do:

```console
$ reana-client delete -w hello.1 --include-workspace 
hello.1 has been deleted
```

which would delete the workspace just fine.

IOW, it still works to delete the workspace, but users don't see anymore that they have some workflows (with big disk usage) that they could delete.

(But they could see those using web UI.)

**Expected behaviour**

- Option 1: Include also deleted workflow runs in the `list` output by default. Advantage: people will see all their runs as in the web UI. Disadvantage: the output may be long. Would need a CLI option such as `--exclude-deleted`. (See also notes about filtering.)

- Option 2: Do not change semantics of the `list` command, i.e. do not show deleted workflows by default; however add a new CLI option `--include-deleted` that would also include those deleted workflow runs. Pros: shorter list. Cons: mismatch with web UI defaults on REANA 0.7 release series.

- Option 3: BTW, when deleting a workflow, consider deleting also its workspace by default. Pros: this may be the most natural thing to do; when a user wants to delete a workflow, it is very probable that the user wants to _also_ delete all its file. Pros: no hanging workspaces, behaving according to expectations. Cons: this would change current semantics, so it may be dangerous, unless we add some  confirmation option such as `--force`.

**Notes**

[...]

PR #484, opened 2021-03-23
cli: list deleted workflows by default

ParthS007:
Add --exclude-deleted option to hide deleted workflows

Addresses #482 

Screenshot 2021-03-24 at 10 58 34 AM

### DIFF
--- a/reana_client/cli/workflow.py
+++ b/reana_client/cli/workflow.py
@@ -81,7 +81,11 @@ def workflow_execution_group(ctx):
     help="Get output in JSON format.",
 )
 @click.option(
-    "--all", "show_all", count=True, help="Show all workflows including deleted ones."
+    "--all",
+    "show_all",
+    count=True,
+    default=True,
+    help="Show all workflows including deleted ones.",
 )
 @click.option(
     "-v",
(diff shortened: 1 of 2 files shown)


reana_client/cli/workflow.py
```diff
@@ -81,7 +81,14 @@ def workflow_execution_group(ctx):
     help="Get output in JSON format.",
 )
 @click.option(
-    "--all", "show_all", count=True, help="Show all workflows including deleted ones."
+    "--all", "show_all", count=True, help="Show all workflows including deleted ones.",
+)
+@click.option(
```
tiborsimko, line 86: As discussed IRL, since we'll introduce filters in `master`, we can rely here on just flipping the default, so that CLI UI and Web UI would return the same thing on 0.7, and if somebody would like to filter deleted ones, they would be able to do something like:

```console
$ reana-client list | grep -v 'deleted$'
```

Please keep `--all` for backwards compatibility even if the default is true...

reana_client/cli/workflow.py
```diff
@@ -81,7 +81,11 @@ def workflow_execution_group(ctx):
     help="Get output in JSON format.",
 )
 @click.option(
-    "--all", "show_all", count=True, help="Show all workflows including deleted ones."
```
tiborsimko, line 84: Can you please add a line to the `CHANGES.rst` about this functionality change, for example:

- Changes ``list`` command to include deleted workflows by default.

tiborsimko (approved):

PR #485, opened 2021-03-24
cli: list deleted workflows by default

### LINKED ISSUE #482: cli: list command does not incude deleted runs

ParthS007:
Screenshot 2021-03-25 at 11 25 09 AM
Screenshot 2021-03-24 at 5 00 19 PM
closes #482

### DIFF
--- a/reana_client/cli/workflow.py
+++ b/reana_client/cli/workflow.py
@@ -82,9 +82,6 @@ def workflow_execution_group(ctx):
     help="Get output in JSON format.",
 )
 @click.option(
-    "--all", "show_all", count=True, help="Show all workflows including deleted ones."
-)
-@click.option(
     "-v",
     "--verbose",
     count=True,
@@ -115,7 +112,6 @@ def workflow_workflows(  # noqa: C901
     _format,
     output_format,
     access_token,
-    show_all,
     verbose,
     human_readable_or_raw,
     sort_columm_name,
@@ -176,8 +172,6 @@ def workflow_workflows(  # noqa: C901
         data = []
         for workflow in response:
             workflow["size"] = workflow["size"][human_readable_or_raw]
-            if workflow["status"] == "deleted" and not show_all:
-                continue
             name, run_number = get_workflow_name_and_run_number(workflow["name"])
             workflow["name"] = name
             workflow["run_number"] = run_number
--- a/setup.py
+++ b/setup.py
@@ -50,7 +50,7 @@ install_requires = [
     "rfc3987>=1.3.8",  # FIXME remove once yadage-schemas solves deps.
     "six==1.12.0",
     "strict-rfc3339>=0.7",  # FIXME remove once yadage-schemas solves deps.
-    "tablib>=0.12.1,<0.13",
+    "tablib>=0.14.0,<1.0.0",
     "webcolors==1.9.1",  # FIXME remove once yadage-schemas solves deps.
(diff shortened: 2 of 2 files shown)


reana_client/cli/workflow.py
```diff
@@ -81,9 +81,6 @@ def workflow_execution_group(ctx):
     default=None,
     help="Get output in JSON format.",
 )
-@click.option(
```
tiborsimko, line 84: Please keep it for backward compatibility, just update docstring that it is deprecated, that deleted worklows are now listed by default, etc.

ParthS007: ohh okay, I was the reading the issue again. I thought I missed something 😅

Issue #486, opened 2021-03-24
cli: validate workflow hints


tiborsimko:
**Current behaviour**

After the `validate` command was enriched in REANA 0.7.3, we could improve it even more.

For example, currently, the typos in compute backend clause are not detected:

```console
$ cd ~r/reana-demo-root6-roofit
$ git reset --hard 
$ sed -i 's,compute_backend,c_b,g' reana-htcondorcern.yaml
$ git diff            
diff --git a/reana-htcondorcern.yaml b/reana-htcondorcern.yaml
index affb577..a75372f 100644
--- a/reana-htcondorcern.yaml
+++ b/reana-htcondorcern.yaml
@@ -13,13 +13,13 @@ workflow:
     steps:
       - name: gendata
         environment: 'reanahub/reana-env-root6:6.18.04'
-        compute_backend: htcondorcern
+        c_b: htcondorcern
         htcondor_max_runtime: espresso
         commands:
         - mkdir -p results && root -b -q 'code/gendata.C(${events},"${data}")'
       - name: fitdata
         environment: 'reanahub/reana-env-root6:6.18.04'
-        compute_backend: htcondorcern
+        c_b: htcondorcern
         htcondor_max_runtime: espresso
         commands:
         - root -b -q 'code/fitdata.C("${data}","${plot}")'
$ reana-client validate --environments -f ./reana-htcondorcern.yaml
==> Verifying REANA specification file... /home/tibor/private/project/reana/src/reana-demo-root6-roofit/reana-htcondorcern.yaml
  -> SUCCESS: Valid REANA specification file.
==> Verifying workflow parameters and commands... 
  -> SUCCESS: Workflow parameters and commands appear valid.
==> Verifying environments in REANA specification file...
  -> SUCCESS: Environment image reanahub/reana-env-root6:6.18.04 has the correct format.
  -> SUCCESS: Environment image reanahub/reana-env-root6:6.18.04 exists locally.
  -> SUCCESS: Environment image reanahub/reana-env-root6:6.18.04 exists in Docker Hub.
  -> WARNING: Environment image UID is recommended to be 1000. UID 0 was found.
  -> SUCCESS: Environment image reanahub/reana-env-root6:6.18.04 has the correct format.
  -> SUCCESS: Environment image reanahub/reana-env-root6:6.18.04 exists locally.
  -> SUCCESS: Environment image reanahub/reana-env-root6:6.18.04 exists in Docker Hub.
[...]

PR #489, opened 2021-03-25
cli: validate workflow step clause values for serial workflows

ParthS007:
Screenshot 2021-03-26 at 12 08 00 PM

addresses #486

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,7 @@ Changes
 Version 0.7.4 (UNRELEASED)
 --------------------------
 - Fixes environment image validation info message where UIDs were switched.
+- Changes ``list`` command to include deleted workflows by default.
 
 Version 0.7.3 (2021-03-24)
 --------------------------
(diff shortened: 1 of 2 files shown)


mvidalgarcia (commented):
Just wondering, isn't this something that could we achieve via JSON Schema validation?

Not 100% sure but I think we can define the possible step clause names in our JSON schema and ensure that no other keys are used thanks to  [`"additionalProperties": false`](https://json-schema.org/understanding-json-schema/reference/object.html#properties).

ParthS007:
@mvidalgarcia  The `properties` are defined at different places in the yaml. Would it be possible to detect the properties in all three types of workflows?

mvidalgarcia:
> @mvidalgarcia The `properties` are defined at different places in the yaml. Would it be possible to detect the properties in all three types of workflows?

I was just focusing on Serial for now as this PR tackles that. I guess each workflow engine should have its own schema defined, as [Yadage has](https://github.com/yadage/yadage-schemas), and have its own way of calling the validator, e.g. https://github.com/reanahub/reana-client/blob/03460b791804e287063a7b4e56cb6bf7c7d78e19/reana_client/utils.py#L73-L75

I was just thinking if we could benefit from the JSON Schema validation, instead of _reinventing the wheel_ ourselves.

PR #495, opened 2021-04-20
client: allow spec path when creating wf from API

### LINKED ISSUE #494: client: error running workflows with spec file path using Python API
> The Pythonic API allows creating workflows programmatically https://blog.reana.io/posts/2021/reana-client-python-api. However, it only works by creating workflows from JSON specs. This is not compatible with all workflow engines, for example Yadage uses [workflow specification file paths only](https://github.com/reanahub/reana-workflow-engine-yadage/blob/e20432b8d2c20e8c35fb53e4444f98b9612808cb/reana_workflow_engine_yadage/cli.py#L73-L78) as opposed to e.g. [Serial where only JSON specs are accepted](https://github.com/reanahub/reana-workflow-engine-serial/blob/b9219f631f327b6e1f45e66dbc1f6c364ce0e973/reana_workflow_engine_serial/tasks.py#L50-L54).

> This can be achieved by allowing running from workflow file; manually upload workflow file with `file_upload`; use the rest of the existing API, `create_workflow` (to be tested, documented etc...)

_Originally written by @diegodelemos_

mvidalgarcia:
closes #494 
closes reanahub/docs.reana.io#23 

Updated Python API ref: https://reana-client--495.org.readthedocs.build/en/495/#module-reana_client.api.client

Working example: https://gist.github.com/mvidalgarcia/a1e61628dd67ed1a8b80d868684e1352

### DIFF
--- a/reana_client/api/client.py
+++ b/reana_client/api/client.py
@@ -158,18 +158,21 @@ def create_workflow(reana_specification, name, access_token):
 
 
 def create_workflow_from_json(
-    workflow_json,
     name,
     access_token,
+    workflow_json=None,
+    workflow_file=None,
     parameters=None,
     workflow_engine="yadage",
     outputs=None,
 ):
     """Create a workflow from json specification.
 
-    :param workflow_json: workflow specification in json format.
     :param name: name or UUID of the workflow to be started.
     :param access_token: access token of the current user.
+    :param workflow_json: workflow specification in json format.
+    :param workflow_file: workflow specification file path.
+                          Ignores ``workflow_json`` if provided.
     :param parameters: workflow input parameters dictionary.
     :param workflow_engine: one of the workflow engines (yadage, serial, cwl)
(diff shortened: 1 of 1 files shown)


diegodelemos (approved):
Works as expected.
```console
$ python yadage-wf.py
....
Current status:  queued
Current status:  running
...
Current status:  finished
```
image

My main worry is that now the `create_workflow_from_json` is not saying what it fully does. However, I think that the initial naming wasn't very accurate either so perhaps it is good to build this into it, and later we can think of renaming it alltogether.

mvidalgarcia:
> My main worry is that now the create_workflow_from_json is not saying what it fully does. However, I think that the initial naming wasn't very accurate either so perhaps it is good to build this into it, and later we can think of renaming it altogether.

True.. I kept the name for backward compatibility. For instance, the example described in the blog post works as it is. But I agree it's not ideal..

PR #516, opened 2021-05-10
dag: estimate initial workflow complexity for Yadage workflows

audrium:
closes https://github.com/reanahub/reana-server/issues/360

**NOTE**: this PR is included into https://github.com/reanahub/reana-server/pull/366

This PR adds Yadage workflow complexity estimator, which tries to:
- parse the workflow tree to filter out initial stages which will be launched on execution. Note that by default it's looking for `init` dependency, but some other can be passed down to constructor as `initial_step` value.
- It constructs internal representation of filtered tree which holds: `name, parameters, nested stages, memory limit, scatter parameters`
- As a second step it tries to populate the input parameters down the tree starting from the top. Also handles these kind of cases: `input_file: {step: init, output: files}` where `files` value should be taken from parent stage and used as `input_file`
- Third step is to calculate the complexity of the workflow taking into account scatter parameters and nested stages.

For bsm search example it should return the following:
```console
~/Documents/projects/reana-demo-bsm-search master*
reana ❯ rc validate
  -> INFO: Estimated workflow complexity:
[(8, '8Gi'), (5, '8Gi'), (2, '8Gi')]
```

### DIFF
--- a/reana_client/validation/complexity.py
+++ b/reana_client/validation/complexity.py
@@ -75,9 +75,32 @@ class ComplexityEstimatorBase:
 class SerialComplexityEstimator(ComplexityEstimatorBase):
     """REANA serial workflow complexity estimation."""
 
-    def parse_specification(self, initial_step):
+    def _parse_steps(self, steps):
         """Parse serial workflow specification tree."""
-        return {}
+
+        def _get_memory_limit(step):
+            # TODO: convert memory limit value to bytes (`kubernetes_memory_to_bytes`)
+            # TODO: validate memory limit value. Reuse code from (`set_memory_limit` in RJC)
+            # TODO: get `default_memory_limit` from config
+            default_memory_limit = os.getenv(
+                "REANA_KUBERNETES_JOBS_MEMORY_LIMIT", "8Gi"
+            )
+            return step.get("kubernetes_memory_limit", default_memory_limit)
+
+        tree = []
+        for idx, step in enumerate(steps):
+            name = step.get("name", str(idx))
+            memory_limit = _get_memory_limit(step)
+            tree.append({name: {"complexity": [(1, memory_limit)]}})
+        return tree
+
+    def parse_specification(self, initial_step):
+        """Parse and filter out serial workflow specification tree."""
+        spec_steps = self.specification.get("steps", [])
+        steps = self._parse_steps(spec_steps)
+        if initial_step == "init":
+            return steps[0] if steps else {}
+        return next(filter(lambda step: initial_step in step.keys(), steps), {})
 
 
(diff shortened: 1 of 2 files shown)


reana_client/validation/complexity.py
```diff
@@ -0,0 +1,211 @@
...
+                stage_value["complexity"] = complexity
+            return stages
+
+        return _parse_stages(self.steps)
+
+
+class CWLComplexityEstimator(ComplexityEstimatorBase):
+    """REANA CWL workflow complexity estimation."""
+
+    # TODO: implement
```
tiborsimko, line 211: Instead of TODO, please `raise NotImplemented`, or even better please return special value of empty list, which means "cannot detect complexity"... Having the same return type for CWL, Serial, Yadage workflows will allow to start implementing the `reana-server` workflow launcher logic.

reana_client/validation/complexity.py
```diff
@@ -0,0 +1,211 @@
...
+    :param workflow_type: A supported workflow specification type.
+    :param reana_yaml: REANA YAML specification.
+    """
+
+    def build_estimator(workflow_type, reana_yaml):
+        if workflow_type == "serial":
+            return SerialComplexityEstimator(reana_yaml)
+        if workflow_type == "yadage":
+            return YadageComplexityEstimator(reana_yaml)
+        if workflow_type == "cwl":
```
tiborsimko, line 28: You may want to use if/elif/elif/else to take care of possibly new workflow types, e.g. Snakemake is forthcoming...

reana_client/validation/complexity.py
```diff
@@ -0,0 +1,211 @@
...
+        "Estimated workflow complexity: \n{0}".format(complexity),
+        msg_type="info",
+        indented=True,
+    )
+
+
+class ComplexityEstimatorBase:
+    """REANA workflow complexity estimator base class."""
+
+    def __init__(self, reana_yaml, initial_step="init"):
```
tiborsimko, line 44: Optional: I wonder whether `initial_step` should be a parameter of the initaliser, or only of estimate_complexity method?  If the former, one has to initialise class twice for each run.  If the latter, one could initialise the class and launch several estimations for various starting steps. This could be useful for restarting workflow from given later step, for example.

Note: Since we are estimating only the workflow launching complexity, we are not probable to call this function several times... so it is fully OK for the code to go "as is".

mvidalgarcia:
Shall we close this one?

audrium:
The purpose of this PR was to test this functionality in isolation, but I guess now we can close it, since all the logic is already in https://github.com/reanahub/reana-server/pull/366

PR #517, opened 2021-05-19
config: import compute backends from r-commons

mvidalgarcia:
closes reanahub/reana#501


audrium (approved):
LGTM,

```
rc logs -w workflow --filter compute_backend=kubernetes
```
still works just fine :rocket:

PR #519, opened 2021-06-01
installation: unpin six

### LINKED ISSUE #504: un-pin `six`
is there a reason `six` is pinned to 1.12? I'm trynig to install `reana-client` in a docker image based on `FROM docker:20.10.5` which comes with `six 1.15` and this generates a conflict

tiborsimko:
Closes #504.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -20,7 +20,10 @@ history = open("CHANGES.rst").read()
 
 tests_require = [
     "pytest-reana>=0.7.1,<0.8.0",
-    'google-auth>=1.24.0; python_version=="2.7"',
+    'google-auth==1.24.0; python_version=="2.7"',
+    'oauthlib==3.1.0; python_version=="2.7"',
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -20,7 +20,10 @@
 
 tests_require = [
     "pytest-reana>=0.7.1,<0.8.0",
-    'google-auth>=1.24.0; python_version=="2.7"',
+    'google-auth==1.24.0; python_version=="2.7"',
+    'oauthlib==3.1.0; python_version=="2.7"',
+    'pluggy==0.13.1; python_version=="2.7"',
+    'websocket-client==0.59.0; python_version=="2.7"',
```
tiborsimko, line 26: Not pretty, but fixes py27 CI troubles... And since we don't use py27 much anymore except here, I guess it can stay this way...

ParthS007 (approved):
LGTM

PR #523, opened 2021-06-07
fix typo in example docstring

### LINKED ISSUE #520: cli: fix `ls` behaviour
One can get tracebacks, for example:

```console
$ reana-client ls -w hello       
NAME                    SIZE   LAST-MODIFIED      
code/helloworld.py      3253   2021-06-02T12:07:28
data/names.txt          20     2021-06-02T12:07:28
results/greetings.txt   34     2021-06-02T12:07:43
$ reana-client ls -w hello 'code' 
Something went wrong while retrieving file list for workflow hello:
3253 is not of type 'object'

Failed validating 'type' in schema['properties']['items']['items']['properties']['size']:
    {'properties': {'human_readable': {'type': 'string'},
                    'raw': {'type': 'number'}},
     'type': 'object'}

On instance['items'][0]['size']:
    3253
```

Also, the help gives an example but it is not working:
```console
$ reana-client ls --help
...
   reana-client ls --workflow myanalysis.42 'code/\*'
...
$ reana-client ls -w hello 'code/\*'
NAME   SIZE   LAST-MODIFIED
```

ParthS007:
closes #520

### DIFF
--- a/reana_client/cli/files.py
+++ b/reana_client/cli/files.py
@@ -91,8 +91,8 @@ def get_files(
 
     Examples: \n
     \t $ reana-client ls --workflow myanalysis.42 \n
-    \t $ reana-client ls --workflow myanalysis.42 --human-readable
-    \t $ reana-client ls --workflow myanalysis.42 'code/\*'
+    \t $ reana-client ls --workflow myanalysis.42 --human-readable \n
+    \t $ reana-client ls --workflow myanalysis.42 'data/*root*'
     """  # noqa: W605
     import tablib
     from reana_client.api.client import current_rs_api_client, list_files
@@ -370,7 +370,7 @@ def delete_files(ctx, workflow, filenames, access_token):  # noqa: D301
 
     Examples:\n
     \t $ reana-client rm -w myanalysis.42 data/mydata.csv \n
-    \t $ reana-client rm -w myanalysis.42 'code/\*'
+    \t $ reana-client rm -w myanalysis.42 'data/*root*'
     """  # noqa: W605
     from reana_client.api.client import delete_file
 


reana_client/cli/files.py
```diff
@@ -92,7 +92,7 @@ def get_files(
     Examples: \n
     \t $ reana-client ls --workflow myanalysis.42 \n
     \t $ reana-client ls --workflow myanalysis.42 --human-readable
-    \t $ reana-client ls --workflow myanalysis.42 'code/\*'
+    \t $ reana-client ls --workflow myanalysis.42 'code/'
```
mvidalgarcia, line 95: This is not a typo. Check https://github.com/reanahub/reana-client/pull/483#discussion_r616462092 for reference.

tiborsimko: Two cosmetic things:

(i) Yes we can keep the glob, but remove the extra printed backslash, because of single quotes there is no need to have it in the command (and it is actually wrong, since not giving results):

```console
$ reana-client ls -w helloworld-cwl-kubernetes 'code/\*'
NAME   SIZE   LAST-MODIFIED
$ reana-client ls -w helloworld-cwl-kubernetes 'code/*'
NAME                 SIZE   LAST-MODIFIED      
code/helloworld.py   3253   2021-06-07T08:31:22
```

(ii) there is a newline missing at the previous command which make the help not pretty right now:

```console
  Examples:

       $ reana-client ls --workflow myanalysis.42

       $ reana-client ls --workflow myanalysis.42 --human-readable      $
       reana-client ls --workflow myanalysis.42 'code/'
```

(we need `\n` after `--human-readable`)
```

Comment (ParthS007) [reply reana_client/cli/files.py]: If we want to keep `*` then a backslash is required.
More info: https://github.com/reanahub/reana-client/pull/483#discussion_r616462092

File: reana_client/cli/files.py [hunk 2]
```diff
@@ -82,7 +82,7 @@ def get_files(
     size,
     human_readable_or_raw,
 ):  # noqa: D301
-    """List workspace files.
+    r"""List workspace files.
```
tiborsimko, line 85: We cannot keep `r` unless we reformat the whole docstring, see what it would do to `reana-client ls --help`.

reana_client/cli/files.py
tiborsimko: The thing is that backslash is needed for sphinx docstring only, not for CLI input. So it would be confusing to show it to users. The challenge is to fix docstring formatting so that it would be both sphinx-friendly, pep257-friendly, and would show exactly what one needs to enter.

(In the worst of cases we could amend the example to ls `data/*root*` or some such, as it is both sphinx-friendly and also exact WRT CLI syntax)

tiborsimko: ... for example, the following patch on top of this PR would do as a minimal solution:

```diff
diff --git a/reana_client/cli/files.py b/reana_client/cli/files.py
index 3f14697..791c145 100644
--- a/reana_client/cli/files.py
+++ b/reana_client/cli/files.py
@@ -82,7 +82,7 @@ def get_files(
     size,
     human_readable_or_raw,
 ):  # noqa: D301
-    r"""List workspace files.
+    """List workspace files.
 
     The `ls` command lists workspace files of a workflow specified by the
     environment variable REANA_WORKON or provided as a command-line flag
@@ -92,7 +92,7 @@ def get_files(
     Examples: \n
     \t $ reana-client ls --workflow myanalysis.42 \n
     \t $ reana-client ls --workflow myanalysis.42 --human-readable \n
-    \t $ reana-client ls -w helloworld-cwl-kubernetes 'code/*'
+    \t $ reana-client ls --workflow myanalysis.42 'data/*root*'
     """  # noqa: W605
     import tablib
     from reana_client.api.client import current_rs_api_client, list_files
```

mvidalgarcia: There's another occurrence of `'code/\*'`. Should we change it as well?

https://github.com/reanahub/reana-client/blob/aa4fa1859b325f1e060ce26367b5d2bb53bfd0d1/reana_client/cli/files.py#L373

ParthS007: Yes, I think it would be good to change here as well. Will add it.

tiborsimko (approved):

reana_client/cli/files.py
tiborsimko: BTW please also change the second `rm` example to use `data/*root*`, not `code/*root*`, since ROOT files are for data.

Issue #500, opened 2021-04-23
`upload_to_server` fails on empty list


lukasheinrich:
passing an empty ilst to `upload_to_server` seems to fail 

would be great if this could be handled gracefully (ffor automation tools)

tiborsimko:
Currently we raise:

```
    537         raise ValueError("Workflow name or id must be provided")
    538     if not paths:
--> 539         raise ValueError(
    540             "Please provide path(s) to file(s) that " "should be uploaded to workspace."
    541         )
```

We could silently skip uploading if `paths` is a none-like object.

P.S. `def upload_to_server(workflow, paths, access_token)` in `reana_client/api/client.py`.

PR #526, opened 2021-06-14
cli: silently skip upload_to_server if paths is a none-like object

audrium:
closes https://github.com/reanahub/reana-client/issues/500

### DIFF
--- a/reana_client/api/client.py
+++ b/reana_client/api/client.py
@@ -499,9 +499,10 @@ def upload_to_server(workflow, paths, access_token):
     if not workflow:
         raise ValueError("Workflow name or id must be provided")
     if not paths:
-        raise ValueError(
-            "Please provide path(s) to file(s) that " "should be uploaded to workspace."
+        logging.info(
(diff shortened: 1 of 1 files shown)


mvidalgarcia:
Needs rebase.

reana_client/api/client.py
```diff
@@ -499,9 +499,10 @@ def upload_to_server(workflow, paths, access_token):
     if not workflow:
         raise ValueError("Workflow name or id must be provided")
     if not paths:
-        raise ValueError(
-            "Please provide path(s) to file(s) that " "should be uploaded to workspace."
+        logging.info(
+            "No path(s) to file(s) that should be uploaded to workspace was provided."
         )
+        return
```
mvidalgarcia, line 505: I see that with the value returned by this function we then iterate [here](https://github.com/reanahub/reana-client/blob/ae5180ed53d7a3a121402ab0ffe22d772afa93e4/reana_client/cli/files.py#L299). Maybe we should return an empty list then? (haven't tested it)

tiborsimko (approved):

PR #527, opened 2021-06-16
cli: add --filter option in du command

### LINKED ISSUE #522: cli: introduce `du` command file name filtering
**Current behaviour**

The `du` command allows to know the workspace size, but it does not allow to ask for size of subdirectories or for ROOT files.

**Expected behaviour**

Once we fix the `ls` command, see #520 and #521, we could use the same filtering technique to the `du` command.

So that one would be able to do things like:

```console
$ reana-client du -w myanalysis --filter name=mytmpdir
$ reana-client du -w myanalysis --filter name='\.root'
```

Depends on: #521.

ParthS007:
closes #522 

Screenshot 2021-06-25 at 12 44 59 PM

### DIFF
--- a/reana_client/cli/files.py
+++ b/reana_client/cli/files.py
@@ -16,6 +16,7 @@ import traceback
 import click
 from reana_commons.utils import click_table_printer
 
+from reana_client.printer import display_message
 from reana_client.api.utils import get_path_from_operation_id
 from reana_client.cli.utils import (
     add_access_token_options,
@@ -506,10 +507,19 @@ def move_files(ctx, source, target, workflow, access_token):  # noqa: D301
 @check_connection
 @add_access_token_options
 @click.option("-s", "--summarize", is_flag=True, help="Display total.")
+@click.option(
+    "--filter",
+    "filters",
+    multiple=True,
+    help="Filter results to show only files that match certain filtering "
+    "criteria such as file name or size."
+    "Use `--filter <columm_name>=<column_value>` pairs. "
+    "Available filters are `name` and `size`.",
+)
 @human_readable_or_raw_option
 @click.pass_context
 def workflow_disk_usage(
-    ctx, workflow, access_token, summarize, human_readable_or_raw
+    ctx, workflow, access_token, summarize, filters, human_readable_or_raw
 ):  # noqa: D301
     """Get workspace disk usage.
 
@@ -517,7 +527,8 @@ def workflow_disk_usage(
 
     Examples: \n
     \t $ reana-client du -w myanalysis.42 -s \n
-    \t $ reana-client du -w myanalysis.42 -s --human-readable
+    \t $ reana-client du -w myanalysis.42 -s --human-readable \n
+    \t $ reana-client du -w myanalysis.42 --filter name=data/
     """
     from reana_client.api.client import get_workflow_disk_usage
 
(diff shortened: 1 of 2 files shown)


reana_client/cli/files.py
```diff
@@ -506,30 +506,43 @@ def move_files(ctx, source, target, workflow, access_token):  # noqa: D301
 @check_connection
 @add_access_token_options
 @click.option("-s", "--summarize", is_flag=True, help="Display total.")
+@click.option(
+    "--filter",
+    "filters",
+    multiple=True,
+    help="Filter results to show only file sizes that match certain filtering "
```
audrium, line 513: Minor: I guess it would be more clear to say `Filter results to show only files that match certain filtering", WDYT?

reana_client/cli/files.py
```diff
@@ -506,30 +506,43 @@ def move_files(ctx, source, target, workflow, access_token):  # noqa: D301
...
 ):  # noqa: D301
     """Get workspace disk usage.
 
     The `du` command allows to chech the disk usage of given workspace.
 
     Examples: \n
     \t $ reana-client du -w myanalysis.42 -s \n
-    \t $ reana-client du -w myanalysis.42 -s --human-readable
+    \t $ reana-client du -w myanalysis.42 -s --human-readable \n
+    \t $ reana-client du -w myanalysis.42 --filter name=workflow
```
audrium, line 530: Minor: I think we could show how to filter directories in this example? So perhaps to use a name like `results` or `data` would be better instead of `workflow`?

reana_client/cli/files.py
```diff
@@ -506,30 +506,43 @@ def move_files(ctx, source, target, workflow, access_token):  # noqa: D301
...
     """
     from reana_client.api.client import get_workflow_disk_usage
 
     logging.debug("command: {}".format(ctx.command_path.replace(" ", ".")))
     for p in ctx.params:
         logging.debug("{param}: {value}".format(param=p, value=ctx.params[p]))
 
+    search_filter = None
```
audrium, line 538: Very minor: couldn't this default value be set in `click` argument definition?

reana_client/cli/files.py
```diff
@@ -506,30 +506,43 @@ def move_files(ctx, source, target, workflow, access_token):  # noqa: D301
...
+    headers = ["name", "size"]
```
audrium, line 539: Current output:
```console
reana ❯ reana-client du -w workflow
NAME    SIZE
34      ./results/greetings.txt
```
As you can see first column is `NAME` instead if `SIZE`. Headers should be changed to: `headers = ["size", "name"]`

audrium (commented):
Error message is not very clear If user mistypes filter value:
```console
$ reana-client du -w bsm --filter name=test
Disk usage could not be retrieved:
invalid literal for int() with base 10: 'copy.py'
```

reana_client/cli/files.py
ParthS007: `filters` remain None by default here. Here `search_filter` is the None since I am initiating the variable for referencing in the code if `filters` is not present.
- In case when there is no `filters` and It remains `None`
- In case when filters is present and we do not have any `search_filters` for headers `name` and `size`

audrium (approved):
Needs rebase

Works great!
```python
reana ❯ rc du -w workflow --filter name=code --filter size=3253
SIZE   NAME
3253   ./code/helloworld.py

reana ❯ rc du -w workflow --filter name=test
==> ERROR: No files matching filter criteria.

reana ❯ rc du -w workflow --filter test=test
==> ERROR: Filter test is not valid.

reana ❯ rc du -w bsm --filter name="*/read*" --filter size=4096
SIZE   NAME
4096   ./signal/read_0/_packtivity
4096   ./signal/read_1/_packtivity
4096   ./data/read_4/_packtivity
4096   ./data/read_0/_packtivity
4096   ./data/read_1/_packtivity
4096   ./data/read_3/_packtivity
4096   ./data/read_2/_packtivity

reana ❯ rc du -w bsm --filter name="*/steps.yml" --filter name="./plot/postfit.pdf"
SIZE    NAME
4919    ./workflow/steps.yml
19241   ./plot/postfit.pdf
```

audrium:
I guess what is still missing is the release commits for `reana-db` and `reana-commons` and then the dependencies update in all other repos

ParthS007:
Yes, We need a new release of `r-db`, `r-commons` and `r-server`.

PR #539, opened 2021-07-21
setup: stop supporting python 2

mvidalgarcia:


audrium (approved):
LGTM!

Perhaps we also need to change [this part](https://github.com/reanahub/reana-client/blob/1c05acbbe0760abd43a759bf38b48365b6860c0e/.github/workflows/ci.yml#L176-L181) and keep only `pip install requirements-builder` ?

P.S. `ciao-py2` :D

mvidalgarcia:
> Perhaps we also need to change [this part](https://github.com/reanahub/reana-client/blob/1c05acbbe0760abd43a759bf38b48365b6860c0e/.github/workflows/ci.yml#L176-L181) and keep only `pip install requirements-builder` ?

Oh indeed, well spotted!

PR #542, opened 2021-08-03
snakemake: add environment validation function

### LINKED ISSUE #540: snakemake: environment validation
**Current behavior**

Currently we support environment validation on the client-side via `reana-client validate --environments`. To get more familiar with this feature read [this blog post section](https://blog.reana.io/posts/2021/reana-0.7.3/#validating-workflow-environment-images), where the purpose of it is very well explained with some practical examples.

This feature is supported in Serial, Yadage and CWL workflows.

**Expected behavior**

The user should be able to run the very same command with Snakemake workflows, e.g. `reana-client validate --environments -f reana-snakemake.yaml`

All the environment validation logic resides in this [module](https://github.com/reanahub/reana-client/blob/master/reana_client/validation/environments.py), where each workflow engine has a class (`SerialEnvironmentValidator`, `YadageEnvironmentValidator` ...)

The main goal of each of these classes is basically parsing the workflow specification to grab the environment image and tag, to later pass them to the validator functions. We need to implement a `SnakemakeEnvironmentValidator` class with at least a `validate_environment` method that parses/loads the Snakefile and grabs the environment information.

Take [CWL](https://github.com/reanahub/reana-client/blob/863dfeb3bd828ea49b0e8ceea6f0e2e0bd65fcd7/reana_client/validation/environments.py#L453) as a reference, which uses some cwl utils (`cwl_parser`) to load the CWL spec and extract the images. We should investigate snakemake API (or codebase) to see how it loads and stores the Snakefile internally and see if it's viable to get the [environment image values (`container`)](validate_environment) from there.

This investigation will also be useful in the future for calculating the DAG complexity, as we will need to identify the multiple stages a workflow might have.

PS. We can remove [this check](https://github.com/reanahub/reana-client/blob/863dfeb3bd828ea49b0e8ceea6f0e2e0bd65fcd7/reana_client/validation/environments.py#L462-L467) as we don't support Python 2 anymore.

mariaFernando:
closes #540

### DIFF
--- a/reana_client/validation/environments.py
+++ b/reana_client/validation/environments.py
@@ -43,6 +43,11 @@ def validate_environment(reana_yaml, pull=False):
         if workflow_type == "cwl":
             workflow_file = workflow.get("file")
             return CWLEnvironmentValidator(workflow_file=workflow_file, pull=pull)
+        if workflow_type == "snakemake":
+            workflow_steps = workflow.rules()
+            return SnakemakeEnvironmentValidator(
+                workflow_steps=workflow_steps, pull=pull
+            )
 
     workflow = reana_yaml["workflow"]
     validator = build_validator(workflow)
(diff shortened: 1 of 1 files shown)


reana_client/validation/environments.py
```diff
@@ -43,6 +43,11 @@ def build_validator(workflow):
         if workflow_type == "cwl":
             workflow_file = workflow.get("file")
             return CWLEnvironmentValidator(workflow_file=workflow_file, pull=pull)
+        if workflow_type == "snakemake":
+            workflow_steps = workflow.rules()
```
mvidalgarcia, line 47: This statement fails for me with `AttributeError: 'dict' object has no attribute 'rules'`. `workflow` is a dict that has something like: 

```python
{
  'file': 'workflow/snakemake/Snakefile'
  'specification': {}
  'type': 'snakemake'
}
```
so it's not possible to call `rules` like that.

The idea here would be to load the Snakefile ( path in `workflow['file']) using some `snakemake` parsing function to later grab the multiple rules. Maybe there's some code to perform this load that you haven't pushed?

mvidalgarcia (commented):
When running `reana-client validate --environments -f reana-snakemake.yaml` I get the following output:

```console
...
==> Verifying environments in REANA specification file...
==> ERROR: Something went wrong when trying to validate /Users/marco/code/reanahub/reana-demo-helloworld/reana-snakemake.yaml
```

Can't you reproduce the same? Using the `reana-demo-helloworld` example.

mvidalgarcia:
Could you please remove [this check](https://github.com/reanahub/reana-client/blob/863dfeb3bd828ea49b0e8ceea6f0e2e0bd65fcd7/reana_client/validation/environments.py#L462-L467) as well in a separate commit?

The whole exception handling shouldn't be needed as we don't support py27 anymore.

mvidalgarcia:
Superseded by https://github.com/reanahub/reana-client/pull/547

Issue #546, opened 2021-08-17
validate: check desired workspace value


tiborsimko:
After https://github.com/reanahub/reana/issues/532, the administrators will be able to choose which locally mounted volumes are OK to be used for workflow workspace storage, for example `/var/reana/`, `/mybigextenaldisk/myworkflows`, or `/eos/home`.

After https://github.com/reanahub/reana-client/issues/545, the users will be able to select which workspace they would like to run a particular workflow on, for example `mybigextenaldisk/myworkflows`.

The goal of this issue is to extend validation to say whether a particular workspace value is allowed or not. For example, if a user would like to use `/usr/share/`, the `reana-client` command should report that this is not allowed, and the validation should fail.

Note that one REANA instance can allow one set of workspace prefixes, whilst another REANA instance can allow another set of workspace prefixes. Hence the list of allowed ones are not static, but depend on the given REANA cluster instance that the REANA client talks to.  This means that we may need a new REST API on the REANA server side so that client can interrogate the given cluster instance in order to discover which values are allowed and which are not. The new REST API endpoint can simple return list of allowed values for workspaces. We could (Alternatively, the workspace validation request could be dispatched to the server-side.)

PR #550, opened 2021-08-30
cli: add function to read available workspaces

marcdiazsan:
Closes reanahub/reana-client#546

### DIFF
--- a/reana_client/api/client.py
+++ b/reana_client/api/client.py
@@ -1014,3 +1014,32 @@ def list_secrets(access_token):
         raise Exception(e.response.json()["message"])
     except Exception as e:
         raise e
+
+
+def workspaces(access_token):
+    """List available workspaces.
+
+    :param access_token: access token of the current user.
+
+    """
+    try:
+        (response, http_response) = current_rs_api_client.api.workspaces().result()
+        if http_response.status_code == 200:
+            return response
+        else:
+            raise Exception(
+                "Expected status code 200 but replied with "
+                "{status_code}".format(status_code=http_response.status_code)
+            )
+
+    except HTTPError as e:
+        logging.debug(
+            "Workspaces could not be listed: "
+            "\nStatus: {}\nReason: {}\n"
+            "Message: {}".format(
+                e.response.status_code, e.response.reason, e.response.json()["message"]
+            )
+        )
+        raise Exception(e.response.json()["message"])
+    except Exception as e:
+        raise e
--- a/reana_client/cli/ping.py
+++ b/reana_client/cli/ping.py
@@ -81,3 +81,38 @@ def version(ctx):  # noqa: D301
     \t $ reana-client version
     """
     click.echo(__version__)
+
+
+@configuration_group.command("workspaces")
+@click.pass_context
+@add_access_token_options
+@check_connection
+def workspaces(ctx, access_token):  # noqa: D301
+    """List available workspaces.
+
+    The `workspaces` command lists all the available workspaces.
+    It also returns the default workspace defined by the admin.
+
+    Examples: \n
+    \t $ reana-client workspaces
+    """
+    try:
+        from reana_client.api.client import workspaces
+
+        response = workspaces(access_token)
+        click.echo(
+            click.style(
+                "List of available workspaces: {0}\n"
+                "Default workspace: {1}".format(
+                    ", ".join(response.get("workspaces_available")),
+                    response.get("default"),
+                )
+            )
+        )
+
+    except Exception as e:
+        logging.debug(traceback.format_exc())
+        logging.debug(str(e))
+        error_msg = "Could not list available workspaces:\n{0}".format(e)
+        click.echo(click.style(error_msg, fg="red"), err=True)
+        ctx.exit(1)
--- a/reana_client/cli/workflow.py
+++ b/reana_client/cli/workflow.py
@@ -914,8 +914,10 @@ def workflow_logs(
     help="If set, try to pull remote environment image from registry to perform "
     "validation locally. Requires ``--environments`` flag. [default=False]",
 )
+@add_access_token_options
+@check_connection
 @click.pass_context
-def workflow_validate(ctx, file, environments, pull):  # noqa: D301
+def workflow_validate(ctx, file, environments, pull, access_token):  # noqa: D301
     """Validate workflow specification file.
 
     The `validate` command allows to check syntax and validate the reana.yaml
@@ -930,6 +932,7 @@ def workflow_validate(ctx, file, environments, pull):  # noqa: D301
     try:
         load_reana_spec(
             click.format_filename(file),
+            access_token,
             skip_validate_environments=not environments,
             pull_environment_image=pull,
         )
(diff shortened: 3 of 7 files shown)


reana_client/api/client.py
```diff
@@ -1018,3 +1018,29 @@ def list_secrets(access_token):
         raise Exception(e.response.json()["message"])
     except Exception as e:
         raise e
+
+def workspaces(access_token):
+    """List available workspaces."""
+    try:
+        response, http_response = current_rs_api_client.api.workspaces(
+            access_token = access_token,
```
audrium, line 1026: `access_token` is passed, but [workspaces endpoint](https://github.com/reanahub/reana-server/pull/397/files) does not expect it. It should have [signin_required](https://github.com/reanahub/reana-server/blob/master/reana_server/rest/workflows.py#L64) decorator and [access_token](https://github.com/reanahub/reana-server/blob/master/reana_server/rest/workflows.py#L76-L81) parameter defined in openapi spec. But I see this command is placed near `ping` and `version`, if it's final place for it to be, maybe we don't need to require `access_token` for it?

marcdiazsan: You are right about not passing the access-token. I was thinking that only the authorized users with the token can see the available workspaces, but I am not sure, do you think is better without it?

reana_client/cli/ping.py
```diff
@@ -81,3 +81,38 @@ def version(ctx):  # noqa: D301
...
     """
     click.echo(__version__)
+
+
+@configuration_group.command("workspaces")
+@click.pass_context
+@add_access_token_options
+@check_connection
+def workspaces(ctx, access_token):  # noqa: D301
+    """List of available workspaces.
```
audrium, line 91: Very minor: I guess it supposed to be `List available workspaces.` here for consistency as defined in `docs/cmd_list.txt`

reana_client/cli/ping.py
```diff
@@ -81,3 +81,38 @@ def version(ctx):  # noqa: D301
...
+    """
+    try:
+        from reana_client.api.client import workspaces
+
+        response = workspaces(access_token)
+        click.echo(
+            click.style(
+                "List of available workspaces: {0}\n"
+                "Default workspace: {1}".format(
+                    ",".join(response.get("workspaces_available")),
```
audrium, line 107: Minor: empty space could be added after the comma. Here is what I get if there are multiple workspace paths available:
```
reana ❯ reana-client workspaces
List of available workspaces: /tmp/reana,/var/reana
Default workspace: /var/reana
```

reana_client/api/client.py
```diff
@@ -1018,3 +1018,28 @@ def list_secrets(access_token):
         raise Exception(e.response.json()["message"])
     except Exception as e:
         raise e
+
+
+def workspaces(access_token):
+    """List available workspaces."""
```
audrium, line 1024: Minor: `:param access_token: access token of the current user.` is missing from the docstring

audrium (commented):
In general a validation in `create_workflow()` endpoint is working nicely:
```
~/Documents/projects/reana-demo-worldpopulation master*
reana ❯ rc run
[INFO] Creating a workflow...
==> Verifying REANA specification file... /home/amecioni/Documents/projects/reana-demo-worldpopulation/reana.yaml
==> Verifying REANA specification parameters...
  -> SUCCESS: REANA specification parameters appear valid.
==> Verifying workflow parameters and commands...
  -> SUCCESS: Workflow parameters and commands appear valid.
==> Verifying dangerous workflow operations...
  -> SUCCESS: Workflow operations appear valid.
==> ERROR: Cannot create workflow :
==> ERROR: Workspace "/test" not valid.
```

I think we could also enrich `reana-client validate` function to include this type of validation so that user could be sure that everything is correct before starting the workflow

reana_client/cli/ping.py
marcdiazsan: BTW,  you agree that `workspaces` command should be in the same file as `ping` and `version` as a configurations command? Or is better to have it's own file and description in `docs/cmd_list.txt`?

audrium: IMHO it makes sense to have `workspaces` as a configurations command. It does not fit to any other category of commands and to create a separate one just for this one command I think would be too much. @tiborsimko , WDYT?

audrium:
Please note that there is a small `flake` issue, but the code looks good to me.
I guess the only thing which is missing for https://github.com/reanahub/reana-client/issues/546 is to enrich `reana-client validate` function to include `validate_workspace()`. Could be done in a separate PR as well

marcdiazsan:
Everything seems to be correct now with [this PR](https://github.com/reanahub/reana-client/pull/554). As I told you @audrium there is some other problem with `reana-client validation` command, as it cannot read the available workspaces as env variables. If merging is priority I think we can merge and solve in a separate PR

reana_client/utils.py
```diff
@@ -425,3 +441,28 @@ def run_command(cmd, display=True, return_output=False, stderr_output=False):
...
+    :param workspace: workspace path to be validated.
+    :param access_token: access token of the current user.
+
+    :raises ValidationError: Given workspace in REANA spec file does not validate against
+        allowed workspaces.
+    """
+    from reana_client.api.client import workspaces
+
+    available_workspaces = workspaces(access_token).get("workspaces_available", [])
+    try:
```
audrium, line 458: Perhaps here we could reuse [validate_workspace()](https://github.com/reanahub/reana-commons/pull/286/files#diff-0e9c97b2308d9b723f7552586946c165404f92f2ca51ed311d00f6159dd26514R14) from `reana-commons` and modify it to optionally accept workspace path list? Something like `validate_workspace(workspace_option, paths=WORKSPACE_PATHS)`. In that case we would have a central validation place where we pass the list from `reana-client` and and use the envar from `r-server`. WDYT?

reana_client/cli/workflow.py
```diff
@@ -914,8 +914,19 @@ def workflow_logs(
     help="If set, try to pull remote environment image from registry to perform "
     "validation locally. Requires ``--environments`` flag. [default=False]",
 )
+@click.option(
```
audrium, line 917: IMHO this flag is not needed as it seems that workspace path validation is a matter of general reana.yaml validation and it's not that expensive to be excluded as `environments`.

reana_client/utils.py
```diff
@@ -86,9 +87,11 @@ def load_workflow_spec(workflow_type, workflow_file, **kwargs):
 
 def load_reana_spec(
     filepath,
+    access_token,
     skip_validation=False,
     skip_validate_environments=True,
     pull_environment_image=False,
+    skip_validate_workspace=True,
```
audrium, line 94: Very minor: I guess this is a leftover

reana_client/utils.py
```diff
@@ -156,6 +158,13 @@ def _prepare_kwargs(reana_yaml):
             )
             validate_environment(reana_yaml, pull=pull_environment_image)
 
+        workspace = reana_yaml.get("workspace", {}).get("root_path")
+        if workspace:
+            display_message(
+                "Verifying workspace in REANA specification file...", msg_type="info",
+            )
+            _validate_workspace(reana_yaml["workspace"]["root_path"], access_token)
```
audrium, line 166: Very minor: probably `workspace` (as defined in line 161) can be used here instead of `reana_yaml["workspace"]["root_path"]`

audrium (approved):
Works nicely!

PR #554, opened 2021-09-06
ci: added system dependencies for graphviz support

audrium:
Right now CI is not passing in multiple PRs, because of the `graphviz` issues. 
Please have a look into https://github.com/reanahub/reana-client/pull/550/checks?check_run_id=3518962084 for error log.
This PR will fix this issue by installing some system dependencies for `graphviz` support.


mvidalgarcia (approved):
My bad, forgot about `r-client`, thanks for the fix.

PR #556, opened 2021-09-06
snakemake: added workflow types enum

### LINKED ISSUE #530: reana.yaml: support workflow `type: snakemake`
Enrich [JSON schema](https://github.com/reanahub/reana-client/blob/dc91b0e00196ef52cf3b8d4dd4c1b6793a5cfeb5/reana_client/schemas/reana_analysis_schema.json#L137-L141) to only allow the supported workflow types, including new `snakemake`. JSON Schema [enums](https://json-schema.org/understanding-json-schema/reference/generic.html?highlight=enum#enumerated-values) might come handy for this.

E.g. If one has `type: foo` in the reana.yaml, the validation should fail.

~Also update [this configuration](https://github.com/mvidalgarcia/reana-server/blob/3b541d1b790732cf9831036ece9322846e83d0cf/reana_server/config.py#L28) in `r-server`~

~Also update [this configuration](https://github.com/mvidalgarcia/reana/blob/35a12ac7ec9e27d9cc7ec2f10f7251171b21c4da/reana/config.py#L116) in `reana`.~

mariaFernando:
Closes #530

### DIFF
--- a/reana_client/utils.py
+++ b/reana_client/utils.py
@@ -79,9 +79,12 @@ def load_workflow_spec(workflow_type, workflow_file, **kwargs):
         "serial": serial_load,
         "snakemake": snakemake_load,
     }
-    """Dictionary to extend with new workflow specification loaders."""
 
-    return workflow_load[workflow_type](workflow_file, **kwargs)
+    """Dictionary to extend with new workflow specification loaders."""
+    load_function = workflow_load.get(workflow_type)
+    if load_function:
+        return load_function(workflow_file, **kwargs)
+    return {}
 
 
 def load_reana_spec(
--- a/reana_client/schemas/reana_analysis_schema.json
+++ b/reana_client/schemas/reana_analysis_schema.json
@@ -151,7 +151,8 @@
           "id": "/properties/workflow/properties/type",
           "type": "string",
           "title": "Workflow engine.",
-          "description": "Name which represents a supported workflow engine to be used to reproduce the analysis."
+          "description": "Name which represents a supported workflow engine to be used to reproduce the analysis.",
+          "enum": ["cwl","serial","yadage","snakemake"]
         },
         "resources": {
           "id": "/properties/workflow/properties/resources",


mvidalgarcia (commented):
I'm testing with the helloworld example and this diff:

```diff
diff --git a/reana.yaml b/reana.yaml
index 50d1bd8..0e6142f 100644
--- a/reana.yaml
+++ b/reana.yaml
@@ -9,7 +9,7 @@ inputs:
     outputfile: results/greetings.txt
     sleeptime: 0
 workflow:
-  type: serial
+  type: foo
   specification:
     steps:
       - environment: 'python:2.7-slim'
```

If I run the workflow I get this output:

```console
$ reana-client run
[INFO] Creating a workflow...
==> ERROR: Cannot create workflow :
'foo'   
```

Which is not very helpful for user.

It seems that the execution fails before reaching the schema validation [here](https://github.com/reanahub/reana-client/blob/3814b05465dc79ad682a14c84e3ae22db1b4eea8/reana_client/utils.py#L76-L81), as we're trying to access a key that doesn't exist. It never reaches [this point](https://github.com/reanahub/reana-client/blob/3814b05465dc79ad682a14c84e3ae22db1b4eea8/reana_client/utils.py#L149) where the actual JSON Schema validation is performed.

Same happens when we run `reana-client validate`, the message is not very descriptive.

```console
$ reana-client validate                                       
==> ERROR: Something went wrong when trying to validate /Users/marco/code/reanahub/reana-demo-helloworld/reana.yaml
```

Maybe we can slightly adapt the code not to fail in `load_workflow_spec` and instead fail later in the JSON Schema validation, thus printing a descriptive message?

reana_client/utils.py
```diff
@@ -79,8 +79,16 @@ def load_workflow_spec(workflow_type, workflow_file, **kwargs):
...
+    try:
+        """Dictionary to extend with new workflow specification loaders."""
+        load_function = workflow_load[workflow_type]
+    except Exception:
+        logging.warning(
+            'Something went wrong when loading the workflow type: "'
+            + workflow_type
+            + '"'
+        )
+        return
```
mvidalgarcia, line 91: The output is better but I don't think it's enough to inform the user about what's exactly wrong and what the allowed values are (workflow types).

```console
$ reana-client run                        
[INFO] Creating a workflow...
[WARNING] Something went wrong when loading the workflow type: "foo"
==> Verifying REANA specification file... /Users/marco/code/reanahub/reana-demo-helloworld/reana.yaml
==> ERROR: Cannot create workflow : 
None is not of type 'object'

Failed validating 'type' in schema['properties']['workflow']['properties']['specification']:
    {'id': '/properties/workflow/properties/specification',
     'title': 'Workflow specification in yaml format.',
     'type': 'object'}

On instance['workflow']['specification']:
    None
```

Plus, a different exception might be triggered inside the load function and be caught by this except block,as it's too broad.

Actually, the JSON Schema validation is [not being reached](https://github.com/reanahub/reana-client/blob/3814b05465dc79ad682a14c84e3ae22db1b4eea8/reana_client/utils.py#L189), if you see the error message it's pointing to the `specification` directive, which is `None` where it should be an object.

I think that a better solution would be returning an empty dict spec if the type is not among the valid ones and then we can reach the actual JSON Schema validation.

```diff
diff --git a/reana_client/utils.py b/reana_client/utils.py
index 8be18f7..8ec257b 100644
--- a/reana_client/utils.py
+++ b/reana_client/utils.py
@@ -80,8 +80,10 @@ def load_workflow_spec(workflow_type, workflow_file, **kwargs):
         "snakemake": snakemake_load,
     }
     """Dictionary to extend with new workflow specification loaders."""
-
-    return workflow_load[workflow_type](workflow_file, **kwargs)
+    load_function = workflow_load.get(workflow_type)
+    if load_function:
+        return load_function(workflow_file, **kwargs)
+    return {}
 
 
 def load_reana_spec(
```

mvidalgarcia (approved):
Works nicely 👍 

```console
$ reana-client create             
==> Verifying REANA specification file... /Users/marco/code/reanahub/reana-demo-helloworld/reana.yaml
==> ERROR: Cannot create workflow : 
'foo' is not one of ['cwl', 'serial', 'yadage', 'snakemake']

Failed validating 'enum' in schema['properties']['workflow']['properties']['type']:
    {'description': 'Name which represents a supported workflow engine to '
                    'be used to reproduce the analysis.',
     'enum': ['cwl', 'serial', 'yadage', 'snakemake'],
     'id': '/properties/workflow/properties/type',
     'title': 'Workflow engine.',
     'type': 'string'}

On instance['workflow']['type']:
    'foo'
```

Issue #562, opened 2021-09-24
validate: catch problems with misspelled `inputs.files` and similar directives


tiborsimko:
**Current behaviour**

The errors in directives such as `inputs`, `parameters` _etc_ are not caught:

```console
$ cd reana-demo-helloworld
$ sed -i 's,files:,fls:,g' reana.yaml 
$ git diff         
diff --git a/reana.yaml b/reana.yaml
index 50d1bd8..9597909 100644
--- a/reana.yaml
+++ b/reana.yaml
@@ -1,6 +1,6 @@
 version: 0.3.0
 inputs:
-  files:
+  fls:
     - code/helloworld.py
     - data/names.txt
   parameters:
@@ -19,5 +19,5 @@ workflow:
               --outputfile "${outputfile}"
               --sleeptime ${sleeptime}
 outputs:
-  files:
+  fls:
    - results/greetings.txt
$ reana-client validate
==> Verifying REANA specification file... /home/tibor/private/project/reana/src/reana-demo-helloworld/reana.yaml
  -> SUCCESS: Valid REANA specification file.
==> Verifying REANA specification parameters... 
  -> SUCCESS: REANA specification parameters appear valid.
==> Verifying workflow parameters and commands... 
  -> SUCCESS: Workflow parameters and commands appear valid.
==> Verifying dangerous workflow operations... 
  -> SUCCESS: Workflow operations appear valid.
```

This workflow appears to be valid, but it is actually not and it won't run properly.

**Expected behaviour**

Instead of seeing:

```console
-> SUCCESS: Valid REANA specification file.
```

one would expect to see:

```console
-> ERROR: The `reana.yaml` directive `inputs.fls` is not valid.
```

The `validate` should catch these kinds of error situations with `inputs.files`, `inputs.parameters` _etc_ for the user and  should end with an error exit status. (See also #561.)

Issue #404, opened 2020-03-25
schema: clean and align with reana.yaml docs


mvidalgarcia:
In the [reana analysis schema](https://github.com/reanahub/reana-client/blob/master/reana_client/schemas/reana_analysis_schema.json) there are outdated properties that should be removed if they are not used such as `metadata` and `environments`.
It should be consistent with what's explained in [the docs](http://docs.reana.io/reference/reana-yaml/).

mvidalgarcia:
We can consider adding type validation in the future - https://json-schema.org/understanding-json-schema/reference/type.html

Issue #220, opened 2018-12-12
schemas: workflow type should be restricted to the known values


diegodelemos:
Currently, the [workflow type in REANA yaml schema is a string](https://github.com/reanahub/reana-client/blob/master/reana_client/schemas/reana_analysis_schema.json#L133), we should limit it to be one of CWL, Yadata or Serial.

diegodelemos:
The database model should be updated accordingly, see https://github.com/reanahub/reana-db/issues/23.

mvidalgarcia:
Closed by https://github.com/reanahub/reana-client/commit/6575fa67d92164e2e23dfeea60e576b65eacc756

PR #564, opened 2021-09-27
validate: add validate-sections to validate misspelled directives or directives in reana.yaml

marcdiazsan:
Closes reanahub/reana-client#562

### DIFF
--- a/reana_client/schemas/reana_analysis_schema.json
+++ b/reana_client/schemas/reana_analysis_schema.json
@@ -6,91 +6,19 @@
   "title": "REANA analysis specification",
   "description": "Full analysis specification including data, sofware, environment and workflow enabling reproducibility on a REANA cluster.",
   "required": ["workflow"],
+  "additionalProperties": false,
   "properties": {
-    "environments": {
-      "id": "/properties/environments",
-      "type": "array",
-      "title": "Set of analysis' environments.",
-      "description": "All container images needed to reproduce the analysis.",
-      "items": {
-        "id": "/properties/environments/items",
-        "type": "object",
-        "optional": true,
-        "title": "Analysis environments.",
-        "description": "Analysis environments represented by container technology and image name.",
-        "properties": {
-          "image": {
-            "id": "/properties/environments/items/properties/image",
-            "type": "string",
-            "title": "Environment image id.",
-            "description": "String which represents an image used as environment for one or more steps of an analysis.",
-            "default": "busybox"
-          },
-          "type": {
-            "id": "/properties/environments/items/properties/type",
-            "type": "string",
-            "title": "Container technology.",
-            "description": "Name which represents a supported container technology to provide as a REANA environment.",
-            "default": "docker"
-          }
-        }
-      }
-    },
-    "metadata": {
-      "id": "/properties/metadata",
-      "type": "object",
-      "title": "Analysis metadata.",
-      "properties": {
-        "author": {
-          "id": "/properties/metadata/properties/author",
-          "type": "string",
-          "title": "Analysis author.",
-          "description": "User or group who created the analysis.",
-          "default": "@reanahub/developers"
-        },
-        "title": {
-          "id": "/properties/metadata/properties/title",
-          "type": "string",
-          "title": "Analysis title",
-          "default": "REANA analysis example"
-        }
-      }
-    },
-    "outputs": {
-      "id": "/properties/outputs",
-      "type": "object",
-      "title": "Analysis outputs.",
-      "properties": {
-        "files": {
-          "id": "/properties/outputs/properties/files",
-          "type": "array",
-          "title": "Analysis result files.",
-          "description": "Expected output from analysis represented by a set of files.",
-          "optional": true,
-          "items": {
-            "id": "/properties/outputs/properties/files/items",
-            "type": "string",
-            "title": "Relative path to the file."
-          }
-        },
-        "directories": {
-          "id": "/properties/outputs/properties/directories",
-          "type": "array",
-          "title": "Analysis result directories.",
-          "description": "Expected output from analysis represented by a set of directories.",
-          "optional": true,
-          "items": {
-            "id": "/properties/outputs/properties/directories/items",
-            "type": "string",
-            "title": "Relative path to the directory."
-          }
-        }
-      }
+    "version":  {
+      "id": "/properties/version",
+      "type": "string",
+      "title": "REANA version.",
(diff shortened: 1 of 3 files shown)


reana_client/utils.py
```diff
@@ -123,7 +123,11 @@ def _prepare_kwargs(reana_yaml):
     try:
         with open(filepath) as f:
             reana_yaml = yaml.load(f.read(), Loader=yaml.FullLoader)
-
+        try:
+            _validate_sections(reana_yaml)
```
mvidalgarcia, line 127: Instead of implementing our own validation, I think it's simpler and more scalable to enrich the [reana.yaml JSON Schema](https://github.com/reanahub/reana-client/blob/master/reana_client/schemas/reana_analysis_schema.json) and define a more strict structure. Then we can catch the errors and display them properly.
Check the JSON Schema docs to see how to restrict to certain properties and so on: https://json-schema.org/understanding-json-schema/reference/object.html#additional-properties

Related issues:
- https://github.com/reanahub/reana-client/issues/404
- https://github.com/reanahub/reana-client/issues/486
- https://github.com/reanahub/reana-client/issues/220

reana_client/utils.py
```diff
@@ -123,7 +124,6 @@ def _prepare_kwargs(reana_yaml):
     try:
         with open(filepath) as f:
             reana_yaml = yaml.load(f.read(), Loader=yaml.FullLoader)
-
         workflow_type = reana_yaml["workflow"]["type"]
```
mvidalgarcia, line 127: E.g. If we have `typ` instead of `type` we'll have an exception here without even reaching the JSON Schema validation (`validate(reana_yaml, reana_yaml_schema)`). Perhaps we can move the validation to be performed before and detect schema problems earlier?

```diff
diff --git a/reana.yaml b/reana.yaml
index 9fcfa12..4cf462c 100644
--- a/reana.yaml
+++ b/reana.yaml
@@ -11,7 +11,7 @@ inputs:
     year_min: 1500
     year_max: 2012
 workflow:
-  type: serial
+  typ: serial
   specification:
     steps:
       - environment: 'reanahub/reana-env-jupyter:2.0.0'
```

```console
$ reana-client validate    
==> ERROR: Something went wrong when trying to validate ~/code/reanahub/reana-demo-worldpopulation/reana.yaml
```
I have no hints on what is wrong.

mvidalgarcia: Following the IRL conversation, I meant something like this:

```diff
diff --git a/reana_client/utils.py b/reana_client/utils.py
index 5bc4312..e840648 100644
--- a/reana_client/utils.py
+++ b/reana_client/utils.py
@@ -124,7 +124,16 @@ def load_reana_spec(
     try:
         with open(filepath) as f:
             reana_yaml = yaml.load(f.read(), Loader=yaml.FullLoader)
-        workflow_type = reana_yaml["workflow"]["type"]
+        workflow_type = reana_yaml["workflow"].get("type")
+        if not skip_validation:
+            display_message(
+                "Verifying REANA specification file... {filepath}".format(
+                    filepath=filepath
+                ),
+                msg_type="info",
+            )
+            _validate_reana_yaml(reana_yaml)
+            validate_parameters(workflow_type, reana_yaml)
         reana_yaml["workflow"]["specification"] = load_workflow_spec(
             workflow_type,
             reana_yaml["workflow"].get("file"),
@@ -139,16 +148,6 @@ def load_reana_spec(
                     f, Loader=yaml.FullLoader
                 )
 
-        if not skip_validation:
-            display_message(
-                "Verifying REANA specification file... {filepath}".format(
-                    filepath=filepath
-                ),
-                msg_type="info",
-            )
-            _validate_reana_yaml(reana_yaml)
-            validate_parameters(workflow_type, reana_yaml)
-
         if not skip_validate_environments:
             display_message(
                 "Verifying environments in REANA specification file...",
@@ -203,6 +202,9 @@ def _validate_reana_yaml(reana_yaml):
         )
         raise e
     except ValidationError as e:
+        import wdb
+
+        wdb.set_trace()
         logging.info("Invalid REANA specification: {error}".format(error=e.message))
         if e.validator == "additionalProperties":
             invalid_section = (
```

So then the output is more informative:
```console
$ reana-client validate 
==> Verifying REANA specification file... /Users/marco/code/reanahub/reana-demo-worldpopulation/reana.yaml
==> ERROR: /Users/marco/code/reanahub/reana-demo-worldpopulation/reana.yaml is not a valid REANA specification:
==> ERROR:  The `reana.yaml` directive "typ" is not valid.
```

reana_client/utils.py
```diff
@@ -206,7 +204,19 @@ def _validate_reana_yaml(reana_yaml):
         raise e
     except ValidationError as e:
         logging.info("Invalid REANA specification: {error}".format(error=e.message))
-        raise e
+        if e.validator == "additionalProperties":
+            invalid_section = (
+                ".".join(e.path[1], e.message.split("'")[1])
+                if len(e.path) > 1
+                else e.message.split("'")[1]
+            )
```
mvidalgarcia, line 212: For simplicity, perhaps we can just use `e.message` here? Unless I'm missing something, the messages are quite descriptive:

> "Additional properties are not allowed ('typ' was unexpected)"

reana_client/utils.py
marcdiazsan: I agree to move the `_validate_reana_yaml` but `validate_parameters` would need to load the specification file. I would rather separate them

reana_client/utils.py
```diff
@@ -206,7 +204,9 @@ def _validate_reana_yaml(reana_yaml):
         raise e
     except ValidationError as e:
         logging.info("Invalid REANA specification: {error}".format(error=e.message))
-        raise e
+        raise REANAValidationError(
+            '==> ERROR:  The `reana.yaml` is not valid: "{0}".'.format(e.message)
+        )
```
mvidalgarcia, line 209: Couple of things:
- It'd be nice to keep the same colors for the error messages and rely on `display_message()` instead of appending ourselves the `"==> ERROR:"` prefix
- Error code is 0 when failing

image

reana_client/schemas/reana_analysis_schema.json
```diff
@@ -171,7 +100,60 @@
...
       },
-      "required": ["type"]
+      "oneOf": [
+        {
+            "required": ["type","file"]
+        },
+        {
+            "required": ["type","specification"]
+        }
+      ]
```
mvidalgarcia, line 110: Nice! 👍

mvidalgarcia (approved):
👍

PR #581, opened 2021-11-24
build: Require python >= 3.6 for install

matthewfeickert:
Add `python_requires` of `>=3.6` to ensure that installs from PyPI coming from users on **Python 2** get the last release that still had Python 2 support and not a broken (for Python 2) release with Python 3 only syntax.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -14,6 +14,7 @@ The list of contributors in alphabetical order:
 - `Jan Okraska <https://orcid.org/0000-0002-1416-3244>`_
 - `Leticia Wanderley <https://orcid.org/0000-0003-4649-6630>`_
 - `Marco Vidal <https://orcid.org/0000-0002-9363-4971>`_
+- `Matthew Feickert <https://orcid.org/0000-0003-4124-7862>`_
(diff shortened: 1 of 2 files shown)


matthewfeickert:
@tiborsimko @mvidalgarcia @audrium this is ready for review. Please let me know if you have any questions or concerns.

tiborsimko (approved):
Thanks @matthewfeickert! Just in time as we are releasing 0.8 final today :wink: I'll add you to `AUTHORS.rst` and repush.

mvidalgarcia:
Should propagate this change to `r-commons`, `r-db` and `pytest-reana` for future releases?

matthewfeickert:
> Just in time as we are releasing 0.8 final today

Thanks for getting this in in time for it @tiborsimko! :)

> Should propagate this change to `r-commons`, `r-db` and `pytest-reana` for future releases?

@mvidalgarcia yeah, that would be my recommendation. :+1:

tiborsimko:
Yeah, let's propagate it elsewhere as well. However, the end users are installing only `r-client`, never the others... And the developers are on py38+ already...So we do have plenty of time to progressively amend `r-commons` _et al_ after 0.8 is out.

PR #590, opened 2022-02-10
files: improve error messages

### LINKED ISSUE #408: download: nicer error messages
The error messages for download were improved in #405 already.

There are two corner cases left that we can prettify:

(A) When a user asks for not-existing workflow:

```console
$ reana-client download wrong.txt -w notexisting
File wrong.txt could not be downloaded: Error 404 Not Found REANA_WORKON is set to notexisting, but that workflow does not exist. Please set your REANA_WORKON environment variable appropriately.
```

Some new line separation between reasons would be nice to have.

(B) When a user has a wrong token, the situation is even less nice:

```console
$ export REANA_ACCESS_TOKEN=wrong
$ reana-client download plot.png -w something
File plot.png downloaded to /home/simko/private/project/reana/src/reana-client.
$  cat plot.png
{
  "message": "Add your REANA_ACCESS_TOKEN as a URL param to access the resource. RESOURCE_URL?access_token=REANA_ACCESS_TOKEN"
}
```

(stems from #405)

mdonadoni:
closes #408


reana_client/utils.py
```diff
@@ -48,6 +48,7 @@ def workflow_uuid_or_name(ctx, param, value):
             "environment variable",
             msg_type="error",
         )
+        sys.exit(1)
```
tiborsimko, line 51: I'm not sure we should `sys.exit()` in a `utils.py` file. It can be that somebody imports a utility function in Python API, and we would kind of force unexpectedly an exit in case of error situations. 

Granted, we can make sure that every utility function where `sys.exit()` is used, is only used in the CLI context -- such as `parse_secret_from_path()` is only used in the CLI context -- and then this can be OK. 

Just checked that all the utility functions are of this kind -- so this PR is OK :+1: 

Beginning of big parenthesis. However, it may be better to never do `sys.exit()` in a utility function, but simply reraise the original exception -- and leave the handling of it upon the user. In this way, such an exception would be caught in `cli/foo.py` and `sys.exit()`-ed there, and in Python API it could be caught differently without exiting. Might be more friendly towards third-party submit tools such as recast. That said, we have been having quite a few `sys.exit()` calls in `util.py` even before this patch! So if we go this way, we should improve those as well. (Noting that some utility functions such as `run_command()` can exit, because they are obviously called in the CLI context.) End of big parenthesis.

tiborsimko: ... but the exit code is still 0 and not 1 for me with this PR for some of the model commands?

```console
$ reana-client download results/greetings.txt.wrong -w helloworld-serial-kubernetes
==> ERROR: File results/greetings.txt.wrong could not be downloaded: Error 404 Not Found results/greetings.txt.wrong does not exist.
$ echo $?
0
```

tiborsimko (approved):

PR #592, opened 2022-02-16
cli: add option to write files to stdout

### LINKED ISSUE #509: add `reana-client cat`
sometimes it's convenient to be able to just dump the contents of a file to stdout (e.g. a JSON file) for this a `reana-client cat`  would be nice to have

mdonadoni:
Closes #509

### DIFF
--- a/reana_client/cli/files.py
+++ b/reana_client/cli/files.py
@@ -7,11 +7,13 @@
 # under the terms of the MIT License; see LICENSE file for more details.
 """REANA client output related commands."""
 
+import io
 import json
 import logging
 import os
 import sys
 import traceback
+import zipfile
 
 import click
 from reana_commons.utils import click_table_printer
@@ -28,7 +30,7 @@ from reana_client.cli.utils import (
     parse_filter_parameters,
     parse_format_parameters,
 )
-from reana_client.config import JSON, URL
+from reana_client.config import JSON, STD_OUTPUT_CHAR, URL
 from reana_client.errors import FileDeletionError, FileUploadError
 
 FILES_BLACKLIST = (".git/", "/.git/")
@@ -191,7 +193,8 @@ def get_files(
     "-o",
     "--output-directory",
     default=os.getcwd(),
-    help="Path to the directory where files will be downloaded.",
+    help="Path to the directory where files will be downloaded. "
+    "If ``-`` is specified as path, the files will be written to the standard output.",
 )
 @add_access_token_options
 @click.pass_context
@@ -207,9 +210,25 @@ def download_files(
 
     Examples: \n
     \t $ reana-client download # download all output files \n
-    \t $ reana-client download mydata.tmp outputs/myplot.png
+    \t $ reana-client download mydata.tmp outputs/myplot.png \n
+    \t $ reana-client download -o - data.txt # write data.txt to stdout
     """
     from reana_client.api.client import download_file, get_workflow_specification
+
+    def display_files_content(binary_file: bytes, multiple_files_zipped: bool) -> None:
+        """Write file(s) content to the standard output.
+
+        :param binary_file: The file(s) to print to the standard output.
+        :param multiple_files_zipped: Flag to determine if ``binary_file`` is a single
+            file or a zip archive containing multiple files.
+        """
+        if multiple_files_zipped:
+            with zipfile.ZipFile(io.BytesIO(binary_file)) as zip_file:
+                for entry in zip_file.infolist():
+                    if not entry.is_dir():
+                        click.echo(zip_file.read(entry), nl=False)
+        else:
+            click.echo(binary_file, nl=False)
 
     logging.debug("command: {}".format(ctx.command_path.replace(" ", ".")))
     for p in ctx.params:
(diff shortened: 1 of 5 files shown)


reana_client/cli/files.py
```diff
@@ -247,16 +250,27 @@ def download_files(
                     )
                 )
 
-                outputs_file_path = os.path.join(output_directory, file_name)
-                if not os.path.exists(os.path.dirname(outputs_file_path)):
-                    os.makedirs(os.path.dirname(outputs_file_path))
+                if output_directory == "-":
+                    if multiple_files_zipped:
```
mvidalgarcia, line 254: If it's a zip file, ie. multiple files to display, it would be nice to display each filename before its contents. Right now everything is mixed up so it's not easy to know which content corresponds to which file.

For example, executing this command in the roofit example returns the following:
```console
$ reana-client download -w roofit 'code/' -o -
#ifndef __CINT__                                                      
#include "RooGlobalFunc.h"                                            
#endif                                                                                                                                      
#include "RooRealVar.h"                                               
#include "RooDataSet.h"                                               
#include "RooGaussian.h"                                              
#include "RooChebychev.h"                                             
#include "RooAddPdf.h"                                                
#include "RooExtendPdf.h"                                             
#include "TCanvas.h"   
....
#ifndef __CINT__                                                      
#include "RooGlobalFunc.h"                                            
#endif                                                                
#include "RooRealVar.h"                                               
#include "RooDataSet.h"                                               
#include "RooGaussian.h"                                              
#include "RooChebychev.h"                                             
#include "RooAddPdf.h"                                                
#include "RooExtendPdf.h"                                             
#include "TCanvas.h"           
...
```

which corresponds to the 2 files that are in the `code` folder. However I don't know what's the correspondence of each content.

```console
$ reana-client ls -w roofit
NAME                SIZE     LAST-MODIFIED      
code/fitdata.C      1648     2022-02-17T10:43:09
code/gendata.C      1937     2022-02-17T10:43:09
results/data.root   154458   2022-02-17T10:43:23
results/plot.png    15450    2022-02-17T10:43:29
```

Maybe we can add some headers to differentiate.

```
==> File: code/fitdata.C
```

Not sure if this is the intention of the task though, or if we just want the raw content. But as a user is hard to understand the content of multiple files at the same time. cc/ @tiborsimko

mvidalgarcia (commented):
Please add a new entry in the changelog (`CHANGES.rst`)

```
Version 0.8.2 (UNRELEASED)
--------------------------

- Adds ... 
```

reana_client/cli/files.py
```diff
@@ -247,16 +250,27 @@ def download_files(
                     )
                 )
 
-                outputs_file_path = os.path.join(output_directory, file_name)
-                if not os.path.exists(os.path.dirname(outputs_file_path)):
-                    os.makedirs(os.path.dirname(outputs_file_path))
+                if output_directory == "-":
```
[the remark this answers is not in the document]
mdonadoni: I think this is a good idea, even though it is unlikely that we will change it in the future, given that the dash is widely used by command line utilities to refer to `stdin`/`stdout`.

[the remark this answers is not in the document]
mvidalgarcia: Yep, I agree that is very likely that it won't be changed, just to improve readability and perhaps reusability in other functions.

reana_client/cli/files.py
mdonadoni: I agree that showing an header for each file would make the output more clear. However, printing only the raw content of the files makes it possible to pipe the output of `reana-client` to another command line utility, for example:
```bash
reana-client download -o - results.json | jq ...
```
This is probably only interesting when dealing with a single file, so perhaps we could only show the headers when multiple files are being displayed?

mvidalgarcia: > This is probably only interesting when dealing with a single file, so perhaps we could only show the headers when multiple files are being displayed?

Definitely, that was my point :) hence my "If it's a zip file,...". I should have explained myself better.

tiborsimko: The goal of the issue was to emulate Unix `cat`, see the original issue description. I would go for simply concatenating the output files without any in-between headers. In this way, we can say that outputting to stdout is doing exactly what Unix `cat` is doing, and users will understand the behaviour by that analogy. That should do for MVP.

reana_client/cli/files.py
```diff
@@ -207,10 +210,20 @@ def download_files(
 
     Examples: \n
     \t $ reana-client download # download all output files \n
-    \t $ reana-client download mydata.tmp outputs/myplot.png
+    \t $ reana-client download mydata.tmp outputs/myplot.png \n
+    \t $ reana-client download -o - data.txt # write data.txt to stdout
     """
     from reana_client.api.client import download_file, get_workflow_specification
 
+    def display_files_content(binary_file, multiple_files_zipped):
```
mvidalgarcia, line 218: Could you please add a docstring here?

**Note.** Since we've dropped Python 2 support some months ago, we're starting to gradually include [type hints](https://docs.python.org/3/library/typing.html) to our functions.

In this case, you could use something like this:

```python
def display_files_content(binary_file: bytes, multiple_files_zipped: boolean) -> None:
```

mvidalgarcia (approved):
A couple of minor comments.

It works as expected 👍 
Docs PR is merged so the tests should pass here now.

PR #601, opened 2022-04-28
validation: switch to docker registry v2 API for environments validation

audrium:
Docker registry v1 API is deprecated: https://www.docker.com/blog/registry-v1-api-deprecation/
and seems to be missing some of the data:

```
https://index.docker.io/v1/repositories/reanahub/reana-env-root6/tags/6.18.04 <-- works
https://index.docker.io/v1/repositories/reanahub/reana-env-aliphysics/tags/vAN-20180614-1 <-- does not work
```
Using the new registry v2 API both cases works fine:
```
https://hub.docker.com/v2/repositories/reanahub/reana-env-root6/tags/6.18.04
https://hub.docker.com/v2/repositories/reanahub/reana-env-aliphysics/tags/vAN-20180614-1
```

PR also fixes an issue found while trying to get docker image UID. Gist of the issue is here:
```
docker run -i -t --rm reanahub/reana-env-root6:6.18.04 sh -c "/usr/bin/id -u" --> 0
docker run -i -t --rm reanahub/reana-env-aliphysics:vAN-20180614-1 sh -c "/usr/bin/id -u" --> uid=0(root) gid=0(root) groups=0(root)
```

### DIFF
--- a/reana_client/config.py
+++ b/reana_client/config.py
@@ -44,7 +44,7 @@ RUN_STATUSES = [
 ENVIRONMENT_IMAGE_SUSPECTED_TAGS_VALIDATOR = ["latest", "master", ""]
 """Warns user if above environment image tags are used."""
 
-DOCKER_REGISTRY_INDEX_URL = "https://index.docker.io/v1/repositories/{image}/tags/{tag}"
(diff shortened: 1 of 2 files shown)


VMois:
Let's split the PR into two commits. One commit is about dockerhub v2 API and another is to fix docker image UID.

PR #604, opened 2022-05-04
validation: check for whitespaces in the image specification

audrium:
To test apply these changes to the `helloworld` serial demo:

```diff
   type: serial
   specification:
     steps:
-      - environment: 'python:2.7-slim'
+      - environment: 'python:2.7-slim '
         commands:
           - python "${helloworld}"
               --inputfile "${inputfile}"
```

and do `reana-client validate --environments`. It should return:
```sh
...
-> ERROR: Environment image 'python:2.7-slim ' contains illegal characters.
```


tiborsimko (commented):
Can you enrich `CHANGES.rst` with something like:

- Fixes ``validate --environment`` command to detect illegal white space characters in image names.

PR #608, opened 2022-05-20
files: respect .reanaignore file during upload

### LINKED ISSUE #406: upload: upload full directory when files are not specified
Due to moving the workflow load logic from reana-client to the workflow-engine, we need all the workflow files in the workspace to be able to load the specification. 

**Current behavior**

When running `reana-client upload -w foo` we only upload the files specified in `inputs.files` and we have to manually upload what's in `workflow.file` and its dependencies.

**Expected behavior**

When running `reana-client upload -w foo`, upload all the files in the current `reana.yaml` directory.

**Extra considerations**
- Remove validation on `reana-client create`?
- Compare remote and local workspace before `reana-client start` and warn user if there're differences?
- Introduce the concept of `.reanaignore`. First check `.gitignore`, after `.reanaignore` and prevent uploading the present files.

VMois:
closes #406

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,7 @@ Changes
 Version 0.9.0 (UNRELEASED)
 --------------------------
 
+- Adds support for ``.reanaignore`` during file upload. Files that match ``.reanaignore`` will not be uploaded.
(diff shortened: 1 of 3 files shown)


mdonadoni (approved):
Nice work!

Could you please also update `CHANGES.rst` and rebase the PR?

PR #613, opened 2022-06-03
cli: add --show-deleted-runs option to list command

### LINKED ISSUE #612: cli: list --show-deleted-runs
**Current behaviour**

The `reana-client` is showing all runs by default, included deleted ones:

```console
$ reana-client list
NAME                                RUN_NUMBER   CREATED               STARTED               ENDED                 STATUS
root6-roofit-yadage-kubernetes      1            2022-06-01T15:07:49   2022-06-01T15:08:02   2022-06-01T15:09:10   finished
root6-roofit-snakemake-kubernetes   1            2022-06-01T15:06:58   2022-06-01T15:07:11   2022-06-01T15:07:41   deleted
root6-roofit-serial-kubernetes      1            2022-06-01T15:06:18   2022-06-01T15:06:30   2022-06-01T15:06:49   finished
root6-roofit-cwl-kubernetes         1            2022-06-01T15:05:37   2022-06-01T15:05:55   2022-06-01T15:06:11   finished
```

**Expected behaviour**

In https://github.com/reanahub/reana-ui/issues/218 we are switching to exclude deleted runs from Web UI by default:

show-deleted-runs

The toggle will be "off" by default and deleted runs won't be shown by default to the Web users.

This issue is about doing the same change for CLI users.

We can achieve the same UX by introducing a new option for the `list` command:

```console
$ reana-client list --show-deleted-runs
```

so that Web UI and CLI UI would behave fully consistently.

**Notes**

**(1)** Note that the new option `--show-deleted-runs` would work orthogonally alongside `--include-progress` and `--include-workspace-size`  options, since `--show-...` would operate on rows and `-include-...` would operate on columns of the output matrix.

**(2)** Note that we may need to take care about `--filter status=deleted` and how this would play with `--show-deleted-runs`. In Web UI, we solved it by not allowing the former until the latter is flipped. We can do the same here.

**(3)** Note that this is seemingly bringing the old `reana-client` behaviour back, before we switched to showing deleted runs by default, see https://github.com/reanahub/reana-ui/issues/218 for the past motivations. However, I think that these concerns can be addressed by doing two more changes:

- in CLI, `reana-client delete` would include by default also the workspace.
- in Web UI, we won't allow to "untick" option to preserve workspace in the following screenshot:

delete-also

In this way, the deletion of a workflow runs would always trigger the deletion of its workspace. If we give enough warnings to the user prior to performing deletions, I think that this should do. I cannot think of a use case where one would like to "delete" a run but "preserve" its workspace; it seems contradictory. I think we can afford removing this option.

VMois:
Desired behavior for this PR:

```bash
$ reana-client list  # do not show deleted workflows, show all other statuses
$ reana-client list --show-deleted-runs  # shows deleted workflows and all other statuses
$ reana-client list --all  # shows deleted workflows and all other statuses

$ reana-client list --filter name=myworkflow  # show workflows that match the name, do not show deleted workflows if any
$ reana-client list --filter name=myworkflow  --show-deleted-runs  # show workflows that match the name, show deleted workflows if any

$ reana-client list --filter status=deleted  # show deleted workflows, no need to use  "--show-deleted-runs"
```

How to test:

- check that PR behaves in the desired way

closes #612

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,8 @@ Version 0.9.0 (UNRELEASED)
 
 - Adds support for ``.reanaignore`` during file upload. Files that match ``.reanaignore`` will not be uploaded.
 - Adds support for ``.gitignore`` during file upload. Files that match ``.gitignore`` will not be uploaded.
+- Changes ``list`` to hide deleted workflows by default.
+- Changes ``list`` command to allow displaying deleted workflows via ``--all`` and ``--show-deleted-runs`` options.
 - Changes REANA specification loading and validation functionality by porting some of the logic to ``reana-commons``.
 - Fixes ``validate --environment`` command to detect illegal white space characters in image names.
 
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -19,3 +19,4 @@ The list of contributors in alphabetical order:
 - `Rokas Maciulaitis <https://orcid.org/0000-0003-1064-6967>`_
 - `Sinclert Perez <https://www.linkedin.com/in/sinclert>`_
 - `Tibor Simko <https://orcid.org/0000-0001-7202-5803>`_
+- `Vladyslav Moisieienkov <https://orcid.org/0000-0001-9717-0775>`_
--- a/reana_client/cli/utils.py
+++ b/reana_client/cli/utils.py
@@ -12,7 +12,7 @@ import json
 import os
 import shlex
 import sys
-from typing import Callable, NoReturn, Optional, Union
+from typing import Callable, NoReturn, Optional, List, Tuple, Union
 
 import click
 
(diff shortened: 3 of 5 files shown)


CHANGES.rst
```diff
@@ -4,8 +4,11 @@ Changes
 Version 0.9.0 (UNRELEASED)
 --------------------------
 
+- Adds new option ``--show-deleted-runs`` to ``list`` command to allow to display deleted workflows.
```
audrium, line 7: Minor: WDYT about this suggestion? IMO it's more in line with other statements, meaning that command goes first, then the option
```
- Changes ``list`` to hide deleted workflows by default.
- Changes ``list`` command to allow displaying deleted workflows via ``--all`` and ``--show-deleted-runs`` options.
```

audrium (commented):
Works nicely. But I don't know if we really need additional `--show-deleted-runs` option here? Right now `--all` and `--show-deleted-runs` are identical, there is no difference, perhaps `--all` would be enough if user wants to additionally display deleted runs?

PR #615, opened 2022-06-10
workflow: delete workspace by default

### LINKED ISSUE #614: cli: reana-client delete command would delete workspace too by default
Originated in https://github.com/reanahub/reana-client/issues/612#issue-1259662879

`reana-client delete` command would include by default also the workspace.

In this way, the deletion of a workflow runs would always trigger the deletion of its workspace. If we give enough warnings to the user prior to performing deletions, I think that this should do. I cannot think of a use case where one would like to "delete" a run but "preserve" its workspace; it seems contradictory. I think we can afford removing this option.

VMois:
closes #614

### DIFF
--- a/reana_client/api/client.py
+++ b/reana_client/api/client.py
@@ -670,8 +670,17 @@ def get_workflow_specification(workflow, access_token):
         raise e
 
 
-def delete_workflow(workflow, all_runs, workspace, access_token):
-    """Delete a workflow."""
+def delete_workflow(workflow, all_runs: bool, workspace: bool, access_token: str):
+    """Delete a workflow.
+
+    Please note that the workspace will be always deleted, even if ``workspace`` is set to ``False``.
+    """
+    if not workspace:
+        logging.warning(
+            "Parameter workspace=False was specified in delete_workflow() but workspace will still be deleted."
+        )
+    workspace = True
+
     try:
         parameters = {
             "all_runs": all_runs,
--- a/reana_client/cli/workflow.py
+++ b/reana_client/cli/workflow.py
@@ -1176,7 +1176,7 @@ def workflow_run(
 )
 @click.option(
     "--include-workspace",
-    "workspace",
+    "should_delete_workspace",
     is_flag=True,
     help="Delete workspace from REANA.",
 )
@@ -1184,23 +1184,23 @@ def workflow_run(
 @add_access_token_options
 @check_connection
 @click.pass_context
-def workflow_delete(ctx, workflow, all_runs, workspace, access_token):  # noqa: D301
+def workflow_delete(
+    ctx, workflow: str, all_runs: bool, should_delete_workspace: bool, access_token: str
+):  # noqa: D301
     """Delete a workflow.
 
-    The ``delete`` command allows to remove workflow runs from the database and
-    the workspace. By default, the command removes the workflow and all its
-    cached information and hides the workflow from the workflow list. Note that
-    workflow workspace will still be accessible until you use
-    ``--include-workspace`` flag. Note also that you can remove all past runs of
-    a workflow by specifying ``--include-all-runs`` flag.
+    The ``delete`` command removes workflow run(s) from the database.
+    Note that the workspace will always be deleted, even when ``--include-workspace`` is not specified.
+    Note also that you can remove all past runs of a workflow by specifying ``--include-all-runs`` flag.
 
     Example: \n
     \t $ reana-client delete -w myanalysis.42 \n
-    \t $ reana-client delete -w myanalysis.42 --include-all-runs \n
-    \t $ reana-client delete -w myanalysis.42 --include-workspace
+    \t $ reana-client delete -w myanalysis.42 --include-all-runs
     """
     from reana_client.api.client import delete_workflow
     from reana_client.utils import get_api_url
+
+    should_delete_workspace = True
 
     logging.debug("command: {}".format(ctx.command_path.replace(" ", ".")))
     for p in ctx.params:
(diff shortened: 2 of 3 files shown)


reana_client/cli/workflow.py
```diff
@@ -1144,15 +1144,17 @@ def workflow_run(
 )
 @click.option(
     "--include-workspace",
-    "workspace",
+    "should_delete_workspace",
     is_flag=True,
     help="Delete workspace from REANA.",
 )
```
mdonadoni, line 1150: Should we remove this option? cc @tiborsimko 
If we keep it, we should change the help message to inform the user that this option does not change the behaviour of reana-client anymore

mdonadoni (commented):
Works well!

Please note that the docstring of `workflow_delete` needs to be updated.

In any case, how do we inform the user of this change? In `reana-ui` we now display a message which informs the user that the workspace will be deleted. Considering `reana-client`, the user will not know about this change of behaviour before he actually deletes a workflow and its workspace. How should we handle this? One option would be to ask the user for confirmation, but this would break existing scripts that use `reana-client delete`.

cc @tiborsimko

reana_client/cli/workflow.py
VMois: I kept this option for backward compatibility. But, modifying the help message shouldn't be a problem.  I will do it.

VMois:
Another thing I forgot to change is our python API. I assume we need to force deletion there as well.

reana_client/cli/workflow.py
tiborsimko: Yes, let's keep it for backwards compatibility, and document that as of REANA 0.9 release series it is forced on by default, that we always delete the workspace when we delete the workflow.

> In any case, how do we inform the user of this change?

We'll make a publicity for this breaking feature during reana-client 0.9 release. For new we can simply document it in CHANGES.rst. It is not necessary to do anything in the code, since we do want to flip the defaults (and the power users agree).

VMois:
*question:* should I change r-server to force delete workspace as well? cc: @tiborsimko

reana_client/api/client.py
```diff
@@ -670,8 +670,10 @@ def get_workflow_specification(workflow, access_token):
         raise e
 
 
-def delete_workflow(workflow, all_runs, workspace, access_token):
-    """Delete a workflow."""
+def delete_workflow(workflow, all_runs: bool, workspace: bool, access_token: str):
+    """Delete a workflow. Forces to delete workspace."""
+    workspace = True
```
mdonadoni, line 675: I think it might be a good idea to print a warning when `workspace` is set to `False`, what do you think?

reana_client/api/client.py
```diff
@@ -670,8 +670,10 @@ def get_workflow_specification(workflow, access_token):
         raise e
 
 
-def delete_workflow(workflow, all_runs, workspace, access_token):
-    """Delete a workflow."""
+def delete_workflow(workflow, all_runs: bool, workspace: bool, access_token: str):
+    """Delete a workflow. Forces to delete workspace."""
```
mdonadoni, line 674:
```suggestion
    """Delete a workflow. Please note that the workspace will be always deleted, even if ``workspace`` is set to ``False``."""
```
WDYT?

reana_client/api/client.py
VMois: We do not do it with CLI so I am not sure.

mdonadoni: I think this is a little bit different from the CLI case. In the CLI case, the user cannot specify that he would like to keep the workspace (i.e. there is no `--keep-workspace` or a similar flag). Instead, in this case the user can call the function `delete_workflow(..., ..., False, ...)`: even though he has specified he would like to keep the workspace intact, we delete it anyway.

In this case, I think it makes sense to remove the `workspace` parameter.

What do you think about this? cc @tiborsimko

VMois: I added a warning to the `delete_workflow`. Ideally would be good to remove the `workspace` parameter if it is no longer required.

reana_client/cli/workflow.py
```diff
@@ -1144,40 +1144,40 @@ def workflow_run(
...
     """Delete a workflow.
 
-    The ``delete`` command allows to remove workflow runs from the database and
-    the workspace. By default, the command removes the workflow and all its
-    cached information and hides the workflow from the workflow list. Note that
-    workflow workspace will still be accessible until you use
-    ``--include-workspace`` flag. Note also that you can remove all past runs of
-    a workflow by specifying ``--include-all-runs`` flag.
+    The ``delete`` command removes workflow run(s) from the database.
+    It will always remove workspace.
```
mdonadoni, line 1161:
```suggestion
    Note that the workspace will always be deleted, even when ``--include-workspace`` is not specified.
```
WDYT?

audrium:
What is blocking this PR? 
PS. Also rebase is needed

reana_client/api/client.py
VMois: I would say let's just go with the warning message. PR has been stuck for a long time now.

VMois:
> What is blocking this PR?

This conversation - https://github.com/reanahub/reana-client/pull/615#discussion_r898082817.

tiborsimko:
> How should we handle this? One option would be to ask the user for confirmation, but this would break existing scripts that use `reana-client delete`.

Since this is more a "bug" rather than anything -- I mean wanting to delete a run but keeping its assets is a bit of a contradictory wish -- I think we have right to simply flip the default and emit a warning and make it clear in the docs and 0.9.0 release notes that deletion now includes the workspace by default. That is, do not introduce any checks in the code base and explain the change in the docs.

(That said, I wonder how many people alias `rm` to `rm -i` in their shells; we could think of such a "softer" deletion behaviour globally, outside of this ticket, as a kind of increased safety guard... One would be required to use something like `-f` to force deletion... Anyway we can merge this regardless, and think about possibly softening the deletion commands later before 0.9.0 is out.)

mdonadoni (approved):
> Anyway we can merge this regardless, and think about possibly softening the deletion commands later before 0.9.0 is out

Good! This PR needs https://github.com/reanahub/docs.reana.io/pull/129 to make all the CI checks pass

PR #624, opened 2022-08-16
workflow: always upload REANA specification on create and restart

### LINKED ISSUE #623: upload REANA specification file to server by default
Originated in https://github.com/reanahub/reana-client/issues/620#issuecomment-1211638448 (point 1)

We do not upload the REANA specification files to the server. Having REANA specifications in the workspace will allow for easier debugs and more reproducibility (if we decide to export workflows to, for example, Zenodo).

The focus of this issue is to upload specifications by default in reana-client. This will help with transitioning to https://github.com/reanahub/reana-db/issues/162.

In addition, we can also add a validation step to reana-server to check if reana.yaml is uploaded.

VMois:
closes #623

### DIFF
--- a/reana_client/version.py
+++ b/reana_client/version.py
@@ -14,4 +14,4 @@ This file is imported by ``reana_client.__init__`` and parsed by
 
 from __future__ import absolute_import, print_function
 
-__version__ = "0.9.0a7"
+__version__ = "0.9.0a8"


reana_client/cli/workflow.py
```diff
@@ -599,61 +611,70 @@ def workflow_restart(
         "restart": True,
     }
     if file:
+        specification_filename = click.format_filename(file)
         parsed_parameters["reana_specification"] = load_validate_reana_spec(
             click.format_filename(file)
         )
-    if workflow:
```
VMois, line 605: ``workflow`` must always exist, so I removed this check.

PR #632, opened 2022-09-22
cli: allow moving files while workflow is running

### LINKED ISSUE #630: cli: `mv` command displaying all the workspace files
Currently, when the user inputs a non-existing file as the target, the `mv` command displays all the workspace files of the given workflow. This is a problem in a realistic workspace, since this list will be very big.
For example:
```
$ reana-client mv -w root6-roofit.4 bad bad
==> ERROR: Source file(s) bad/* does not exist in workspace ['code/gendata.C', 'code/fitdata.C', 'results/plot.png', 'results/data.root']
$ reana-client mv bad bad -w h4l 2>&1 | wc -c
57767
```

As you can see, the latter is huge and too big to include in this issue. Instead, we should follow the way `reana-client-go` handled this and just print the workflow name, like this:
```
$ reana-client mv -w my-roofit-analysis bad bad
==> ERROR: source file(s) bad does not exist in my-roofit-analysis workspace
```

mdonadoni:
Closes reanahub/reana-workflow-controller#469
Closes #630


audrium (approved):
Works nicely

PR #634, opened 2022-10-03
workflow: show error when workflow fails after `start`

### LINKED ISSUE #629: cli: `start` showing success message when workflow fails
In the `start` command, when including the `--follow` flag, the program prints a success message when the workflow stops, even if it was stopped or failed, resulting in a message like `==> SUCCESS: workflow has been failed`.

We should improve this behavior and only display a success message when the workflow finished correctly. For the other cases, an error or info message is probably more better.

mdonadoni:
Closes #629

How to test:

1. Create a new demo workflow: `reana-client create -w test`
2. Start the workflow without uploading inputs: `reana-client start -w test --follow`

Expected result: the workflow fails because of missing files and reana-client prints `==> ERROR: test has been failed`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -195,7 +195,7 @@ jobs:
 
           - name: Install Python dependencies
             run: |
-              pip install -I -r .${{ matrix.testenv }}-requirements.txt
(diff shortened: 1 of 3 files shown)


audrium (approved):
Works nicely!

PR #635, opened 2022-10-10
cli: improve how reana spec is handled in `download` and `upload` commands

audrium:
After introducing [missing workflow spec schema to get_workflow_specification endpoint](https://github.com/reanahub/reana-server/pull/549), reana-client started to include and set all the missing properties to `None` in the response of this endpoint. This affected `upload` and `download` commands since internally this endpoint is used to retrieve the spec and get the files to upload/download. This PR fixes this issue

### DIFF
--- a/reana_client/cli/files.py
+++ b/reana_client/cli/files.py
@@ -250,10 +250,10 @@ def download_files(
             )
             sys.exit(1)
 
-        if "outputs" in reana_spec:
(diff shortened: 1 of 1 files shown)


mdonadoni (approved):

mdonadoni:
minor: what do you think about adding the description of the PR to the commit message?

PR #652, opened 2023-03-08
Add prune command

### LINKED ISSUE #651: new command: prune workflow's workspace
**Current behaviour**

When a user has consumed all the possible disk quota, there are several techniques that can help to liberate disk space:

- remove unnecessary workflow runs by deleting them;
- keep workflow runs but remove any unnecessary large temporary files in their workspaces;
- configure automated workspace file retention and deletion rules so that any temporary files will be garbage-collected by REANA itself.

See:

- https://docs.reana.io/advanced-usage/user-quotas/#what-to-do-when-quotas-are-exceeded
- https://blog.reana.io/posts/2022/workspace-file-retention-rules/

This works well, but it may be either a bit tedious to set up, or by deleting workflows manually one could loose possibly interesting information about which workflow runs succeeded, which runs failed, etc.

(The behaviour of REANA is unlike that of a typical CI, which deletes all artefacts unless declared otherwise, whilst REANA keeps all artefacts unless declared otherwise.)

**Expected behaviour**

It would be useful to offer one more way to the user to easily remove any unnecessary temporary files. Let's call it "pruning" the workspace.

We could introduce a new `prune` command:

```console
$ reana-client prune -w myanalysis.42
```

which, when run, will immediately remove all workspace files for the myanalysis run 42 _except for_ those inputs and outputs that are declared in the `reana.yaml` file.

In this way, the workspace would get rid of any large temporary files immediately, the inputs would be kept though so that workflow could eventually be relaunched any time if necessary, and the outputs would be kept too so that there would be no need to relaunch anything by default if the user only wants to access the results some time in the future.

**Notes**

- Note that some workflow-engine-produced output files, such as Yadage-produced workflow animated gif, or Snakemake-produced report, would be interesting to keep as well.

- The `prune` command could offer additional options such as `--include-all-runs` which would prune all runs of the given workflow; similarly to the behaviour of `reana-client delete`.

- We could also plug this option into the web UI in the hamburger menu and offer a new 'Prune workspace' option alongside the 'Delete workflow' option in the menu there.

giuseppe-steduto:
closes #651

### DIFF
--- a/docs/cmd_list.txt
+++ b/docs/cmd_list.txt
@@ -39,6 +39,7 @@ Workspace file management commands:
   du        Get workspace disk usage.
   ls        List workspace files.
   mv        Move files within workspace.
+  prune     Prune workspace files.
   rm        Delete files from workspace.
   upload    Upload files and directories to workspace.
 
--- a/reana_client/api/client.py
+++ b/reana_client/api/client.py
@@ -1091,3 +1091,32 @@ def get_workflow_retention_rules(workflow, access_token):
             )
         )
         raise Exception(e.response.json()["message"])
+
+
+def prune_workspace(workflow, include_inputs, include_outputs, access_token):
+    """Prune workspace files."""
+    try:
+        (response, http_response) = current_rs_api_client.api.prune_workspace(
+            workflow_id_or_name=workflow,
+            include_inputs=include_inputs,
+            include_outputs=include_outputs,
+            access_token=access_token,
+        ).result()
+
+        if http_response.status_code == 200:
+            return response
+        else:
+            raise Exception(
+                "Expected status code 200 but replied with "
+                "{status_code}".format(status_code=http_response.status_code)
+            )
+
+    except HTTPError as e:
+        logging.debug(
+            "Workspace could not be pruned: "
+            "\nStatus: {}\nReason: {}\n"
+            "Message: {}".format(
+                e.response.status_code, e.response.reason, e.response.json()["message"]
+            )
+        )
+        raise Exception(e.response.json()["message"])
--- a/reana_client/cli/files.py
+++ b/reana_client/cli/files.py
@@ -534,6 +534,56 @@ def move_files(ctx, source, target, workflow, access_token):  # noqa: D301
         sys.exit(1)
 
 
+@files_group.command("prune")
+@check_connection
+@add_workflow_option
+@add_access_token_options
+@click.option(
+    "--include-inputs",
+    "include_inputs",
+    is_flag=True,
+    help="Delete also the input files of the workflow. Note that this includes the workflow specification file.",
+)
+@click.option(
+    "--include-outputs",
+    "include_outputs",
+    is_flag=True,
+    help="Delete also the output files of the workflow.",
+)
+@click.pass_context
+def prune_files(
+    ctx, workflow, access_token, include_inputs, include_outputs
+):  # noqa: D301
+    """Prune workspace files.
+
+    The ``prune`` command deletes all the intermediate files of a given workflow that are not present
+    in the input or output section of the workflow specification.
+
+    Examples:\n
+    \t $ reana-client prune -w myanalysis.42\n
+    \t $ reana-client prune -w myanalysis.42 --include-inputs
+    """
+    from reana_client.api.client import prune_workspace
+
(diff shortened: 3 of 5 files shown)


reana_client/cli/files.py
```diff
@@ -555,7 +607,7 @@ def workflow_disk_usage(
 ):  # noqa: D301
     """Get workspace disk usage.
 
-    The ``du`` command allows to chech the disk usage of given workspace.
+    The ``du`` command allows to check the disk usage of given workspace.
```
tiborsimko, line 610: Please fix the same typo also in `reana-client-go` repository.  (But we can also do it later when porting the command to Golang client.)

Also, `docs.reana.io` client CLI API would need updating.

reana_client/cli/files.py
```diff
@@ -534,6 +534,58 @@ def move_files(ctx, source, target, workflow, access_token):  # noqa: D301
...
+    r"""
+    Prune the workspace.
+
+    The ``prune`` command deletes all the files in the workflow that are not present
+    in the input or output section of the workflow definition.
+
+    Examples:\n
+    \t $ reana-client prune -w myanalysis.42
+    \t $ reana-client prune -w myanalysis.42 --include-all-runs
+    \t $ reana-client prune -w myanalysis.42 --preserve-reports
```
tiborsimko, line 566: I think we should always preserve reports, if we are preserving workflow outputs.

If a user would like to prune everything except inputs, perhaps the option would be better called `--include-outputs`.  

We could also add another option called `--include-inputs` which would prune also inputs. This will allow to really wipe out everything from the workspace if somebody wants:

```console
$ reana-client prune -w myanalysis --include-inputs --include-outputs --include-all-runs
```

docs/cmd_list.txt
```diff
@@ -39,6 +39,7 @@ Workspace file management commands:
   du        Get workspace disk usage.
   ls        List workspace files.
   mv        Move files within workspace.
+  prune     Prune the workspace.
```
tiborsimko, line 42: "Prune workspace files" may be more in line with the other existing options.

We could also consider using somewhat longer version "Prune intermediate workspace files" but that is perhaps too long; and if we add two options `--include-inputs` and `--include-outputs` (see below) it may be good to keep the help rather generic (i.e. without "intermediate).

reana_client/api/client.py
```diff
@@ -1091,3 +1091,33 @@ def get_workflow_retention_rules(workflow, access_token):
             )
         )
         raise Exception(e.response.json()["message"])
+
+
+def prune_workspace(workflow, parameters, access_token):
```
mdonadoni, line 1096: What do you think about accepting as parameters `include_inputs` and `include_outputs` instead of the generic `parameters`?

mdonadoni (commented):
This works well, good job! I have just left a few small comments

reana_client/cli/files.py
```diff
@@ -534,6 +534,59 @@ def move_files(ctx, source, target, workflow, access_token):  # noqa: D301
...
 
+@files_group.command("prune")
+@check_connection
+@add_workflow_option
+@add_access_token_options
+@click.option(
+    "--include-inputs",
+    "include_inputs",
+    is_flag=True,
+    help="Also prune the input files associated with the workflow.",
```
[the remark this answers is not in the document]
giuseppe-steduto: More concise and clear :smile:

reana_client/cli/files.py
```diff
@@ -534,6 +534,59 @@ def move_files(ctx, source, target, workflow, access_token):  # noqa: D301
...
+)
+@click.option(
+    "--include-outputs",
+    "include_outputs",
+    is_flag=True,
+    help="Also prune the output files associated with the workflow.",
+)
+@click.pass_context
+def prune_files(ctx, workflow, access_token, include_inputs, include_outputs):
+    r"""
```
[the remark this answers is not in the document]
giuseppe-steduto: At first I added it because flake8 would mark the absence of the `r` as an error, but I have just realized I should include `# noqa: D301` to ignore that warning, similarly to how the other functions of the file do. Thanks for pointing that out!

mdonadoni (commented):
Could you please change the `added` in the commit messages into `add` and put `Closes #651` in the commit message?

Also, I think the two commits can be squashed together, I don't think we need to keep two different commits.

mdonadoni (approved):

PR #653, opened 2023-03-08
cli: fix docstring typo and update license

giuseppe-steduto:


LICENSE
```diff
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (C) 2017, 2018, 2019, 2020, 2021, 2022 CERN.
+Copyright (C) 2017, 2018, 2019, 2020, 2021, 2022, 2023 CERN.
```
tiborsimko, line 3: Please edit also `docs/conf.py` where we have `2017-2020` and use `2017-2023` there.

tiborsimko: BTW it is not necessary to split this in two commits; the addition of a new author and the update of the copyright year can come with the same commit as the feature.  (So that people wouldn't wonder who this new comitter is.)

reana_client/cli/files.py
```diff
@@ -555,7 +555,7 @@ def workflow_disk_usage(
 ):  # noqa: D301
     """Get workspace disk usage.
 
-    The ``du`` command allows to chech the disk usage of given workspace.
+    The ``du`` command allows to check the disk usage of given workspace.
```
tiborsimko, line 558: Please update also `CHANGES.rst` to add something like:

```
Fixed ``du`` command help message typo.
```

Even though such a release message is perhaps not very important, we could use this to practice our review habits :wink:

CHANGES.rst
```diff
@@ -18,6 +18,7 @@ Version 0.9.0 (2023-01-26)
 - Changes ``mv`` command to allow moving files while a workflow is running.
 - Changes ``upload`` command to prevent uploading symlinks.
 - Changes ``validation --environment`` command to use Docker registry v2 APIs to check that a Docker image exists in DockerHub.
+- Fixes ``du`` command help message typo.
```
tiborsimko, line 21: This line cannot go into the old release notes, because 0.9.0 has been already published... Please start a new section like:

```markdown
Version 0.9.1 (UNRELEASED)
--------------------------

- Fixes ``du`` command help message typo.
```

See the other repositories such as r-w-controller for example.

(And please squash the three commits together into one commit.)

tiborsimko (commented):
Thanks, please just use "cli: fix typo in du command docstring" in the commit message.

(The copyright update is kind of self-understood, we have to do it naturally with every commit when it applies...)

tiborsimko (approved):

PR #654, opened 2023-03-09
release: 0.9.1a1

mdonadoni:


reana_client/version.py
```diff
@@ -14,4 +14,4 @@
 
 from __future__ import absolute_import, print_function
 
-__version__ = "0.9.0"
+__version__ = "0.9.1"
```
audrium, line 17: Please note that it should be `0.9.1a1`

tiborsimko: Cosmetics: Please fix also the formatting in `CHANGES.rst` (missing newline after the UNRELEASED heading).

P.S. Do we need a new alpha release of the client already? Perhaps we can hold on with this and go directly for 0.9.1 when `prune` is implemented?

audrium (approved):
LGTM

PR #658, opened 2023-06-06
workflow: add message about autoclosure of the interactive session.

### LINKED ISSUE #657: interactive session: notify user of auto-closure
Administrators of a REANA cluster can specify a time limit after which inactive interactive sessions (such as Jupyter notebooks) will be automatically closed (https://github.com/reanahub/reana-workflow-controller/issues/498).

In https://github.com/reanahub/reana-ui/pull/312, the UI notification about the successful creation of an interactive session, was enriched with the inactivity time after which it will be automatically closed, if present.
However, this message is not shown in `reana-client` when creating an interactive session with `reana-client open`. We should include a similar message ("Please note that it will be automatically closed after X days of inactivity") also in the CLI client.

giuseppe-steduto:
When opening an interactive session with the `open` command, the user is now notified about the maximum inactivity time before the session is automatically closed, if the autoclosure was enabled by the REANA administrators.

Closes #657

### DIFF
--- a/reana_client/cli/workflow.py
+++ b/reana_client/cli/workflow.py
@@ -1361,7 +1361,7 @@ def workflow_open_interactive_session(
     Examples:\n
     \t $ reana-client open -w myanalysis.42 jupyter
     """
-    from reana_client.api.client import open_interactive_session
+    from reana_client.api.client import open_interactive_session, info
 
     if workflow:
         try:
@@ -1389,6 +1389,15 @@ def workflow_open_interactive_session(
             display_message(
                 "It could take several minutes to start the interactive session."
             )
+            reana_info = info(access_token)
+            max_inactivity_days_entry = (
+                reana_info.get("maximum_interactive_session_inactivity_period") or {}
+            )
+            max_inactivity_days = max_inactivity_days_entry.get("value")
+            if max_inactivity_days:
+                display_message(
+                    f"Please note that it will be automatically closed after {max_inactivity_days} days of inactivity."
+                )
         except Exception as e:
             logging.debug(traceback.format_exc())
             logging.debug(str(e))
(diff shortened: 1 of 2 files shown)


audrium (approved):
Looks good!

mdonadoni:
Works well, good job! 

PS: This PR needs to be rebased

mdonadoni (approved):

mdonadoni (changes requested):
There is an issue with this PR when connecting to an older version of REANA (e.g. reana.cern.ch)

```$ reana-client open -w workflow.1
==> SUCCESS: Interactive session opened successfully
https://reana.cern.ch/[...]
It could take several minutes to start the interactive session.
==> ERROR: Interactive session could not be opened: 
'NoneType' object has no attribute 'get'
```

We already had this issue in the past, see: https://github.com/reanahub/reana-client/pull/635. When adding a new field to an endpoint's response specification, bravado sets the field to `None` even if it is not present in the actual JSON response sent by the server. Thus, `reana_info.get("maximum_interactive_session_inactivity_period", {})` returns `None` instead of `{}`.

You should be able to reproduce this by installing the  ~~master~~ latest version from your PRs of reana-client and reana-commons in the same virtualenv.

We have two options:
1) Release a new alpha version of r-commons, bump its version in the `setup.py` of reana-client and fix the issue assuming that "maximum_interactive_session_inactivity_period" is always there
2) Fix the issue so that it works with both the old and the new (future) version of reana-commons

giuseppe-steduto:
> We have two options:
> 
> 1. Release a new alpha version of r-commons, bump its version in the `setup.py` of reana-client and fix the issue assuming that "maximum_interactive_session_inactivity_period" is always there
> 2. Fix the issue so that it works with both the old and the new (future) version of reana-commons

I went for option 2 since the fix was rather straightforward, as was done in #635, and added a small testcase.

mdonadoni (approved):
Great, all good now!

PR #662, opened 2023-08-09
utils: display warning messages about REANA spec validation

### LINKED ISSUE #660: reana.yaml: consider allowing unexpected properties
Up until now, the `reana.yaml` file has been validated by defining its JSON schema ([source](https://github.com/reanahub/reana-commons/blob/66821403d8b40a8929677b988c5a4262a13574e5/reana_commons/validation/schemas/reana_analysis_schema.json)). This validation is rather strict, as for example a `reana.yaml` file that contains additional properties not defined in the schema is considered invalid.

This helps us catch errors such as misspelled property names, but at the same time it limits a lot which changes we can make to the schema while being backward compatible with older reana-client versions. As an example, we have recently introduced the `tests` property, but we cannot update the workflow demos we have on GitHub as they will not run when using an older version of reana-client.

We should investigate whether we should make validation less strict, by reporting unexpected properties as warnings instead of errors. We should also investigate whether we should expose these warnings in the UI, for example when launching workflow with Launch-on-REANA.

giuseppe-steduto:
Closes #660

### DIFF
--- a/reana_client/validation/utils.py
+++ b/reana_client/validation/utils.py
@@ -51,12 +51,39 @@ def validate_reana_spec(
             f"Verifying REANA specification file... {filepath}",
             msg_type="info",
         )
-        validate_reana_yaml(reana_yaml)
-        display_message(
-            "Valid REANA specification file.",
-            msg_type="success",
-            indented=True,
-        )
+        validation_warnings = validate_reana_yaml(reana_yaml)
+        if validation_warnings:
+            display_message(
+                "The REANA specification appears valid, but some warnings were found.",
+                msg_type="warning",
+                indented=True,
+            )
+        for warning_key, warning_value in validation_warnings.items():
+            if warning_key == "additional_properties":
+                # warning_values is a list of unexpected properties
+                message = (
+                    f"Unexpected properties found in REANA specification file: "
+                    f"{', '.join(warning_value)}"
+                )
+            else:
+                message = warning_value
+            display_message(
+                message,
+                msg_type="warning",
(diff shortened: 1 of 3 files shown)


reana_client/validation/utils.py
```diff
@@ -51,12 +51,35 @@ def validate_reana_spec(
...
+            display_message(
+                message,
+                msg_type="warning",
+                indented=True,
+            )
+        if validation_warnings:
+            display_message(
+                "The REANA specification file is valid, but some warnings were "
+                "found. Please check the warnings above and make sure that the "
+                "REANA specification file is correct.",
```
mdonadoni, line 73: Given that we already print the warnings before, maybe we can shorten this to just say _Please make sure that the REANA specification file is correct_?
Or we could also print something like this:
```
  -> WARNING: Issues found while validating the REANA specification file
  -> WARNING: [... list of warnings ...]
  -> WARNING: Please make sure that the REANA specification file is correct
```

In any case, what I would like to avoid is to print a very long line in the terminal.

What do you think?

reana_client/validation/utils.py
mdonadoni, line 1: Maybe let's add a line to CHANGES.rst?

reana_client/validation/utils.py
giuseppe-steduto: Okay for shortening the single lines printed in the terminal! I think it is still important to write that the REANA specification file appears valid.

mdonadoni (approved):

PR #669, opened 2023-08-16
docs: autoload client methods and add missing parameter descriptions

### LINKED ISSUE #665: docs: add missing client methods
There are many methods from `reana_client.api.client` missing in the docs, due to them not being explicitly listed in `docs/index.rst`:
https://github.com/reanahub/reana-client/blob/4aeaa2cadfc2fbc8d6879a9179763d2d5bc12003/docs/index.rst?plain=1#L24

Ideally the docs should discover all the methods automatically, if not possible we should at least add the missing ones.

giuseppe-steduto:
Configure Sphinx to autoload all the public methods from `api.client`
and add a small descriptions of the parameters of a function where it
was missing.

Closes #665.


reana_client/api/client.py
```diff
@@ -166,7 +184,12 @@ def get_workflow_status(workflow, access_token):
 
 
 def create_workflow(reana_specification, name, access_token):
-    """Create a workflow."""
+    """Create a workflow.
+
+    :param reana_specification: An object representing the REANA specification of the workflow.
```
mdonadoni, line 189:
```suggestion
    :param reana_specification: a dictionary representing the REANA specification of the workflow.
```
What do you think? (note also the lowercase `a` instead of `An`)

reana_client/api/client.py
```diff
@@ -896,7 +939,12 @@ def mv_files(source, target, workflow, access_token):
 
 
 def get_workflow_disk_usage(workflow, parameters, access_token):
-    """Display disk usage workflow."""
+    """Display disk usage workflow.
+
+    :param workflow: name or id which identifies the workflow.
```
mdonadoni, line 944:
```suggestion
    :param workflow: name or id of the workflow.
```
What do you think about this, to be consistent with the rest?

reana_client/api/client.py
```diff
@@ -896,7 +939,12 @@ def mv_files(source, target, workflow, access_token):
...
+    :param parameters: additional input parameters and operational options.
```
mdonadoni, line 945: This should be fixed, as these are parameters to customise the return value and they are not workflow parameters. We should probably fix this also in the OpenAPI spec.

reana_client/api/client.py
```diff
@@ -1094,7 +1145,13 @@ def get_workflow_retention_rules(workflow, access_token):
 
 
 def prune_workspace(workflow, include_inputs, include_outputs, access_token):
-    """Prune workspace files."""
+    """Prune workspace files.
+
+    :param workflow: name or id of the workflow.
+    :param include_inputs: whether to also delete inputs.
+    :param include_outputs: whether to also delete outputs.
```
mdonadoni, line 1152: We should decide between `whether ...` or `if ``True``, ...` and use that everywhere to describe boolean parameters

reana_client/api/client.py
```diff
@@ -928,10 +976,10 @@ def add_secrets(secrets, overwrite, access_token):
...
 
     :param secrets: dictionary containing all the secrets to be sent.
-      The dictionary with secret names for keys and for each key there is
-       a dictionary with two fields:
-      - 'value':  a base64 encoded file or literal string
-      - 'type': 'file' or 'env'
+        The dictionary with secret names for keys and for each key there is
+        a dictionary with two fields:
+        - 'value':  a base64 encoded file or literal string
+        - 'type': 'file' or 'env'
```
mdonadoni, line 982: This is not displayed as a list: https://reana-client--669.org.readthedocs.build/en/669/#reana_client.api.client.add_secrets

What about either making the list work if possible, or rephrasing this to avoid the bullet points?

reana_client/api/client.py
```diff
@@ -33,7 +33,10 @@
 
 
 def ping(access_token):
-    """Health check REANA server."""
+    """Health check REANA server.
```
mdonadoni, line 36: What do you think about rephrasing this to start the phrase with a verb, as for all the other docstrings?

docs/conf.py
```diff
@@ -33,6 +36,11 @@
 #
 # needs_sphinx = '1.0'
 
+# Manually mock the modules to connect to the REANA server API.
+# Note that autodoc_mock_imports does not work.
+MOCK_MODULES = ["reana_commons.api_client"]
+sys.modules.update((mod_name, Mock()) for mod_name in MOCK_MODULES)
```
mdonadoni, line 42: Did you also try using `:exclude-members:` in `index.rst` to exclude the problematic members?

mdonadoni (commented):
Should we also add the return value to the docstrings of every function?

docs/conf.py
giuseppe-steduto: Yes - but unfortunately it does not work as expected, because it still tries to import `current_rs_api_client`. The alternative would be to manually declare all the functions that we want to include in the docs, but I personally prefer mocking the failing import. Do you agree?

giuseppe-steduto:
> Should we also add the return value to the docstrings of every function?

Done! However, it's not super detailed, as I thought that would in a way make the documentation too overloaded (kinda like an OpenAPI specification, especially since many functions return an endpoint response), so I wrote an "overview" of the return value in a more relaxed way. Do you think it's better?

reana_client/api/client.py
```diff
@@ -166,7 +202,16 @@ def get_workflow_status(workflow, access_token):
...
-    """Create a workflow."""
+    """Create a workflow.
+
+    :param reana_specification: a dictionary representing the REANA specification of the workflow.
+    :param name: name of the workflow.
+    :param access_token: access token of the current user.
+
+    :return: if the workflow was created successfully, a dictionary with the information about
+             the ``workflow_id`` and ``workflow_name``, along with a ``message`` of success.
+             Otherwise, only a ``message`` key is returned, with the error message.
```
mdonadoni, line 213: I think the error message is not returned but an exception is raised, am I wrong? If so, let's just remove this phrase.

reana_client/api/client.py
```diff
@@ -205,17 +250,21 @@ def create_workflow_from_json(
...
+    :param workflow_json: workflow specification in JSON format.
     :param workflow_file: workflow specification file path.
                           Ignores ``workflow_json`` if provided.
     :param parameters: workflow input parameters dictionary.
     :param workflow_engine: one of the workflow engines (yadage, serial, cwl)
     :param outputs: dictionary with expected workflow outputs.
 
+    :return: if the workflow was created successfully, a dictionary with the information about
+             the ``workflow_id`` and ``workflow_name``, along with a ``message`` of success.
+             Otherwise, only a ``message`` key is returned, with the error message.
```
mdonadoni, line 266: Same as before

reana_client/api/client.py
```diff
@@ -669,7 +750,15 @@ def get_workflow_specification(workflow, access_token):
...
-    Please note that the workspace will be always deleted, even if ``workspace`` is set to ``False``.
+    :param workflow: name or id of the workflow.
+    :param all_runs: whether to delete all the runs of the workflow.
+    :param workspace: whether to delete the workspace of the workflow.
+    :param access_token: access token of the current user.
+
+    :return: a dictionary that cointains info about the deleted workflow (``workflow_id``, ``workflow_name``,
+             ``status``, ``user``), and a ``message`` key.
+
+    Please note that the workspace will always be deleted, even if ``workspace`` is set to ``False``.
```
mdonadoni, line 761: What do you think about putting this (_Please note ..._)  just after the first description line, and before the parameters' list?

reana_client/api/client.py
```diff
@@ -319,10 +373,12 @@ def start_workflow(workflow, access_token, parameters):
 def upload_file(workflow, file_, file_name, access_token):
     """Upload file to workflow workspace.
 
-    :param workflow: name or id which identifies the workflow.
+    :param workflow: name or id of the workflow.
     :param file_: content of a file that will be uploaded.
     :param file_name: name of a file that will be uploaded.
     :param access_token: access token of the current user.
+
+    :return: a dictionary with a ``message`` about the success or failure of the operation.
```
mdonadoni, line 381: Same as above, I think the message returned is only in the case of a success?

mdonadoni (approved):
Looks good, just small changes left!

reana_client/api/client.py
giuseppe-steduto: At first I thought that it was better to put this note after explaining the parameters, but after a quick look online, it seems common practice to put the whole textual description, even referring to parameters, before the actual list, so I changed it!

mdonadoni (approved):
LGTM!

PR #674, opened 2023-10-02
cli: allow sorting using uppercase column names

### LINKED ISSUE #673: cli: sorting by run number behaves correctly only with lowercase column name
In #619, we solved the `reana-client list` issue of incorrectly sorting the workflow list by run number, because the run number was considered as a string, and thus `2` came after `178`.
However, the fix only works when the column name is specified in lowercase: IOW, `reana-client list --sort run_number` works, but `reana-client list --sort RUN_NUMBER` does not. The other sorting keys appear to work fine in both uppercase and lowercase.

This is because in the case of the run number, the sorting key is overridden only when the column name passed by the user as an argument is lowercase:
https://github.com/reanahub/reana-client/blob/9c94abed31f522a1ff01bd64abd7975b4c675980/reana_client/cli/workflow.py#L273-L287

We can solve this quickly by:
- Changing the `get_sort_key` function so that it takes the column name from the `column_id` argument
- Changing the `get_sort_key` function so that it converts the column name to lowercase before comparing it to `run_number`

giuseppe-steduto:
Closes #673.

### DIFF
--- a/reana_client/cli/workflow.py
+++ b/reana_client/cli/workflow.py
@@ -109,7 +109,7 @@ def workflow_execution_group(ctx):
 @human_readable_or_raw_option
 @click.option(
     "--sort",
-    "sort_columm_name",
+    "sort_column_name",
     default="CREATED",
     help="Sort the output by specified column",
 )
@@ -164,7 +164,7 @@ def workflows_list(  # noqa: C901
     show_all,
     verbose,
     human_readable_or_raw,
-    sort_columm_name,
+    sort_column_name,
     page,
     size,
     filters,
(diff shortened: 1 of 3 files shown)


reana_client/cli/workflow.py
```diff
@@ -271,15 +271,16 @@ def workflows_list(  # noqa: C901
 
         # Sort by given column, making sure that `None` is at the bottom of the list.
         def get_sort_key(x, column_id):
```
mdonadoni, line 273: Nice catch! It took me a while to understand what was going wrong.

Maybe to avoid this issues in the future we could avoid using `sort_column_name` inside of `get_sort_key` , and instead use `headers[column_id]`? In this way the "single source of truth" is the index, and the column name is only used to find the correct position in the array.

Also, to improve readability, we could maybe move the definition of `get_sort_key` to be just before the call to `sorted`?

mdonadoni (approved):
Works perfectly! Could you also add a line in CHANGES.rst?

PR #675, opened 2023-10-18
api: fix create_workflow_from_json to load workflow specification

### LINKED ISSUE #666: client: `create_workflow_from_json` does not work with snakemake workflows
When using the `workflow_file` parameter of `create_workflow_from_json`, snakemake workflows (and probably also the others) do not run correctly as the Snakemake specification is never loaded:

From reana-demo-helloword:
```python
import os

import yaml
from reana_client.api.client import (
    create_workflow_from_json,
    upload_to_server,
    start_workflow,
)

WORKFLOW_NAME = "api-client-snakemake"
REANA_ACCESS_TOKEN = os.environ["REANA_ACCESS_TOKEN"]
REANA_YAML = "reana-snakemake.yaml"

with open(REANA_YAML) as f:
    spec = yaml.safe_load(f)

create_workflow_from_json(
    workflow_file=spec["workflow"]["file"],
    name=WORKFLOW_NAME,
    access_token=REANA_ACCESS_TOKEN,
    parameters=spec["inputs"],
    workflow_engine=spec["workflow"]["type"],
    outputs=spec["outputs"],
)

abs_path_to_input_files = [os.path.abspath(f) for f in spec["inputs"]["files"]]
upload_to_server(WORKFLOW_NAME, abs_path_to_input_files, REANA_ACCESS_TOKEN)

start_workflow(WORKFLOW_NAME, REANA_ACCESS_TOKEN, {})

```

giuseppe-steduto:
Amend `create_workflow_from_json` so that the workflow specification is
always loaded in the REANA specification that is passed to reana-server.

Closes #666.

### DIFF
--- a/reana_client/api/client.py
+++ b/reana_client/api/client.py
@@ -18,6 +18,10 @@ from urllib.parse import urljoin
 import requests
 from bravado.exception import HTTPError
 from reana_commons.validation.utils import validate_reana_yaml, validate_workflow_name
+from reana_commons.specification import (
+    load_workflow_spec_from_reana_yaml,
+    load_input_parameters,
+)
 from reana_commons.api_client import get_current_api_client
 from reana_commons.config import REANA_WORKFLOW_ENGINES
 from reana_commons.errors import REANASecretAlreadyExists, REANASecretDoesNotExist
@@ -248,6 +252,7 @@ def create_workflow_from_json(
     parameters=None,
     workflow_engine="yadage",
     outputs=None,
+    workspace_path=None,
 ):
     """Create a workflow from JSON specification.
 
@@ -259,6 +264,7 @@ def create_workflow_from_json(
     :param parameters: workflow input parameters dictionary.
     :param workflow_engine: one of the workflow engines (yadage, serial, cwl)
     :param outputs: dictionary with expected workflow outputs.
+    :param workspace_path: path to the workspace where the workflow is located.
 
     :return: if the workflow was created successfully, a dictionary with the information about
              the ``workflow_id`` and ``workflow_name``, along with a ``message`` of success.
@@ -290,21 +296,29 @@ def create_workflow_from_json(
         )
     try:
         reana_yaml = dict(workflow={})
-        if workflow_file:
-            reana_yaml["workflow"]["file"] = workflow_file
-        else:
-            reana_yaml["workflow"]["specification"] = workflow_json
         reana_yaml["workflow"]["type"] = workflow_engine
         if parameters:
             reana_yaml["inputs"] = parameters
         if outputs:
             reana_yaml["outputs"] = outputs
+        if workflow_file:
+            reana_yaml["workflow"]["file"] = workflow_file
+            reana_yaml["workflow"][
+                "specification"
+            ] = load_workflow_spec_from_reana_yaml(reana_yaml, workspace_path)
+        else:
+            reana_yaml["workflow"]["specification"] = workflow_json
+        # The function below loads the input parameters into the reana_yaml dictionary
+        # taking them from the parameters yaml files (used by CWL and Snakemake workflows),
+        # and replacing the `input.parameters.input` field with the actual parameters values.
+        # For this reason, we have to load the workflow specification first, as otherwise
+        # the specification validation would fail.
+        input_params = load_input_parameters(reana_yaml, workspace_path)
+        if input_params is not None:
+            reana_yaml["inputs"]["parameters"] = input_params
         validate_reana_yaml(reana_yaml)
-        reana_specification = reana_yaml
         (response, http_response) = current_rs_api_client.api.create_workflow(
-            reana_specification=json.loads(
-                json.dumps(reana_specification, sort_keys=True)
-            ),
+            reana_specification=json.loads(json.dumps(reana_yaml, sort_keys=True)),
             workflow_name=name,
             access_token=access_token,
         ).result()
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,7 @@ Changes
 Version 0.9.2 (UNRELEASED)
 --------------------------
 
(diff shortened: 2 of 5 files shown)


reana_client/api/client.py
```diff
@@ -290,21 +294,24 @@ def create_workflow_from_json(
...
         if outputs:
             reana_yaml["outputs"] = outputs
+        if workflow_file:
+            reana_yaml["workflow"]["file"] = workflow_file
+            reana_yaml["workflow"][
+                "specification"
+            ] = load_workflow_spec_from_reana_yaml(reana_yaml)
+            input_params = load_input_parameters(reana_yaml)
+            if input_params:
+                reana_yaml["inputs"]["parameters"] = input_params
```
mdonadoni, line 309: I guess this part should go inside the `if parameters:` section?

mdonadoni (commented):
Works well!

reana_client/api/client.py
```diff
@@ -290,21 +294,24 @@ def create_workflow_from_json(
...
+        else:
+            reana_yaml["workflow"]["specification"] = workflow_json
         validate_reana_yaml(reana_yaml)
-        reana_specification = reana_yaml
         (response, http_response) = current_rs_api_client.api.create_workflow(
-            reana_specification=json.loads(
-                json.dumps(reana_specification, sort_keys=True)
-            ),
+            reana_specification=json.loads(json.dumps(reana_yaml, sort_keys=True)),
```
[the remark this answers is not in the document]
giuseppe-steduto: That's a good question. Passing directly `reana_yaml` appears to be working, but I am a bit hesitant to remove the whole `json.loads(json.dumps(...))` thing. 

A not-so-likely guess in my opinion, is that it is needed to sort the keys of the dictionary (note the `sort_keys=True`): starting from Python 3.7, indeed, [the dictionary order is guaranteed to be insertion order](https://docs.python.org/3.7/library/stdtypes.html#dict). However, I don't see why it would be needed, if not only for debugging purposes.

My most educated guess would instead be that it is needed to ensure that the `reana_specification` parameter is JSON-serializable (especially because this is a r-server API call), or anyway in a JSON-compatible format (e.g. by replacing tuples with lists). Since the dictionary is built using user-provided arguments, `reana_yaml` could in theory contain non-serializable objects: in that case, `json.dumps()` will raise an exception.

reana_client/api/client.py
giuseppe-steduto: TL;DR I think that putting this part in the `if parameters:` section would make the validation of the Snakemake specification fail, because the parameters would be loaded too soon and replace the config files needed to validate the snakefile.

`reana_yaml` before `load_input_parameters` (pointing to the yaml file with the configuration parameters):
```yaml
inputs:
   parameters:
       input: workflow/Snakemake/helloworld-parameters.yaml
```

And `reana_yaml` after `load_input_parameters` (indeed loading the parameters contained in the yaml file):
```yaml
inputs:
   parameters:
      helloworld: code/helloworld.py
      inputfile: data/names.txt
      sleeptime: 0
```

In the Snakefile, the parameters are taken directly from the configfile:
```python
rule helloworld:
    input:
        helloworld=config["helloworld"],
        inputfile=config["inputfile"],
```

`load_workflow_spec_from_reana_yaml` loads the Snakemake / CWL / yadage specification and calls the respective validation functions. When `r-commons` [calls the snakemake validation function](https://github.com/reanahub/reana-commons/blob/3729df242049fd73c067aa2dcac3f16c242bbd09/reana_commons/snakemake.py#L172), it expects the configfile to be [provided in the kwargs](https://github.com/reanahub/reana-commons/blob/3729df242049fd73c067aa2dcac3f16c242bbd09/reana_commons/snakemake.py#L170). By the way the kwargs are prepared, the configfile is [taken from the `inputs.parameters.input` field](https://github.com/reanahub/reana-commons/pull/413/files#diff-4f3e2c0b81229cc70894a2058f2af7202cb2d2870624680715a09c3ea7c69e05R82-R85), thus this field should not be replaced by the actual parameters yet: that's why I put `load_input_parameters` after `load_workflow_spec_from_reana_yaml`.

It's not a very straightforward process, and the different types of "inputs" don't make things easier, but I hope this was not too hard to follow. What do you think?

reana_client/api/client.py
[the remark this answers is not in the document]
mdonadoni: Let's keep it like this, I could not find any explanation of why it's done like this in older commits.

reana_client/api/client.py
```diff
@@ -290,21 +296,29 @@ def create_workflow_from_json(
...
+            ] = load_workflow_spec_from_reana_yaml(reana_yaml, workspace_path)
+        else:
+            reana_yaml["workflow"]["specification"] = workflow_json
+        # The function below loads the input parameters into the reana_yaml dictionary
+        # taking them from the parameters yaml files (used by CWL and Snakemake workflows),
+        # and replacing the `input.parameters.input` field with the actual parameters values.
+        # For this reason, we have to load the workflow specification first, as otherwise
+        # the specification validation would fail.
+        input_params = load_input_parameters(reana_yaml, workspace_path)
+        if input_params:
```
mdonadoni, line 317:
```suggestion
        if input_params is not None:
```

mdonadoni (commented):
Looking good, just one last small change!

Could you also already update the version pin for r-commons in `setup.py` to `>=0.9.4a1`, so that the CI will pass as soon we release the new version?

mdonadoni:
Snakemake tests are not passing on Python>=3.11 and need to be skipped, as done here: https://github.com/reanahub/reana-commons/blob/2efdd4f9c0d05cb0f7ba292d9a1eb9b70eb13692/tests/test_snakemake.py#L20-L23

Could you please update them so that the CI passes?

mdonadoni (approved):
LGTM! Nice job!

PR #682, opened 2023-11-01
cli: add include-command flag to list and status commands

giuseppe-steduto:
Add `--include-command` flag to the `list` and `status` commands that,
when set, will display info about the command currently being executed
by the workflow (or the last command). In case there is no info about
the command, the step name will be displayed, if possible.

Closes reanahub/reana-workflow-controller#486.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -45,7 +45,7 @@ install_requires = [
     "click>=7",
     "pathspec==0.9.0",
     "jsonpointer>=2.0",
-    "reana-commons[yadage,snakemake,cwl]>=0.9.4a1,<0.10.0",
+    "reana-commons[yadage,snakemake,cwl]>=0.9.4a3,<0.10.0",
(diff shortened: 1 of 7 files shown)


mdonadoni (approved):

setup.py
```diff
@@ -45,7 +45,7 @@
     "click>=7",
     "pathspec==0.9.0",
     "jsonpointer>=2.0",
-    "reana-commons[yadage,snakemake,cwl]>=0.9.4a1,<0.10.0",
+    "reana-commons[yadage,snakemake,cwl]>=0.9.4a2,<0.10.0",
```
mdonadoni, line 48:
```suggestion
    "reana-commons[yadage,snakemake,cwl]>=0.9.4a3,<0.10.0",
```

mdonadoni:
No need to change API after performance improvements of https://github.com/reanahub/reana-db/pull/213

PR #683, opened 2023-11-04
cli: implement share-add command

### LINKED ISSUE #680: cli: implement new share-add command
For the workflow sharing sprint, design and implement an initial `share` command allowing to share workflow runs with colleagues.

For example, Alice shares her myanalysis run 42 with Bob:

```console
$ reana-client share -w myanalysis.42 --share-with bob@example.org
myanalysis.42 is now read-only shared with bob@example.org
```

For example, Alice enquires with whom she shared her myanalysis run 42:

```console
$ reana-client share -w myanalysis.42
myanalysis.42 has been shared with:
- bob@example.org
- cecile@example.org
```

The CLI could have JSON output for easy piping into user scripts (as the `list`, `status` etc commands):

```console
$ reana-cilent share -w myanalysis.42 --json | jq
[ 
    "bob@example.org",
    "cecile@example.org"
]
```

P.S. This issue depends on https://github.com/reanahub/reana-db/issues/206

DaanRosendal:
Adds a new command to the CLI to share a workflow with a user.

Closes #680

### DIFF
--- a/reana_client/cli/workflow.py
+++ b/reana_client/cli/workflow.py
@@ -15,12 +15,8 @@ import time
 import traceback
 
 import click
-from jsonschema.exceptions import ValidationError
-from reana_commons.config import INTERACTIVE_SESSION_TYPES, REANA_COMPUTE_BACKENDS
-from reana_commons.errors import REANAValidationError
-from reana_commons.validation.operational_options import validate_operational_options
 import yaml
-
+from jsonschema.exceptions import ValidationError
 from reana_client.cli.files import get_files, upload_files
 from reana_client.cli.utils import (
     add_access_token_options,
@@ -51,6 +47,9 @@ from reana_client.validation.utils import (
     validate_input_parameters,
     validate_workflow_name_parameter,
 )
+from reana_commons.config import INTERACTIVE_SESSION_TYPES, REANA_COMPUTE_BACKENDS
+from reana_commons.errors import REANAValidationError
+from reana_commons.validation.operational_options import validate_operational_options
 
 
 @click.group(help="Workflow management commands")
@@ -64,6 +63,13 @@ def workflow_management_group(ctx):
 @click.pass_context
 def workflow_execution_group(ctx):
     """Top level wrapper for execution related interaction."""
+    logging.debug(ctx.info_name)
+
+
+@click.group(help="Workflow sharing commands")
+@click.pass_context
+def workflow_sharing_group(ctx):
+    """Top level wrapper for workflow sharing."""
     logging.debug(ctx.info_name)
 
 
@@ -451,12 +457,12 @@ def workflow_start(
     \t $ reana-client start -w myanalysis.42 -p sleeptime=10 -p myparam=4\n
     \t $ reana-client start -w myanalysis.42 -p myparam1=myvalue1 -o CACHE=off
     """
-    from reana_client.utils import get_api_url
     from reana_client.api.client import (
         get_workflow_parameters,
         get_workflow_status,
         start_workflow,
     )
+    from reana_client.utils import get_api_url
 
     def display_status(workflow: str, current_status: str):
         """Display the current status of the workflow."""
@@ -586,12 +592,12 @@ def workflow_restart(
     \t $ reana-client restart -w myanalysis.42 -o TARGET=gendata\n
     \t $ reana-client restart -w myanalysis.42 -o FROM=fitdata
     """
-    from reana_client.utils import get_api_url
     from reana_client.api.client import (
         get_workflow_parameters,
         get_workflow_status,
         start_workflow,
     )
+    from reana_client.utils import get_api_url
 
     logging.debug("command: {}".format(ctx.command_path.replace(" ", ".")))
     for p in ctx.params:
@@ -1373,7 +1379,7 @@ def workflow_open_interactive_session(
     Examples:\n
     \t $ reana-client open -w myanalysis.42 jupyter
     """
-    from reana_client.api.client import open_interactive_session, info
+    from reana_client.api.client import info, open_interactive_session
 
     if workflow:
         try:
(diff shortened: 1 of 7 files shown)


reana_client/cli/workflow.py
```diff
@@ -1458,3 +1464,84 @@ def workflow_close_interactive_session(workflow, access_token):  # noqa: D301
...
+    $ reana-client share-add -w myanalysis.42 --user bob@example.org
+
+    $ reana-client share-add -w myanalysis.42 --user bob@example.org --user cecile@example.org --message "Please review my analysis" --valid-until 2025-12-31
+    """
+    from reana_client.api.client import share_workflow
+
+    share_errors = []
+    shared_users = []
+
+    if workflow:
```
mdonadoni, line 1511: No need to do `if workflow: ... else: display_message(...)`, as this is already handled by `@add_workflow_option`. I think there are still some commands that do this, but it's a leftover from the past (that we should probably clean up at some point :sweat_smile: )

reana_client/cli/workflow.py
```diff
@@ -1458,3 +1464,84 @@ def workflow_close_interactive_session(workflow, access_token):  # noqa: D301
...
+                        access_token,
+                        message=message,
+                        valid_until=valid_until,
+                    )
+                    shared_users.append(user)
+                except Exception as e:
+                    share_errors.append(
+                        f"Failed to share {workflow} with {user}: {str(e)}"
+                    )
+                    logging.debug(traceback.format_exc())
```
mdonadoni, line 1528: nitpick: You can use `logging.exception("... message ...")` instead of calling `logging.debug` manually

reana_client/cli/workflow.py
```diff
@@ -1458,3 +1464,84 @@ def workflow_close_interactive_session(workflow, access_token):  # noqa: D301
...
+        except Exception as e:
+            logging.debug(traceback.format_exc())
+            logging.debug(str(e))
+            display_message(
+                "An error occurred while sharing workflow:\n{}".format(str(e)),
+                msg_type="error",
+            )
```
mdonadoni, line 1535: In which cases can we end up here? All the exceptions are already caught by the other `try ... except` block, aren't they?

reana_client/cli/workflow.py
```diff
@@ -1458,3 +1464,84 @@ def workflow_close_interactive_session(workflow, access_token):  # noqa: D301
...
+
+        if shared_users:
+            display_message(
+                f"{workflow} is now read-only shared with {', '.join(shared_users)}",
+                msg_type="success",
+            )
+        if share_errors:
```
mdonadoni, line 1542: No need to check `if share_errors: ...`, as if the array is empty then the for loop will not do anything

reana_client/cli/workflow.py
```diff
@@ -1458,3 +1464,84 @@ def workflow_close_interactive_session(workflow, access_token):  # noqa: D301
...
+            for error in share_errors:
+                display_message(error, msg_type="error")
```
mdonadoni, line 1544: If there are errors (here or before), then we reana-client should exit with non-zero exit code

reana_client/cli/workflow.py
```diff
@@ -1458,3 +1464,84 @@ def workflow_close_interactive_session(workflow, access_token):  # noqa: D301
...
+    help="Users to share the workflow with.",
+    required=True,
+)
+@click.option(
+    "-m",
+    "--message",
+    help="Optional message that is sent to the user(s) with the sharing invitation.",
+)
+@click.option(
+    "-v",
```
mdonadoni, line 1487: The shorthand option `-v` is normally used for `--verbose` in other scripts, so I think we should either change it to another letter or even drop the shorthand version. What do you think?

mdonadoni:
Late comment: JSON output is missing, but if you prefer we can tackle this later on

reana_client/cli/workflow.py
```diff
@@ -1458,3 +1464,84 @@ def workflow_close_interactive_session(workflow, access_token):  # noqa: D301
...
+    "--valid-until",
+    help="Optional date when access to the workflow will expire for the given user(s) (format: YYYY-MM-DD).",
+)
+@click.pass_context
+def share_workflow_add(
```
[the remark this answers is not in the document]
DaanRosendal: Done

The changes were applied to the following superseding PR: #692

reana_client/cli/workflow.py
```diff
@@ -1458,3 +1464,84 @@ def workflow_close_interactive_session(workflow, access_token):  # noqa: D301
...
+@click.option(
+    "-m",
+    "--message",
+    help="Optional message that is sent to the user(s) with the sharing invitation.",
+)
+@click.option(
+    "-v",
+    "--valid-until",
+    help="Optional date when access to the workflow will expire for the given user(s) (format: YYYY-MM-DD).",
+)
```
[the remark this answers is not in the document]
DaanRosendal: Thanks for the suggestion. Added.

The changes were applied to the following superseding PR: #692

reana_client/cli/workflow.py
DaanRosendal: Nicely spotted. Changed

The changes were applied to the following superseding PR: #692

reana_client/cli/workflow.py
DaanRosendal: Done 

The changes were applied to the following superseding PR: #692

reana_client/cli/workflow.py
DaanRosendal: Nicely spotted. Changed

The changes were applied to the following superseding PR: #692

reana_client/cli/workflow.py
DaanRosendal: This would cancel out any subsequent errors. E.g. imagine someone shares with 20 people at a time where some emails have typos; the current implementation would then show all errors after 1 command execution, while the implementation you suggest requires the user to run the command again after each error to find the next error. 

We can still change this if you prefer the latter scenario for a reason that I'm not seeing.

reana_client/cli/workflow.py
DaanRosendal: Changed it to allow for `--valid-until` only, removed the `-v` shorthand.

The changes were applied to the following superseding PR: #692

DaanRosendal:
@mdonadoni I can not reply directly to your last comment so I'll do it like this.

I created an issue for this so we don't forget about it: https://github.com/reanahub/reana-client/issues/712

reana_client/cli/workflow.py
mdonadoni: Yeah, it's good that the errors are all shown at the end! What I meant was to add something like
```python
if share_errors:
    sys.exit(1)
```
at the end of the command, so that in case of errors reana-client returns an appropriate status code. No need to call sys.exit it when the error happens.

DaanRosendal: Agreed :+1:

mdonadoni:
Included in https://github.com/reanahub/reana-client/pull/692

PR #701, opened 2024-01-29
fix(status): correctly display duration of stopped workflows (#701)

### LINKED ISSUE #699: cli: duration of stopped workflows is not displayed correctly
Similarly to what was happening in reanahub/reana-ui#386, when running `reana-client status --include-duration` for a stopped workflow, the displayed duration is not the time between the moment in which the run was started and the moment in which the run was stopped, but the total time passed from the beginning of the run until the current moment.

```console
❯ reana-client status -w recast --include-duration --json | jq '.[0].status'
"stopped"
❯ reana-client status -w recast --include-duration --json | jq '.[0].duration'
2598
# After a few seconds...
❯ reana-client status -w recast --include-duration --json | jq '.[0].duration'
2604
```

This is very likely because in [`reana_client.utils.get_workflow_duration`](https://github.com/reanahub/reana-client/blob/5b278f131b59d3ecfd3c7f129040a126cd01b60a/reana_client/utils.py#L134) we only look at `run_finished_at`, but not at `run_stopped_at`.

giuseppe-steduto:
Closes #699

### DIFF
--- a/reana_client/utils.py
+++ b/reana_client/utils.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of REANA.
-# Copyright (C) 2017, 2018, 2019, 2020, 2021, 2022 CERN.
+# Copyright (C) 2017, 2018, 2019, 2020, 2021, 2022, 2023, 2024 CERN.
 #
 # REANA is free software; you can redistribute it and/or modify it
 # under the terms of the MIT License; see LICENSE file for more details.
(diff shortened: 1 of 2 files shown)


giuseppe-steduto:
Note for the reviewer: I chose `status` as the scope of the commit because I think that the commands affected by the PR are a nice user-facing scope for changes to the client. For example, this PR fixes `status`, `list`, and in general the commands that use the `--include-duration` flag, I chose `status` as it is the one most evidently impacted. WDYT?

mdonadoni (approved):
LGTM! Needs a rebase :wink:

mdonadoni:
Regarding the scope, I think `fix(status)` can be a good choice. WDYT @tiborsimko ?

PR #704, opened 2024-02-12
build(appimage): upgrade to Python 3.8.18

tiborsimko:
Fixes AppImage building procedure by upgrading to Python 3.8.18.

### DIFF
--- a/scripts/create-appimage.sh
+++ b/scripts/create-appimage.sh
@@ -1,10 +1,13 @@
 #!/bin/sh
 #
 # This file is part of REANA.
-# Copyright (C) 2022 CERN.
+# Copyright (C) 2022, 2024 CERN.
 #
 # REANA is free software; you can redistribute it and/or modify it
 # under the terms of the MIT License; see LICENSE file for more details.
+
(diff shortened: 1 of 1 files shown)


giuseppe-steduto (approved):
LGTM!
A small note: I had to run the script as superuser in order to have the permission to complete the AppImage build process successfully. Is this intended behavior?

tiborsimko:
>  had to run the script as superuser in order to have the permission to complete the AppImage build process successfully. Is this intended behavior?

Hmm, I'm running it as a regular use. Perhaps we can see IRL about the troubles you've met?

tiborsimko:
> Hmm, I'm running it as a regular use. Perhaps we can see IRL about the troubles you've met?

I managed to reproduce a runtime error: it happens when I launch this script whilst being in the `reana` virtual environment. However, when you `deactivate` first, and use the vanilla operating system Python, then the script runs fine. Perhaps you tried it out whilst being in the `reana` venv?

mdonadoni (approved):
Works well after the latest changes (outside the `reana` venv)

PR #721, opened 2024-07-12
build(docker): upgrade to ubuntu 24.04 and Python 3.12

mdonadoni:
Closes reanahub/reana#808


Dockerfile
mdonadoni, line 1: TODO: update also AppImage script

tiborsimko (approved):

PR #724, opened 2024-07-26
feat(test): implement test command

### LINKED ISSUE #723: cli: Implement test command
Add fleshed out implementation of the test command, using the gherkin parser module in reana-commons.

ajclyall:
Closes #723

### DIFF
--- a/reana_client/cli/test.py
+++ b/reana_client/cli/test.py
@@ -1,0 +1,169 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of REANA.
+# Copyright (C) 2024 CERN.
+#
+# REANA is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+"""REANA client test commands."""
+
+import sys
+import click
+import logging
+import traceback
+import time
+from reana_client.cli.utils import add_access_token_options, check_connection
+from reana_commons.gherkin_parser.parser import (
+    parse_and_run_tests,
+    AnalysisTestStatus,
+)
+from reana_commons.gherkin_parser.data_fetcher import DataFetcherBase
+from reana_commons.gherkin_parser.errors import FeatureFileError
+from reana_client.printer import display_message
+from reana_client.api.client import (
+    list_files,
+    get_workflow_disk_usage,
+    get_workflow_logs,
+    get_workflow_status,
+    get_workflow_specification,
+    download_file,
+)
+from reana_client.cli.utils import add_workflow_option
+
+
+class DataFetcherClient(DataFetcherBase):
+    """Implementation of the DataFetcherBase using reana_client.api.client methods."""
+
+    def __init__(self, access_token):
+        """Initialize DataFetcherClient with access token."""
+        self.access_token = access_token
+
+    def list_files(self, workflow, file_name=None, page=None, size=None, search=None):
+        """Return the list of files for a given workflow workspace."""
+        return list_files(workflow, self.access_token, file_name, page, size, search)
+
+    def get_workflow_disk_usage(self, workflow, parameters):
+        """Display disk usage workflow."""
+        return get_workflow_disk_usage(workflow, parameters, self.access_token)
+
+    def get_workflow_logs(self, workflow, steps=None, page=None, size=None):
+        """Get logs from a workflow engine, use existing API function."""
+        return get_workflow_logs(workflow, self.access_token, steps, page, size)
+
+    def get_workflow_status(self, workflow):
+        """Get of a previously created workflow."""
+        return get_workflow_status(workflow, self.access_token)
+
+    def get_workflow_specification(self, workflow):
+        """Get specification of previously created workflow."""
+        return get_workflow_specification(workflow, self.access_token)
+
+    def download_file(self, workflow, file_path):
+        """Download the requested file if it exists."""
+        return download_file(workflow, file_path, self.access_token)
+
+
+@click.group(help="Workflow run test commands")
+def test_group():
+    """Workflow run test commands."""
+
+
+@test_group.command("test")
+@click.option(
+    "-n",
+    "--test-files",
+    multiple=True,
+    default=None,
+    help="Gherkin file for testing properties of a workflow execution. Overrides files in reana.yaml if provided.",
+)
+@click.pass_context
+@add_access_token_options
+@check_connection
+@add_workflow_option
+def test(ctx, workflow, test_files, access_token):
+    r"""
+    Test workflow execution, based on a given Gherkin file.
+
+    Gherkin files can be specified in the reana specification file (reana.yaml),
+    or by using the ``-n`` option.
+
+    The ``test`` command allows for testing of a workflow execution,
+    by assessing whether it meets certain properties specified in a
+    chosen gherkin file.
+
+    Example:
+        $ reana-client test -w myanalysis -n test_analysis.feature
+        $ reana-client test -w myanalysis
+        $ reana-client test -w myanalysis -n test1.feature -n test2.feature
+    """
+    start_time = time.time()
+    try:
+        workflow_status = get_workflow_status(
+            workflow=workflow, access_token=access_token
+        )
+        status = workflow_status["status"]
+        workflow_name = workflow_status["name"]
+    except Exception as e:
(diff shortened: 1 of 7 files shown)


ajclyall:
The tests are failing because gherkin-parser doesn't exist in reana-commons yet.

reana_client/cli/test.py
```diff
@@ -0,0 +1,174 @@
...
+    "--workflow",
+    default="workflow",
+    help="Name of workflow to be tested, default is 'workflow'",
+)
+@click.option(
+    "-n",
+    "--test_file",
+    default=None,
+    help="Gherkin file for testing properties of a workflow execution. Overrides files in reana.yaml if provided.",
+)
```
mdonadoni, line 74: Maybe we can accept multiple test files? But let's discuss it together.

Also regarding cosmetics, let's use a dash instead of an underscore in the option name (`--test-file`).

reana_client/cli/test.py
```diff
@@ -0,0 +1,174 @@
...
+    """Workflow run test commands."""
+
+
+@test_group.command("test")
+@click.option(
+    "-w",
+    "--workflow",
+    default="workflow",
+    help="Name of workflow to be tested, default is 'workflow'",
+)
```
mdonadoni, line 68: You can use the decorator `@add_workflow_option` from `utils.py`, so that also the `REANA_WORKON` environment variable is considered.

For example: https://github.com/reanahub/reana-client/blob/822a5387acacb40d53ebe951b7f2332e26f0c5dc/reana_client/cli/workflow.py#L407

reana_client/cli/test.py
```diff
@@ -0,0 +1,174 @@
...
+    try:
+        reana_specification = load_validate_reana_spec(
+            specification_filename,
+            access_token=access_token,
+            skip_validation=skip_validation,
+            server_capabilities=True,
+        )
+    except Exception:
+        display_message(f"Error loading {file} specification file.", msg_type="error")
+        sys.exit(1)
```
mdonadoni, line 117: I'm wondering whether we should fetch the specification of the workflow passed with `--workflow` directly from the server, instead of reading it from disk?

mdonadoni: In this case, we would also not need the `--skip-validation` flag.

reana_client/cli/test.py
```diff
@@ -0,0 +1,144 @@
...
+                "No test files specified in reana.yaml and no -n option provided.",
+                msg_type="error",
+            )
+            sys.exit(1)
+
+    data_fetcher = DataFetcherClient(access_token)
+    for test_file in test_files:
+        click.secho(f"\nUsing test file {test_file}", fg="cyan", bold=True)
+        try:
+            results = parse_and_run_tests(id, test_file, workflow, data_fetcher)
```
mdonadoni, line 128: The workflow ID is being passed as the analysis run ID, but analysis runs are  things specific to continuous-reuse (see also the related comment in reana-commons).

Currently, what the user has passed as `workflow` is what is being used as an identifier to gather details from REANA API. This can create some problems if the user specifies the workflow name without the run number (e.g. `my-roofit` instead of `my-roofit.2.1`), as different calls to the REANA API might return details of different workflows (e.g. if the user has spawned another `my-roofit` workflow in the meantime).  Should we pass either the full name (that we can get from the response of `get_workflow_status`) or the ID of the workflow as the `workflow` parameter of `parse_and_run_tests`? 

(we can discuss this live given that `workflow` is quite ambiguous in this context)

mdonadoni (commented):
Looks good! I have left just a few small comments.

This PR also needs to be rebased ;)

reana_client/cli/test.py
```diff
@@ -0,0 +1,144 @@
...
+            )
+            sys.exit(1)
+
+        click.secho(f"Summary of {test_file}:", bold=True)
+        for scenario in results[1]:
+            click.echo(f"Tested {scenario.scenario}: ", nl=False)
+            click.secho(
+                f"{scenario.result.name}",
+                fg=("green" if scenario.result == AnalysisTestStatus.passed else "red"),
+            )
```
mdonadoni, line 144: One last thing:  we should exit with non-zero status code (e.g. `sys.exit(1)`) if any of the tests failed

mdonadoni (approved):
Looks good to me! Nice job! :rocket:

mdonadoni (approved):
LGTM! CI should pass after installing a new version of reana-commons

tests/test_cli_test.py
```diff
@@ -0,0 +1,391 @@
...
+                "test",
+                "-w",
+                "myanalysis",
+                "-t",
+                "000000",
+                "-n",
+                "test_analysis.feature",
+            ],
+        )
+        assert "Using test file test_analysis.feature" in result.output
```
tiborsimko, line 155: I have tested a few combinations of client and server and test files, and things seem to work nicely :+1:

However I have some cosmetic comments to the output formatting.

Currently, when a test fails, the output looks like:

```
Using test file tests/serial/log-messages.feature
[ERROR] Scenario failed! Failed testcase: the job logs for the "gendata" step should contain "datasets
--------
ZRooDataSet::modelData(x)"
[ERROR] Error log: The logs do not contain datasets
--------
ZRooDataSet::modelData(x)!
Summary of tests/serial/log-messages.feature:
Tested The workflow start has produced the expected messages: passed
Tested The generation step has produced the expected messages: failed
Tested The fitting step has produced the expected messages: passed

Using test file tests/serial/run-duration.feature
Summary of tests/serial/run-duration.feature:
Tested The workflow terminates in a reasonable amount of time: passed
Tested The data generation step terminates in a reasonable amount of time: passed
Tested The fitting step terminates in a reasonable amount of time: passed

Using test file tests/serial/workspace-files.feature
Summary of tests/serial/workspace-files.feature:
Tested The workspace contains the expected input files: passed
Tested The generation step creates the appropriate data file: passed
Tested The workflow generates the correct final plot: passed
Tested The total workspace size remains within reasonable limits: passed
```

Some observations on the above:

- Note that for multi-line values, it is not clear when the expected text starts and ends. In the first ERROR message, the text is well delimited by quotes. In the second ERROR message, it is not. It would be nice to make the second error message behave like the first one, for consistency.

- BTW the second ERROR message basically just repeats what the first ERROR message already reported; do we need it?

- Regarding the verbiage, we could perhaps make some changes e.g. "using test file" may not be clear (so rather "testing..." rather than "using..."), e.g. "Tested The workspace..." and "Scenario failed" may not be clear that they both refer to scenarios (so rather "Scenario..." than "Tested..."), etc.  See proposal below.

- Also, the output formatting could emulate what `reana-client validate` is outputting with respect to headings and errors/failures, so that researchers would "feel at home" with this new command. See proposal below.

What about emitting the following output? (errors either displayed all first before summary, or inline within respective scenarios and users will just grep for '->')

```
[ERROR] Scenario failed! Failed testcase: the job logs for the "gendata" step should contain "datasets
--------
ZRooDataSet::modelData(x)"

==> Testing file "tests/serial/log-messages.feature"
  -> SUCCESS: Scenario "The workflow start has produced the expected messages"
  -> ERROR: Scenario "The generation step has produced the expected messages"
  -> SUCCESS: Scenario "The fitting step has produced the expected messages"

==> Testing file "tests/serial/workspace-files.feature"
  -> SUCCESS: Scenario "The workspace contains the expected input files"
  -> SUCCESS: Scenario "The generation step creates the appropriate data file"
  -> SUCCESS: Scenario "The workflow generates the correct final plot"
  -> SUCCESS: Scenario "The total workspace size remains within reasonable limits"

6 passed, 1 failed in 30s
```

ajclyall: Regarding the errors, I agree that the second error message can be redundant, especially in tests that are about "presence of files" or "presence of log messages". The first is the step that causes the error, and the second is the reasoning, but for these cases these two ideas coincide.

This is a case where I could see some utility, but I'm not sure if it's worth keeping both (maybe only the reasoning is needed?). It tells you which line in your gherkin file was the problem, and then the precise reason.
```
[ERROR] Scenario failed! Failed testcase: the workflow run duration should be less than 1 minutes
[ERROR] Error log: The workflow took more than 1 minutes to complete! Run duration: 4.133333333333334 minutes
```

PR #490, opened 2021-03-26
cli: support dir and glob pattern on download 

mvidalgarcia:
Depends on https://github.com/reanahub/reana-client/pull/483

Needs `r-commons` release to pass - https://github.com/reanahub/reana-commons/pull/253

closes #402

#### To test
- Create a workflow
- Upload multiple files in multiple levels of directories
- Try to download different directories by name
- Try to download multiple files by passing glob patterns

Pseudo-example
```console
$ reana-client download -w workflow 'mydir'
$ reana-client download -w workflow '**/*.py'
$ reana-client download -w workflow '**/*.txt'
$ reana-client download -w workflow 'mydir/foo/*.py'
...
```
If it the pattern matches more than one file, you should get a zip downloaded, otherwise a single file.

#### TBD

When downloading a single file,  until now, the path where this file is in the workspace is respected locally. This is because the file is saved directly using the value that is passed to the command as an argument. E.g. `reana-client download -w workflow foo/bar/myfile.txt` would create directories `foo/bar` for you as it uses this argument to save the file.

After this PR, since the argument can be a glob pattern, we get the filename from the server-side and save it as it is, thus, causing the path where the file is not to be respected. This is done this way because otherwise, one might end up having folders called `\*\*` or similar. E.g. `reana-client download -w workflow '**/myfile.txt'` would create a dir called `\*\*` and put `myfile.txt` inside.

**Update 19/04/2021**: It is possible to pass `attachment_filename` arg to `send_from_directory` so one can customize the content of `filename` in the `Content-Disposition` header. Passing there a relative path allows us to keep the folder structure and be backwards compatible. Updated on https://github.com/reanahub/reana-workflow-controller/pull/369/files

PR #732, opened 2024-10-09
build(python): add support for Python 3.13 (#732)

jlemesh:
Closes https://github.com/reanahub/reana/issues/837

Add support for Python 3.13.

Replaced `cgi.parse_headers` with `email.message.Message` as recommended in https://peps.python.org/pep-0594/#cgi. Added test to make sure it does the same things that the previous code did. Also tested as in https://github.com/reanahub/reana-client/pull/490.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -189,7 +189,7 @@ jobs:
     strategy:
       matrix:
         testenv: [lowest, release]
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12", "3.13"]
     steps:
       - uses: actions/checkout@v4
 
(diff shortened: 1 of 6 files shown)


jlemesh:
3.13 tests fail with:

```
reana_client/api/client.py:10: in <module>
    import cgi
E   ModuleNotFoundError: No module named 'cgi'
```

CGI module is "Deprecated since version 3.11, will be removed in version 3.13": https://docs.python.org/3/library/cgi.html

tiborsimko:
> 3.13 tests fail with:

We use `cgi` in one place only:

```console
$ git grep cgi
reana_client/api/client.py:10: import cgi
reana_client/api/client.py:504:            value, params = cgi.parse_header(content_disposition)
```

Can you look at updating that part of the code?

tiborsimko (approved):
Nice tests :+1:

PR #741, opened 2025-04-23
ci(runners): upgrade CI runners to Ubuntu 22.04

tiborsimko:
Updates CI runners to use Ubuntu 22.04 in the `maint-0.9` branches because 20.04 is not supported anymore. Also removes Python 3.6 from the test matrix because the runners do not support it anymore.


tiborsimko (commented):
Note: ignoring CI red light for the docs, because they are compared for the `master` branch that contains also workflow sharing etc.

PR #748, opened 2025-07-14
fix(snakemake): make image validation available for Snakemake environments

### LINKED ISSUE #747: validate: the validation of Snakemake environments is not working
### Current behaviour

The advanced validation of CWL, Serial, or Yadage environments is working fine, for example:

```console
$ cd reana-demo-root6-roofit
$ reana-client validate --environments --pull -f reana-cwl.yaml
...
==> Verifying environments in REANA specification file...
WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
  -> SUCCESS: Environment image docker.io/reanahub/reana-env-root6:6.18.04 has the correct format.
  -> WARNING: Environment image docker.io/reanahub/reana-env-root6:6.18.04 does not exist locally.
  -> SUCCESS: Environment image docker.io/reanahub/reana-env-root6:6.18.04 exists in Docker Hub.
  -> INFO: Environment image uses UID 0 but will run as UID 1000.
```

The validation of Snakemake workflows is not working though:

```console
$ cd reana-demo-root6-roofit
$ reana-client validate --environments --pull -f reana-snakemake.yaml
==> Verifying environments in REANA specification file...
```

This is because the specs are empty for Snakemake, something that was necessary for some of the past Snakemake 7 to 8 upgrades.

### Expected behaviour

The Snakemake image validation should be also available for Snakemake workflows.

Let's change the logic so that we extract all Snakefile paths and parse the Snakefiles to extract environment information. (I.e. not going to open the can of worms of why the specs is empty for Snakemake workflows.)

### Notes

This issue is a pre-requisite to https://github.com/reanahub/reana-server/issues/728

CameronMcClymont:
Add Snakemake image validation by parsing and extracting images from the Snakefile specified in the REANA yaml.

Closes #747

### DIFF
--- a/reana_client/validation/environments.py
+++ b/reana_client/validation/environments.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of REANA.
-# Copyright (C) 2021, 2022, 2023, 2024, 2025 CERN.
+# Copyright (C) 2021, 2022, 2023, 2024, 2025, 2026 CERN.
 #
 # REANA is free software; you can redistribute it and/or modify it
 # under the terms of the MIT License; see LICENSE file for more details.
@@ -157,7 +157,7 @@ class EnvironmentValidatorBase:
     def _validate_environment_image(self, image, kubernetes_uid=None):
         """Validate image environment.
 
-        :param image: Full image name with tag if specified.
+        :param image: Full image name with tag if specified. E.g. `reanahub/reana-env-jupyter:2.0.0`.
         :param kubernetes_uid: Kubernetes UID defined in workflow spec.
         """
         if image not in self.validated_images:
@@ -523,10 +523,56 @@ class EnvironmentValidatorCWL(EnvironmentValidatorBase):
 class EnvironmentValidatorSnakemake(EnvironmentValidatorBase):
     """REANA Snakemake workflow environments validation."""
 
+    def __init__(self, **kwargs):
+        """Validate environments in REANA Snakemake workflow."""
+        super().__init__(**kwargs)
+        # Ordered set (dict keys) of rules whose container is resolved per job,
+        # reported as a single aggregated warning once validation is done.
+        self._dynamic_container_rules = {}
+
+    @staticmethod
+    def _is_dynamic_image(image):
+        """Return True when a container value is resolved per job by Snakemake.
+
+        A ``container:`` directive may be a callable or a wildcard-parameterised
+        string, in which case Snakemake only resolves it once the wildcards of a
+        concrete job are known (see ``Rule.expand_container_img``).  The rule-level
+        value is then not a usable image name and cannot be validated.
+        """
+        return not isinstance(image, str) or "{" in image
+
+    def _note_dynamic_container(self, rule_name):
+        """Record a rule whose container image is resolved per job.
+
+        A single ``container:`` directive at the top of a Snakefile applies to
+        every rule, so these are collected and reported once by
+        ``_warn_dynamic_containers()`` instead of warning per rule.
+        """
+        self._dynamic_container_rules[rule_name] = None
+
+    def _warn_dynamic_containers(self):
+        """Emit one aggregated warning for all rules with a dynamic container."""
+        rules = list(self._dynamic_container_rules)
+        if not rules:
+            return
+        max_listed = 5
+        listed = ", ".join(f"'{name}'" for name in rules[:max_listed])
+        if len(rules) > max_listed:
+            listed += f" and {len(rules) - max_listed} more"
+        self.messages.append(
+            {
+                "type": "warning",
+                "message": "Skipping image validation for {} {} with a dynamic "
+                "container directive: {}.".format(
+                    len(rules), "rule" if len(rules) == 1 else "rules", listed
+                ),
+            }
+        )
+
     def validate_environment(self):
         """Validate environments in REANA Snakemake workflow."""
         for step in self.workflow_steps:
-            image = step["environment"]
+            image = step.get("environment")
             if not image:
                 self.messages.append(
                     {
@@ -535,5 +581,10 @@ class EnvironmentValidatorSnakemake(EnvironmentValidatorBase):
                     }
                 )
                 continue
+            if self._is_dynamic_image(image):
+                self._note_dynamic_container(step.get("name"))
+                continue
             kubernetes_uid = step.get("kubernetes_uid")
             self._validate_environment_image(image, kubernetes_uid=kubernetes_uid)
+
+        self._warn_dynamic_containers()
--- a/setup.py
+++ b/setup.py
@@ -26,7 +26,7 @@ extras_require = {
         "sphinx-click>=1.0.4",
     ],
     "tests": [
-        "reana-commons[kubernetes,tests]>=0.95.0a21,<0.96.0",
+        "reana-commons[kubernetes,tests]>=0.95.0a22,<0.96.0",
     ],
 }
 
@@ -41,7 +41,7 @@ install_requires = [
     "pathspec>=0.9.0,<1.0 ; python_version<'3.9'",
     "pathspec>=0.9.0 ; python_version>='3.9'",
     "jsonpointer>=2.0",
-    "reana-commons[yadage,snakemake,cwl]>=0.95.0a21,<0.96.0",
+    "reana-commons[yadage,snakemake,cwl]>=0.95.0a22,<0.96.0",
     "tablib>=0.12.1,<0.13",
     "werkzeug>=0.14.1 ; python_version<'3.10'",
     "werkzeug>=0.15.0 ; python_version>='3.10'",
--- a/tests/test_validate_snakemake_environments.py
+++ b/tests/test_validate_snakemake_environments.py
@@ -1,0 +1,202 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of REANA.
+# Copyright (C) 2026 CERN.
+#
+# REANA is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""REANA client Snakemake environment validation tests.
+
+Container images are validated from ``workflow.specification.steps``, which
+reana-commons populates by loading the Snakefile (expanding ``include:``
+directives and applying the workflow configuration).  These tests therefore
+exercise the validator against steps shaped the way reana-commons produces them.
+"""
+
+from unittest.mock import patch
+
+import pytest
+
+from reana_client.validation.environments import (
+    EnvironmentValidatorSnakemake,
+    validate_environment,
+)
+from reana_commons.config import REANA_DEFAULT_SNAKEMAKE_ENV_IMAGE
+
+
+def _make_validator(steps):
+    """Return a validator over REANA-spec ``steps``."""
+    return EnvironmentValidatorSnakemake(workflow_steps=steps)
+
+
+def _run_validation(validator):
+    """Run environment validation, returning the images it validated."""
+    images = []
+    with patch.object(
+        validator,
+        "_validate_environment_image",
+        side_effect=lambda image, **kwargs: images.append(image),
+    ):
+        validator.validate_environment()
+    return images
+
+
+def _dynamic_warnings(validator):
+    """Return the warnings emitted for skipped dynamic ``container:`` directives."""
+    return [
+        m["message"]
+        for m in validator.messages
+        if m["type"] == "warning" and "dynamic container directive" in m["message"]
+    ]
+
+
+# ---------------------------------------------------------------------------
+# Step-level environments
+# ---------------------------------------------------------------------------
+
+
+class TestStepLevelEnvironments:
+    """Container images recorded on the loaded workflow steps."""
+
+    def test_step_without_environment_warns_with_default(self):
+        validator = _make_validator([{"name": "make"}])
+        assert _run_validation(validator) == []
+        assert any(
+            REANA_DEFAULT_SNAKEMAKE_ENV_IMAGE in m["message"]
+            for m in validator.messages
+            if m["type"] == "warning"
+        )
+
+    def test_step_environment_validated_with_uid(self):
+        validator = _make_validator(
+            [{"name": "make", "environment": "python:3.11", "kubernetes_uid": 1000}]
+        )
+        with patch.object(validator, "_validate_environment_image") as mock_image:
+            validator.validate_environment()
+        mock_image.assert_called_once_with("python:3.11", kubernetes_uid=1000)
+
+    def test_every_step_is_validated(self):
+        validator = _make_validator(
+            [
+                {"name": "a", "environment": "python:3.11"},
+                {"name": "b", "environment": "reanahub/reana-env-root6:6.18.04"},
+            ]
+        )
+        assert _run_validation(validator) == [
+            "python:3.11",
+            "reanahub/reana-env-root6:6.18.04",
+        ]
+
+    def test_invalid_step_image_exits_nonzero(self):
+        # Two colons make ``_validate_image_tag`` raise.
+        validator = _make_validator([{"name": "make", "environment": "python:3.11:x"}])
+        with pytest.raises(SystemExit) as exc_info:
+            validator.validate()
+        assert exc_info.value.code == 1
+
+
+# ---------------------------------------------------------------------------
+# Dynamic ``container:`` directives resolved per job by Snakemake
+# ---------------------------------------------------------------------------
+
+
+class TestDynamicContainerImages:
+    """Per-job container values are skipped with a warning, never validated.
+
+    Snakemake resolves a callable or wildcard-parameterised ``container:`` value
+    only once a concrete job's wildcards are known, so the value recorded on the
+    step is not a usable image name.  Validating it would fail a working
+    workflow, so it is deliberately skipped instead.
+    """
+
+    @pytest.mark.parametrize(
+        "image",
+        [
+            "docker://python:{version}",  # wildcard-parameterised
+            lambda wildcards: "docker://python:3.11",  # callable
+            None,  # unset
+        ],
+    )
+    def test_dynamic_values_classified(self, image):
+        assert EnvironmentValidatorSnakemake._is_dynamic_image(image) is True
+
+    @pytest.mark.parametrize("image", ["python:3.11", "ghcr.io/org/image:tag"])
+    def test_static_values_classified(self, image):
+        assert EnvironmentValidatorSnakemake._is_dynamic_image(image) is False
+
+    def test_wildcard_container_skipped(self):
+        validator = _make_validator(
+            [{"name": "wild", "environment": "python:{version}"}]
+        )
+        assert _run_validation(validator) == []
+        assert _dynamic_warnings(validator) == [
+            "Skipping image validation for 1 rule with a dynamic container "
+            "directive: 'wild'."
+        ]
+        assert not any(m["type"] == "error" for m in validator.messages)
+
+    def test_static_steps_still_validated_alongside_dynamic_ones(self):
+        validator = _make_validator(
+            [
+                {"name": "wild", "environment": "python:{version}"},
+                {"name": "static", "environment": "python:3.11"},
+            ]
+        )
+        assert _run_validation(validator) == ["python:3.11"]
+        assert len(_dynamic_warnings(validator)) == 1
+
+    def test_global_container_warns_once_for_all_rules(self):
+        # A single top-level ``container:`` applies to every rule, so
+        # reana-commons puts the same dynamic value on every executable step.
+        # That must produce one warning naming the rules, not one warning each.
+        validator = _make_validator(
+            [
+                {"name": name, "environment": "python:{version}"}
+                for name in ("a", "b", "c")
+            ]
+        )
+        assert _run_validation(validator) == []
+        assert _dynamic_warnings(validator) == [
+            "Skipping image validation for 3 rules with a dynamic container "
+            "directive: 'a', 'b', 'c'."
+        ]
+
+    def test_many_dynamic_rules_are_summarised(self):
+        validator = _make_validator(
+            [{"name": f"r{i}", "environment": "python:{version}"} for i in range(8)]
+        )
+        assert _run_validation(validator) == []
+        assert _dynamic_warnings(validator) == [
+            "Skipping image validation for 8 rules with a dynamic container "
+            "directive: 'r0', 'r1', 'r2', 'r3', 'r4' and 3 more."
+        ]
+
+
+# ---------------------------------------------------------------------------
+# Wiring from the REANA specification
+# ---------------------------------------------------------------------------
+
+
+class TestValidateEnvironmentWiring:
+    """``validate_environment`` must build the validator from the loaded steps."""
+
+    def test_steps_forwarded(self):
+        captured = {}
+
+        def fake_validate(self):
+            captured["steps"] = self.workflow_steps
+
+        steps = [{"name": "make", "environment": "python:3.11"}]
+        reana_yaml = {
+            "workflow": {
+                "type": "snakemake",
+                "file": "Snakefile",
+                "specification": {"steps": steps},
+            },
+        }
+        with patch.object(
+            EnvironmentValidatorSnakemake, "validate", fake_validate
+        ), patch.object(EnvironmentValidatorSnakemake, "display_messages"):
+            validate_environment(reana_yaml)
+        assert captured["steps"] == steps


tiborsimko:
> Currently, the Snakefile is extracted from the REANA yaml but the related issue (#747) suggests "extracting all Snakefile paths". Is it possible to have multiple Snakefiles? If so, I assume it would be preferable to look in the `workflow/snakemake` directory parse them all?

We cannot rely on a `workflow/snakemake` like path structure, since a researcher could structure their repositories differently. And Snakemake has a directive to include subworkflows from other files, see https://snakemake.readthedocs.io/en/stable/snakefiles/modularization.html#includes. So ideally we should really "dive into snakemake" to get a structured representation of the DAG, and parse that in order to extract all the container image names.

reana_client/validation/environments.py
```diff
@@ -501,3 +504,34 @@ def validate_environment(self):
...
+    def _parse_snakefile(self, filename):
+        """Parse Snakefile to extract container/environment directives."""
+        try:
+            with open(filename, "r") as f:
+                content = f.read()
+
+            # Look for container directives in Snakefile
+            # Pattern: container: "docker://image:tag"
+            return re.findall(
+                r'(?:container|singularity):\s*"docker://([^"]+)"', content
```
tiborsimko, line 520: This works nicely for a first approach, but does not address the situation where the main workflow file would include subworkflows, such as:

```
#include: "Snakemake-subworkflow1"
```

Could you have a look whether it would be easily possible to obtain full DAG structure after Snakemake reads and expands the specs and rather parse that?

CameronMcClymont:
The implementation now handles subworkflows that are added to Snakefiles via `include`.

Adding `include: "Snakefile-htcondorcern"` to the main Snakefile for the root6-roofit demo; changing the rule names in Snakefile-htcondorcern (to avoid clashes); and then updating one of the container images to `docker://docker.io/reanahub/reana-env-root6:6.02.12` confirms that subworkflow image validation is working:

```
$ reana-client validate --environments --pull -f reana-snakemake.yaml

==> Verifying REANA specification file...
...
  -> SUCCESS: Environment image reanahub/reana-env-root6:6.02.12 has the correct format.
  -> SUCCESS: Environment image reanahub/reana-env-root6:6.02.12 exists locally.
  -> SUCCESS: Environment image reanahub/reana-env-root6:6.02.12 exists in Docker Hub.
  -> INFO: Environment image uses UID 0 but will run as UID 1000.
  -> SUCCESS: Environment image reanahub/reana-env-root6:6.18.04 has the correct format.
  -> SUCCESS: Environment image reanahub/reana-env-root6:6.18.04 exists locally.
  -> SUCCESS: Environment image reanahub/reana-env-root6:6.18.04 exists in Docker Hub.
  -> INFO: Environment image uses UID 0 but will run as UID 1000.
```

reana_client/validation/environments.py
```diff
@@ -501,3 +503,68 @@ def validate_environment(self):
...
+                }
+            )
+
+        except Exception as e:
+            self.messages.append(
+                {
+                    "type": "error",
+                    "message": f"Error building Snakemake DAG: {str(e)}",
+                }
+            )
```
michaelbuchar: Snakemake image validation failures are swallowed and can exit 0.
_validate_snakefile_environments() catches all exceptions, including EnvironmentValidationError raised by _validate_environment_image(), appends a type: "error" message, and returns. But validate() only exits non-zero when EnvironmentValidationError escapes. Result: invalid/missing images extracted from the Snakefile can print an error while reana-client validate --environments still succeeds.

reana_client/validation/environments.py
```diff
@@ -501,3 +503,68 @@ def validate_environment(self):
...
+
+            # Load workflow
+            with SnakemakeApi() as snakemake_api:
+                config = {
+                    key: "fake_value" for key in self._get_snakemake_config_keys()
+                }
+                workflow_api = snakemake_api.workflow(
+                    resource_settings=ResourceSettings(),
+                    config_settings=ConfigSettings(config=config),
+                    snakefile=Path(self.workflow_filename),
```
michaelbuchar: The Snakemake config file is ignored and replaced with fake values.
The code scans only the main Snakefile for config["..."] keys and passes every key as "fake_value" to Snakemake. This means workflows using inputs.parameters.input: config.yaml can be evaluated with fake config instead of real REANA input parameters. If container: config["image"], the validator validates fake_value, not the real image. If includes/rules depend on config values, valid workflows can fail or be misvalidated.

reana_client/validation/environments.py
```diff
@@ -501,3 +503,68 @@ def validate_environment(self):
...
+                )
+
+                # Extract all container images
+                wf = workflow_api._workflow
+                images = {rule.container_img for rule in wf.rules if rule.container_img}
+
+                # Trim prefixes
+                images = {image.replace("docker://docker.io/", "") for image in images}
```
michaelbuchar: docker:// image URI normalization only works for Docker Hub.
The code strips only docker://docker.io/. Valid Snakemake container URIs like docker://ghcr.io/org/image:tag or docker://gitlab-registry.cern.ch/group/image:tag are passed into _validate_image_tag() with the scheme intact, where the first colon is treated as the image/tag separator. That misparses the image and can fail validation incorrectly.

michaelbuchar:
No regression tests were added for the Snakemake validation path.
The PR changes only reana_client/validation/environments.py. Given this issue is specifically about Snakemake environment validation silently doing nothing, maybe we can add tests for static containers, included Snakefiles, config-file-driven containers, non-Docker-Hub docker:// URIs, and fatal exit behavior on invalid images.

reana_client/validation/environments.py
CameronMcClymont: Ahh of course, nice catch. Just for my own understanding, we can't just inject fake values into the config used in the Snakefile in case one of the values is used to specify a container image. Makes sense! Fixing now...

reana_client/validation/environments.py
CameronMcClymont: Ah yes, thanks. Generalised to work for non Docker Hub images now.

michaelbuchar:
There are some merge conflicts here as well.. :/

reana_client/validation/environments.py
michaelbuchar: This comment still holds

reana_client/validation/environments.py
```diff
@@ -501,3 +518,98 @@ def validate_environment(self):
...
+                    config.setdefault(key, "fake_value")
+                config_settings = ConfigSettings(config=config)
+
+            # Load workflow with Snakemake API
+            with SnakemakeApi() as snakemake_api:
+                workflow_api = snakemake_api.workflow(
+                    resource_settings=ResourceSettings(),
+                    config_settings=config_settings,
+                    snakefile=Path(self.workflow_filename),
+                )
```
michaelbuchar: Relative Snakemake configfile: paths next to a Snakefile in a subdirectory still break. The PR builds the DAG with snakefile=Path(self.workflow_filename) at lines 579-584 (line 579), but it never switches context to the Snakefile directory. Reproduction: with workflow.file: workflow/Snakefile and workflow/Snakefile containing configfile: "config.yaml", validation looked for /repo/config.yaml instead of /repo/workflow/config.yaml and emitted a DAG error. That breaks a valid REANA layout and also weakens the intended support for included subworkflows that depend on config.

CameronMcClymont: Fixed by changing into the Snakefile's directory before building the DAG, then changing back.

reana_client/validation/environments.py
CameronMcClymont: Oops, missed this, thanks. Handled by re-raising any `EnvironmentValidationError` so it's not swallowed by the general `except Exception as e` block.

reana_client/validation/environments.py
```diff
@@ -537,3 +558,117 @@ def validate_environment(self):
...
+
+            if configfile:
+                config_settings = ConfigSettings(
+                    # Resolve to absolute before chdir so the path stays valid.
+                    configfiles=[Path(configfile).resolve()],
+                    config=direct_params,
+                )
+            else:
+                config = dict(direct_params)
+                for key in self._get_snakemake_config_keys():
```
michaelbuchar: Included Snakefiles are still missed here. This only scans the top-level Snakefile for config[...] keys, so a workflow that does include: "rules.smk" and uses container: config["image"] inside the included file can still fail DAG construction before image validation runs. I reproduced this against the current PR head with Error building Snakemake DAG: 'image'.

CameronMcClymont: Oh I see, yeah I guess this is worth handling, it adds a bit of complexity though. So now the scan is recursive and follows every `include` directive it finds, collecting all config references before building the DAG.

reana_client/validation/environments.py
```diff
@@ -537,3 +558,136 @@ def validate_environment(self):
...
+                        for image in images
+                    }
+
+                    # Validate all collected images
+                    for image in images:
+                        self._validate_environment_image(image)
+            finally:
+                os.chdir(original_dir)
+
+        except ImportError:
```
michaelbuchar: The except ImportError is still too broad. It wraps the full Snakemake DAG-build path starting at line 596, so any ImportError raised while evaluating the Snakefile or an included file gets downgraded to Snakemake is not installed, skipping image validation. I reproduced this by making workflow_api.workflow() raise ImportError("No module named custom_helper"); the validator recorded only that warning and returned successfully. That means a real workflow-side import failure can still make validate --environments look successful instead of surfacing the actual error. The handler should be narrowed to the Snakemake import statements, or at least only treat missing snakemake imports as the “not installed” case.

CameronMcClymont: Thanks, moved the Snakemake imports to their own try/catch block now.

reana_client/validation/environments.py
```diff
@@ -537,3 +558,140 @@ def validate_environment(self):
...
+                    # Validate all collected images
+                    for image in images:
+                        self._validate_environment_image(image)
+            finally:
+                os.chdir(original_dir)
+
+        except EnvironmentValidationError:
+            raise
+
+        except Exception as e:
```
michaelbuchar: Snakemake DAG-build failures still let validate --environments succeed. In environments.py (line 691), non-EnvironmentValidationError exceptions are converted into an "error" message and returned instead of failing validation. validate() only exits non-zero when EnvironmentValidationError escapes. I reproduced this on the current PR head (a46d1b8): a Snakefile that fails during DAG construction returned normally from validate() while recording Error building Snakemake DAG: .... That means a real Snakemake validation failure can still exit 0.

reana_client/validation/environments.py
```diff
@@ -537,3 +558,140 @@ def validate_environment(self):
...
+                k: v for k, v in self.workflow_input_parameters.items() if k != "input"
+            }
+
+            if configfile:
+                config_settings = ConfigSettings(
+                    # Resolve to absolute before chdir so the path stays valid.
+                    configfiles=[Path(configfile).resolve()],
+                    config=direct_params,
+                )
+            else:
```
michaelbuchar: The fake-config fallback still breaks valid workflows that use nested config values for container selection. In environments.py (line 648), the fallback injects "fake_value" for each discovered top-level config["key"]. That works for flat keys, but not for patterns like container: config["images"]["tool"]: config["images"] becomes a string, so Snakemake fails before image validation runs. I reproduced this with a real adjacent config.yaml plus configfile: "config.yaml" in the Snakefile; the validator produced string indices must be integers, not 'str' and never reached _validate_environment_image. The new tests cover flat config cases only, e.g. test_validate_snakemake_environments.py (line 433), so this gap is still untested.

reana_client/validation/environments.py
michaelbuchar: Thanks!

CameronMcClymont:
We've kind of got ourselves in a tangle after the many reviews, so I have kind of overhauled the implementation and simplified things. I've designed 5 cases below to demonstrate that this PR is working as intended. The cases assume a deployment with the following Helm values:

```
vetted_container_images:
  enabled: true # Toggle container image vetting
  allowlist: ["python:3.9", "python:3.11", "reanahub/reana-env-root6:6.18.04", ]
```

### Case 1 (PASS): allowed image written directly in Snakefile

reana.yaml:

```
version: 0.9.0
workflow:
  type: snakemake
  file: Snakefile
inputs:
  parameters:
    sample_count: 5
```

Snakefile:

```
rule all:
    input: "report.txt"

rule analyse:
    container: "docker://python:3.11"
    output: "report.txt"
    shell: "echo {config[sample_count]} > {output}"
```

### Case 2: allowed image defined in separate config file declared via the `configfile` directive

reana.yaml:

```
version: 0.9.0
workflow:
  type: snakemake
  file: Snakefile
inputs:
  files:
    - config.yaml
```

Snakefile:

```
configfile: "config.yaml"

rule all:
    input: "report.txt"

rule analyse:
    container: config["image"]
    output: "report.txt"
    shell: "echo {config[sample_count]} > {output}"
```

config.yaml:

```
image: docker://python:3.9
sample_count: 5
```

### Case 3 (PASS): allowed image defined in separate config file, but overridden in `reana.yaml`

reana.yaml:

```
version: 0.9.0
workflow:
  type: snakemake
  file: Snakefile
inputs:
  files:
    - config.yaml
  parameters:
    image: docker://python:3.11
```

Snakefile:

```
configfile: "config.yaml"

rule all:
    input: "report.txt"

rule analyse:
    container: config["image"]
    output: "report.txt"
    shell: "echo {config[sample_count]} > {output}"
```

config.yaml:

```
image: docker://python:3.9
sample_count: 5
```

### Case 4 (FAIL): disallowed image defined in separate config file and referenced in `Snakefile`

reana.yaml:

```
version: 0.9.0
workflow:
  type: snakemake
  file: Snakefile
inputs:
  files:
    - config.yaml
```

Snakefile:

```
configfile: "config.yaml"

rule all:
    input: "report.txt"

rule analyse:
    container: config["image"]
    output: "report.txt"
    shell: "echo {config[sample_count]} > {output}"
```

config.yaml:

```
image: docker://python:3.10
sample_count: 5
```

### Case 5 (FAIL): disallowed image defined in `Snakefile`, but overridden by allowed image in `reana.yaml`

This is the expected behaviour. Even though the disallowed image will never be used (because it's overridden), it should still not be referenced anywhere.

reana.yaml:

```
version: 0.9.0
workflow:
  type: snakemake
  file: Snakefile
inputs:
  files:
    - config.yaml
  parameters:
    image: docker://python:3.11
```

Snakefile:

```
configfile: "config.yaml"

rule all:
    input: "report.txt"

rule analyse:
    container: config["image"]
    output: "report.txt"
    shell: "echo {config[sample_count]} > {output}"
```

config.yaml:

```
image: docker://python:3.10
sample_count: 5
```

michaelbuchar:
Looks good to me, moved to in-merge!

michaelbuchar (dismissed):

reana_client/validation/environments.py
```diff
@@ -537,3 +557,78 @@ def validate_environment(self):
...
+        try:
+            os.chdir(snakefile_path.parent)
+            with SnakemakeApi() as snakemake_api:
+                workflow = snakemake_api.workflow(
+                    resource_settings=ResourceSettings(),
+                    config_settings=config_settings,
+                    snakefile=snakefile_path,
+                )._workflow
+                images = {
+                    rule.container_img for rule in workflow.rules if rule.container_img
```
tiborsimko: **🤖 RC748-01: Dynamic `container:` directives crash validation or fail it falsely (MEDIUM)**

`rule.container_img` holds the *raw* value of the `container:` directive. Snakemake resolves that value only per job — `Rule.expand_container_img` (snakemake `rules.py:1256-1267`) calls the value if it is a callable and applies wildcards if it is a wildcard-parameterised string, and jobs consume the result via `Job.container_img_url`. The new extraction in `reana_client/validation/environments.py:618-620` reads the rule-level raw value, so two valid Snakemake constructs misbehave (both reproduced with Snakemake 9.22.0):

```python
container: "docker://python:{version}"
```

is extracted and validated as the literal `python:{version}` — the registry lookup can never succeed, so a workflow that runs fine is reported as failing validation; and

```python
container: lambda wildcards: "docker://python:3.11"
```

makes `rule.container_img` a function object, so `image.startswith(prefix)` at `reana_client/validation/environments.py:632` raises `AttributeError: 'function' object has no attribute 'startswith'`. The prefix-stripping loop sits *after* the `try/except` that wraps the workflow construction, and `validate()` only catches `EnvironmentValidationError`, so this surfaces as a raw traceback rather than a validation message. (Terminology aside: the code constructs the workflow via `snakemake_api.workflow(...)` without ever calling `workflow_api.dag()`, so the docstring and the `"Error building Snakemake DAG"` message slightly oversell what happens — "loading the workflow" would be more accurate.)

These are not hypothetical constructs on the REANA side: the runtime engine resolves images per job precisely to support them — `reana-workflow-engine-snakemake/reana_workflow_engine_snakemake/executor.py:302-303` uses `job.container_img_url`. The wildcard form reaches runtime today and fails client validation falsely; callable containers are supported by the engine but also expose a separate reana-commons loading bug (spec loading crashes on them at `reana_commons/snakemake.py:338`), which needs the same hardening in a separate reana-commons PR.

Importantly, the Snakefile extraction is not the only affected path: reana-commons ≥ 0.95.0a21 copies the same raw `rule.container_img` into each step's `environment` (`reana_commons/snakemake.py:338`), and `validate_environment()` validates those step images *first* (`reana_client/validation/environments.py:546-559`). A wildcard image therefore already fails at the step loop before `_validate_snakefile_environments()` runs, so a guard on the extraction alone is not enough — both loops need it. (The step loop must keep validating static images: on Python 3.8–3.10 the Snakemake API pass is skipped entirely, leaving the step loop as the only validation path, and it is also what carries `kubernetes_uid` validation.)

Suggested fix — the highest-fidelity fix would be to enumerate DAG *jobs* and validate `job.container_img_url`, i.e. exactly the values the runtime engine consumes. This works (verified empirically: both examples above resolve to `python:3.11`), but it has real costs with the Snakemake 9 API: it requires the private `_prepare_dag()`/`_build_dag()` methods on top of the `._workflow` access already used, it creates a `.snakemake/` metadata directory in the user's analysis directory as a side effect of DAG initialisation, and it only covers rules reachable from the configured targets, so images in unreached rules would silently escape validation.

Given those trade-offs, the pragmatic fix for this PR is to keep the rule-level walk (full rule coverage, no side effects) and skip dynamic values with an explicit warning, as a deliberate and documented limitation. A classification-only helper used by *both* loops:

```python
@staticmethod
def _is_dynamic_image(image):
    """Return True when a container value is resolved per job by Snakemake."""
    return not isinstance(image, str) or "{" in image
```

Since both loops see the same rules, emit the warning from a single place keyed by rule name (e.g. a small set of already-warned names on the validator), so each dynamic rule warns exactly once:

```python
"message": f"Rule '{name}' uses a dynamic container directive, "
"skipping its image validation.",
```

Please cover both dynamic forms with tests: the wildcard case should go through `validate_environment()` end to end (steps populated the way reana-commons populates them, i.e. with the raw wildcard string) and should assert exactly one warning and no image-validation call for the dynamic rule — not a false failure; the callable case can only be tested at the extraction level until the reana-commons spec-loading fix lands, and must not produce a traceback.

reana_client/validation/environments.py
```diff
@@ -537,3 +557,78 @@ def validate_environment(self):
...
+        # case is not confused with an import error raised while evaluating the
+        # workflow (which must surface as a validation failure instead).
+        try:
+            from snakemake.api import SnakemakeApi
+            from snakemake.settings.types import ResourceSettings, ConfigSettings
+        except ImportError:
+            self.messages.append(
+                {
+                    "type": "warning",
+                    "message": "Snakemake is not installed, skipping image validation.",
```
tiborsimko: **🤖 RC748-02: "Snakemake is not installed" warning is wrong on Python < 3.11 (LOW)**

reana-commons pins `snakemake==7.32.4 ; python_version<'3.11'` and `snakemake==9.22.0 ; python_version>='3.11'`, so on Python 3.8–3.10 Snakemake *is* installed — only the Snakemake 8+ API (`snakemake.api`) is missing. The ImportError fallback at `reana_client/validation/environments.py:591` then reports:

```python
"message": "Snakemake is not installed, skipping image validation.",
```

On those Pythons the new Snakefile image validation silently no-ops while telling the user a dependency is missing that `pip list` reports as present, sending them down a false debugging path. (The graceful skip itself is fine and the test suite documents it deliberately — only the wording is off.)

Suggested fix — describe the actual condition, which also stays accurate for a genuinely missing installation or a broken import:

```python
"message": "Snakemake 8+ API is not available, skipping Snakefile image validation.",
```

The existing `test_snakemake_not_installed_warns_without_error` asserts `"not installed" in m["message"]`, so its assertion should be updated together with the wording — and renaming it to something like `test_snakemake_api_unavailable_warns_without_error` would keep the name honest too.

reana_client/validation/environments.py
```diff
@@ -8,9 +8,9 @@
 
 """REANA client environment validation."""
 
+import os
```
tiborsimko: **🤖 RC748-03: Copyright year list not extended to 2026 (NIT)**

The header of `reana_client/validation/environments.py` (line 4) still reads:

```python
# Copyright (C) 2021, 2022, 2023, 2024, 2025 CERN.
```

The commit was authored in 2025 but recommitted in 2026, and the PR will be merged in 2026, so the year list should include 2026. The PR's new test file already gets this right with `Copyright (C) 2026 CERN.`

Suggested fix —

```python
# Copyright (C) 2021, 2022, 2023, 2024, 2025, 2026 CERN.
```

reana_client/validation/environments.py
```diff
@@ -71,15 +77,29 @@ def build_validator(workflow):
 class EnvironmentValidatorBase:
     """REANA workflow environments validation base class."""
 
-    def __init__(self, workflow_steps=None, pull=False, access_token=None):
+    def __init__(
+        self,
+        workflow_steps=None,
+        pull=False,
+        access_token=None,
+        workflow_filename=None,
```
tiborsimko: **🤖 RC748-04: Snakemake-only parameters live on the shared base class (NIT)**

`workflow_filename` and `workflow_input_parameters` — including docstring semantics that are specific to Snakemake (configfile via `inputs.parameters.input`, `--config` overrides) — are added to `EnvironmentValidatorBase.__init__` (`reana_client/validation/environments.py:80-102`), yet the Serial/Yadage/CWL validators never receive them.

Suggested fix — give `EnvironmentValidatorSnakemake` its own `__init__` that accepts the two extra keyword arguments and forwards the rest, keeping the base class engine-agnostic:

```python
class EnvironmentValidatorSnakemake(EnvironmentValidatorBase):
    def __init__(
        self, workflow_filename=None, workflow_input_parameters=None, **kwargs
    ):
        """..."""
        super().__init__(**kwargs)
        self.workflow_filename = workflow_filename
        self.workflow_input_parameters = workflow_input_parameters or {}
```

reana_client/validation/environments.py
CameronMcClymont: Extended to include 2026.

reana_client/validation/environments.py
CameronMcClymont: Reworded the message to "Snakemake 8+ API is not available, skipping Snakefile image validation", reworded the test name, and updated the assertion.

reana_client/validation/environments.py
CameronMcClymont: Ah yes much better design, moved these args to EnvironmentValidatorSnakemake as suggested.

reana_client/validation/environments.py
CameronMcClymont: Applied the suggested fix: added `_is_dynamic_image()` plus a helper into both loops (warning once per rule). The limitation is documented in the docstring too along with updated 'DAG' terminology. Tests covering the wildcard case are also added.

reana_client/validation/environments.py
tiborsimko: I guess you forgot to push the updated code?

reana_client/validation/environments.py
```diff
@@ -535,5 +586,95 @@ def validate_environment(self):
...
                 )
                 continue
+            if self._is_dynamic_image(image):
+                self._warn_dynamic_container(step.get("name"))
+                continue
             kubernetes_uid = step.get("kubernetes_uid")
             self._validate_environment_image(image, kubernetes_uid=kubernetes_uid)
+
+        # Extract and validate environments from Snakefile
+        self._validate_snakefile_environments()
```
tiborsimko: **🤖 RC748-05: Environment validation reparses Snakemake independently of the canonical loader (MEDIUM)**

By the time `EnvironmentValidatorSnakemake` runs, `load_validate_reana_spec()` has already called `reana-commons` to load the Snakefile. Since `reana-commons` 0.95.0a21, that loader expands includes and populates `workflow.specification.steps` with the container and `kubernetes_uid` of every executable rule. `validate_environment()` validates those steps and then calls `_validate_snakefile_environments()`, which executes the same user-authored Snakefile again through a separate, partially reconstructed Snakemake configuration.

The canonical load and the second pass receive different configuration. `load_workflow_spec_from_reana_yaml()` runs before `load_input_parameters()`, while `inputs.parameters.input` still names the external config file, but it forwards only that file and ignores direct REANA parameter overrides. I reproduced the resulting divergence with an internal `configfile:` selecting `python:3.9` and a direct REANA override selecting `python:3.11`: `workflow.specification.steps` records `python:3.9`, whereas the runtime engine and this PR's second pass select `python:3.11`. The persisted specification can therefore describe the wrong image, and client capability validation can vet a different image from the one the workflow will execute. This is not an execution-time allowlist bypass—the job controller checks the actual `docker_img` when each job is created—but it delays rejection until workflow execution instead of reporting the real image during client-side capability validation.

The second pass compensates for that bug at the wrong lifecycle point. It applies the pre-normalisation contract—`input` is a config-file path—to the post-normalisation parameter mapping, where `load_input_parameters()` may already have replaced that path with the file contents. If those contents contain a legitimate `input` parameter, the validator treats its value as another config file, removes the real `config["input"]` override, and can raise from `ConfigSettings` before entering the structured error handler. It also executes arbitrary top-level Snakefile Python twice and inspects `norun` rules that the canonical steps deliberately exclude, so validation can fail on an image that no job will pull. The helper-level tests bypass the real load/normalise/validate lifecycle and consequently make these divergent paths appear supported.

A separate `reana-commons` parameter-loading bug must not be hidden by this refactor: `load_input_parameters()` replaces `inputs.parameters` wholesale. A source specification containing both `input: config.yaml` and a sibling override such as `n: 5` silently loses `n` before persistence, so the runtime never receives it either. The canonical-loader change must explicitly define whether file-based and direct parameters can coexist and merge; it must not assume that they already do.

Suggested fix — first update `reana-commons` so `_prepare_kwargs()` forwards both the external config file and the direct parameter mapping (excluding the reserved file-pointer `input` key), and thread both through the version-specific Snakemake loaders—for Snakemake 8+, via `ConfigSettings(configfiles=..., config=...)`. Preserve the same configuration precedence as the runtime engine and ensure the same final mapping is persisted and passed to runtime. Track and resolve the wholesale-replacement bug above rather than building on an assumed merge. Add end-to-end tests for an internal `configfile:` plus a direct override, an external input config file, and the chosen behaviour when both an input file and sibling direct parameters are supplied. Merge and release that `reana-commons` change first; then raise both reana-client minimum `reana-commons` requirements in `setup.py` to that release, rebase this PR, and reduce it to validation of `workflow.specification.steps`: remove `_validate_snakefile_environments()`, `workflow_filename`, `workflow_input_parameters`, and the isolated second-pass tests, while retaining the dynamic-image guard in the step loop.

reana_client/validation/environments.py
```diff
@@ -535,5 +586,95 @@ def validate_environment(self):
...
+                k: v for k, v in self.workflow_input_parameters.items() if k != "input"
+            },
+        )
+
+        # Change into the Snakefile's directory so that relative paths declared
+        # inside it (e.g. ``configfile: "config.yaml"``) resolve against the
+        # Snakefile location rather than the directory validation runs from.
+        original_dir = os.getcwd()
+        try:
+            os.chdir(snakefile_path.parent)
```
tiborsimko: **🤖 RC748-06: Changing into the Snakefile directory reverses runtime `configfile:` semantics (MEDIUM)**

REANA executes Snakemake with the workflow workspace as `workdir`, so relative `configfile:` paths are workspace-relative. The normal `reana-commons` loader follows the same behaviour when validation is launched from the project root. `_validate_snakefile_environments()` instead resolves the Snakefile and changes into its parent directory before loading it:

```python
snakefile_path = Path(self.workflow_filename).resolve()
...
os.chdir(snakefile_path.parent)
```

For a runtime-valid layout with `workflow.file: workflow/Snakefile`, a config file at `workflow/config.yaml`, and this directive:

```python
configfile: "workflow/config.yaml"
```

the normal loader succeeds and the runtime engine resolves `<workspace>/workflow/config.yaml`. The new second pass instead looks for `<workspace>/workflow/workflow/config.yaml`, raises `EnvironmentValidationError`, and makes `validate --environments` exit 1. `test_relative_configfile_in_subdirectory_resolves` does not cover the real command lifecycle: it invokes the private second-pass helper directly, so it never encounters the workspace-relative resolution used by the canonical loader before environment validation starts.

This also means the earlier fix for discussion `3374028419` was based on the wrong path premise: changing directory made the isolated helper test pass, but diverged from the workflow that REANA actually executes.

Suggested fix — resolve this together with RC748-05 by removing the independent second parser and using one runtime-faithful workflow representation. If the second pass is retained temporarily, do not change directory before loading the Snakefile, matching the canonical loader, and add an integration test that starts with `load_validate_reana_spec()` and the workspace-relative layout above rather than testing `_validate_snakefile_environments()` in isolation.

reana_client/validation/environments.py
```diff
@@ -523,10 +534,50 @@ def validate_environment(self):
...
+        concrete job are known (see ``Rule.expand_container_img``).  The rule-level
+        value is then not a usable image name and cannot be validated.
+        """
+        return not isinstance(image, str) or "{" in image
+
+    def _warn_dynamic_container(self, rule_name):
+        """Warn once per rule that its dynamic container image is not validated."""
+        if rule_name in self._dynamic_container_rules:
+            return
+        self._dynamic_container_rules.add(rule_name)
```
tiborsimko: **🤖 RC748-07: A global dynamic container emits one warning for every rule (LOW)**

Snakemake propagates a top-level dynamic container directive to every rule, for example:

```python
container: "docker://python:{version}"
```

`reana-commons` consequently places the same dynamic image value on every executable workflow step. `_warn_dynamic_container()` deduplicates by rule name, so `validate --environments` emits one warning for every rule even though the researcher wrote one global directive. The second Snakefile pass additionally sees `norun` rules such as `rule all`; removing that pass as proposed in RC748-05 eliminates those extra warnings but still leaves one warning per executable step. On a sizeable workflow, the repeated warnings can bury the other validation results without adding new information.

Suggested fix — collect the affected dynamic rules during validation and emit one aggregated warning after the step loop, including either a concise rule list or a count. Avoid merely deduplicating on the image value, because separately declared rules can legitimately use identical dynamic expressions and should still be represented in the aggregate. Add a test with one global dynamic container and multiple rules that asserts a single warning and identifies the affected executable rules.

CameronMcClymont: As suggested, dynamic rules are now collected during validation, aggregated, and reported once. Deduplication happens on rule name. (pushing soon)

reana_client/validation/environments.py
CameronMcClymont: Fixed in a new reana-commons PR (https://github.com/reanahub/reana-commons/pull/555) so direct `inputs.parameters` overrides are now forwarded into the Snakemake loaders and file-based params are merged with direct params. So the second parse that was in this PR is now dropped along with workflow_filename and workflow_input_parameters.

reana_client/validation/environments.py
CameronMcClymont: Resolved as a byproduct of addressing RC748-05 - no second parse means no directory change to get wrong.

reana_client/validation/environments.py
CameronMcClymont: Btw, since this reana-client PR now depends on the new reana-commons changes, the reana-commons pin is bumped to an unreleased a22 here as well.

reana_client/validation/environments.py
tiborsimko: Verified at the current head (`e757591`). The client-side Snakefile scanner and
parser have been removed, and environment validation now consumes the steps
produced by the canonical reana-commons loader. Reproduced both with a main
Snakefile that includes a second rules file and with Snakemake's `module` /
`use rule ... from` composition:

```text
include:         [('sub', 'python:3.9'), ('main', 'python:3.11')]
module/use-rule: [('other_remote', 'python:3.9'), ('main', 'python:3.11')]
```

Included and module-composed rules reach the same environment-validation path
as main-workflow rules, with their respective images. The expanded-DAG design
therefore covers both forms without another client-side scan. This design
question is resolved.

reana_client/validation/environments.py
tiborsimko: **🤖 RC748-01 follow-up (2026-08-19): direct handling improved; end-to-end dynamic validation remains partial (MEDIUM remains)**

The amended client no longer reparses the Snakefile. It now recognises
wildcard-parameterised step images, skips their direct image validation, and
emits an aggregated warning instead of validating the unresolved literal or
raising a traceback. This addresses the original failure in the environment
validator itself.

The companion `reanahub/reana-commons#555` representation does not yet carry
this distinction through the complete path: callable containers become empty
images, while wildcard strings remain unresolved literals. The client
consequently reports the wrong default for callables, and server-capability
preflight can reject wildcards before the new skip warning is reached.

Suggested fix — preserve and centrally classify a serialisable dynamic marker
in `reanahub/reana-commons#555`, then recognise it before the client's
empty-image fallback and during image extraction. Preflight can defer marked
values to the job controller's existing resolved-image check without weakening
allowlist enforcement. Until both callable and wildcard forms follow that path,
the original finding remains partially addressed.

reana_client/validation/environments.py
tiborsimko: **🤖 RC748-02 follow-up (2026-08-19): verified fixed at the current head (`e757591`)**

The independent Snakemake API pass has been removed, along with its import
fallback and the inaccurate "Snakemake is not installed" warning. Environment
validation now consumes the steps produced by the canonical reana-commons
loader on every supported Python version.

The original finding is fully addressed.

reana_client/validation/environments.py
tiborsimko: **🤖 RC748-03 follow-up (2026-08-19): verified fixed at the current head (`e757591`)**

The copyright header of `reana_client/validation/environments.py` now includes
2026, preserving the complete existing year list.

The original finding is fully addressed.

reana_client/validation/environments.py
tiborsimko: **🤖 RC748-04 follow-up (2026-08-19): verified fixed at the current head (`e757591`)**

`EnvironmentValidatorBase` once again contains only engine-agnostic state. The
Snakemake-specific filename and input-parameter arguments disappeared together
with the independent parser, and the Snakemake subclass now initialises only
the state used to aggregate its dynamic-container warnings.

The original finding is fully addressed.

reana_client/validation/environments.py
tiborsimko: **🤖 RC748-05 follow-up (2026-08-19): duplicate parsing removed; configuration fidelity remains partial (MEDIUM remains)**

The client-side second parse has been removed. Environment validation now uses
only `workflow.specification.steps` from the canonical reana-commons loader,
avoiding duplicate Snakefile execution, inspection of non-executable rules, and
the earlier pre-/post-normalisation confusion. The companion reana-commons
change also forwards direct REANA parameters into the Snakemake loader and
preserves siblings of `inputs.parameters.input`.

The companion `reanahub/reana-commons#555` configuration path is still not
identical to runtime: the external parameter file reaches the loader through
`configfiles=`, runtime receives only the persisted mapping through `config=`,
and persistence merges nested values differently. Consequently, the loader can
produce steps from values unavailable to the engine or accept an internal
`configfile:` case that the engine rejects.

Suggested fix — in `reanahub/reana-commons#555`, read and merge the external
parameter file once with Snakemake's own file semantics, pass the result solely
through `config=`, and persist that same mapping. The architectural part of the
original finding is fixed, but its load/runtime fidelity requirement remains
partially addressed.

reana_client/validation/environments.py
tiborsimko: **🤖 RC748-06 follow-up (2026-08-19): verified fixed at the current head (`e757591`)**

The independent Snakefile parse and its directory change have been removed.
The client therefore no longer applies a second path-resolution context that
can disagree with the workspace-relative semantics used by the canonical
loader and workflow engine.

The original finding is fully addressed.

reana_client/validation/environments.py
tiborsimko: **🤖 RC748-07 follow-up (2026-08-19): wildcard warnings aggregated; callable path remains partial (LOW remains)**

Wildcard-parameterised step images are now collected by rule name and reported
in one bounded aggregate warning. A global wildcard container no longer emits
one warning per executable rule, and the new tests cover both small and larger
rule sets.

Callable containers do not reach this aggregation path: reana-commons
serialises them as empty images, after which the client emits the anonymous
"environment image not specified" warning once per rule. Reproduced with a
three-rule global callable container:

```text
steps:    [('a', ''), ('b', ''), ('c', '')]
messages: 3 × "Environment image not specified, using docker.io/snakemake/snakemake:v9.22.0."
```

This still produces the repeated output the finding was meant to avoid, while
also describing the runtime image incorrectly.

Suggested fix — preserve callable containers as explicitly dynamic in the
companion `reanahub/reana-commons#555` step representation and check that marker
before the client's empty-image fallback. Both wildcard and callable forms can
then use the same aggregate warning path. The original finding remains
partially addressed until that representation is shared end to end.

CameronMcClymont:
Closing since the validation of Snakemake environments is now done server-side. #779 moved workflow loading and validation to the server and reanahub/reana-commons#547 added `reana_commons.validation.images`.

One gap was left which is now addressed in reanahub/reana-commons#558. A `container:` directive may be a callable or a wildcard-templated string, which Snakemake expands per job once that job's wildcards are known.

Issue #739, opened 2025-01-28
logs: better capturing of Dask and job logs in error situations


tiborsimko:
### Current behaviour

When a workflow using Dask fails, the errors may not be fully captured and exposed back to users.

Here is one observation from running `reana-demo-dask-coffea` example.

The status shows that the workflow failed:

```console
$ rcg status -w dask-coffea-serial-kubernetes
NAME                            RUN_NUMBER   CREATED               STARTED               ENDED                 STATUS   PROGRESS
dask-coffea-serial-kubernetes   1            2025-01-28T10:02:38   2025-01-28T10:02:51   2025-01-28T10:04:03   failed   0/1
```

The logs show ditto:

```console
$ rcg logs -w dask-coffea-serial-kubernetes

==> Workflow engine logs
2025-01-28 10:02:51,384 | root | MainThread | INFO | Publishing step:0, cmd: python analysis.py, total steps 1 to MQ
2025-01-28 10:04:03,456 | root | MainThread | INFO | Workflow a98dfa47-b7a2-4979-b75f-44ee4b224776 finished. Files available at /var/reana/users/00000000-0000-0000-0000-000000000000/workflows/a98dfa47-b7a2-4979-b75f-44ee4b224776.



==> Job logs
==> Step: process
==> Workflow ID: a98dfa47-b7a2-4979-b75f-44ee4b224776
==> Compute backend: Kubernetes
==> Job ID: reana-run-job-d60a2b71-c263-46c5-8bb1-06793e9083b4
==> Docker image: docker.io/coffeateam/coffea-dask-cc7:0.7.22-py3.10-g7f049
==> Command: python analysis.py
==> Status: failed
==> Started: 2025-01-28T10:02:51
==> Finished: 2025-01-28T10:03:59
==> Logs:

...


                -------------------------------------------------------------------
                -------------------------------------------------------------------
                ----------------         DASK WORKER LOGS          ----------------
                -------------------------------------------------------------------
                -------------------------------------------------------------------




                /usr/local/lib/python3.10/site-packages/distributed/cli/dask_worker.py:264: FutureWarning: dask-worker is deprecated and will be removed in a future release; use `dask worker` instead
  warnings.warn(
2025-01-28 10:02:56,570 - distributed.nanny - INFO -         Start Nanny at: 'tcp://10.244.0.62:44707'
2025-01-28 10:02:57,354 - distributed.worker - INFO -       Start worker at:    tcp://10.244.0.62:41989
[...]

PR #749, opened 2025-07-16
feat(validation): add workflow image vetting to validate command

### LINKED ISSUE #728: validate: check workflow specification for unathorised container images
In some deployment scenarios we may want to allow running only those workflows which use vetted authorised container images.

(This is similar to how we allow to run only certain vetted Jupyter notebook images by users, see `interactive_sessions.environments.jupyter.recommended` Helm value.)

To address these deployment scenarios, it would be useful to:

- allow REANA cluster administrators to specify which images can be used by users in their runtime workflows;
- enrich `reana-client info` to report to users which images are authorised;
- amend `reana-client validate` to report any troubles early in case a user tries to use an unauthorised image;
- make strong check on the REANA server side not to accept user workflows if the workflow uses an unauthorised image for some of its steps.

The implementation could be similar to how we handle Jupyter images, i.e. via a Helm value. However, the list of authorised images may grow, so we may have to be ready to handle potentially large number of allowed items in the vetted container image list.

CameronMcClymont:
Extend `reana-client validate` to check if container images used in workflows are authorized to run on the cluster. The list of authorized images is obtained from the `reana-server` `info` endpoint.

Closes reanahub/reana-server/issues/728

### DIFF
--- a/reana_client/validation/environments.py
+++ b/reana_client/validation/environments.py
@@ -20,7 +20,6 @@ from reana_commons.config import (
 )
 from reana_commons.utils import run_command
 
-
 from reana_client.errors import EnvironmentValidationError
 from reana_client.config import (
     DOCKER_REGISTRY_INDEX_URL,
@@ -28,11 +27,12 @@ from reana_client.config import (
     ENVIRONMENT_IMAGE_SUSPECTED_TAGS_VALIDATOR,
     GITLAB_CERN_REGISTRY_INDEX_URL,
     GITLAB_CERN_REGISTRY_PREFIX,
+    ERROR_MESSAGES,
 )
 from reana_client.printer import display_message
 
 
-def validate_environment(reana_yaml, pull=False):
+def validate_environment(reana_yaml, pull=False, access_token=None):
     """Validate environments in REANA specification file according to workflow type.
 
     :param reana_yaml: Dictionary which represents REANA specification file.
@@ -43,17 +43,23 @@ def validate_environment(reana_yaml, pull=False):
         workflow_type = workflow["type"]
         if workflow_type == "serial":
             workflow_steps = workflow["specification"]["steps"]
-            return EnvironmentValidatorSerial(workflow_steps=workflow_steps, pull=pull)
+            return EnvironmentValidatorSerial(
+                workflow_steps=workflow_steps, pull=pull, access_token=access_token
+            )
         if workflow_type == "yadage":
             workflow_steps = workflow["specification"]["stages"]
-            return EnvironmentValidatorYadage(workflow_steps=workflow_steps, pull=pull)
+            return EnvironmentValidatorYadage(
+                workflow_steps=workflow_steps, pull=pull, access_token=access_token
+            )
         if workflow_type == "cwl":
             workflow_steps = workflow.get("specification", {}).get("$graph", workflow)
-            return EnvironmentValidatorCWL(workflow_steps=workflow_steps, pull=pull)
+            return EnvironmentValidatorCWL(
+                workflow_steps=workflow_steps, pull=pull, access_token=access_token
+            )
         if workflow_type == "snakemake":
             workflow_steps = workflow["specification"]["steps"]
             return EnvironmentValidatorSnakemake(
-                workflow_steps=workflow_steps, pull=pull
+                workflow_steps=workflow_steps, pull=pull, access_token=access_token
             )
 
     workflow = reana_yaml["workflow"]
@@ -65,7 +71,7 @@ def validate_environment(reana_yaml, pull=False):
 class EnvironmentValidatorBase:
     """REANA workflow environments validation base class."""
 
-    def __init__(self, workflow_steps=None, pull=False):
+    def __init__(self, workflow_steps=None, pull=False, access_token=None):
         """Validate environments in REANA workflow.
 
         :param workflow_steps: List of dictionaries which represents different steps involved in workflow.
@@ -73,8 +79,10 @@ class EnvironmentValidatorBase:
         """
         self.workflow_steps = workflow_steps
         self.pull = pull
+        self.access_token = access_token
         self.validated_images = set()
         self.messages = []
+        self._vetting_config = None  # cached (vetting_enabled, allowlist); False = skip
 
     def validate(self):
         """Validate REANA workflow environments."""
@@ -98,15 +106,63 @@ class EnvironmentValidatorBase:
                 indented=True,
             )
 
+    def check_image_authorized(self, image):
+        """Checks if an image is authorized for use in workflows.
+
+        :param image: Full image name with tag.
+        """
+        if self._vetting_config is False:
+            return  # There was no token or fetch error, skip
+
+        if not self.access_token:
+            self.messages.append(
+                {
+                    "type": "warning",
+                    "message": f'No access token provided - skipping container image authorisation check. {ERROR_MESSAGES["missing_access_token"]}',
+                }
+            )
+            self._vetting_config = False
+            return
+
+        if self._vetting_config is None:
+            # Vetting config has not been cached yet - fetch and cache
+            try:
+                from reana_client.api.client import info
+
+                cluster_info = info(self.access_token)
+                vetting = cluster_info.get("vetted_container_images_enabled")
+                allowlist = cluster_info.get("vetted_container_images_allowlist")
+                if vetting is None or allowlist is None:
+                    # Server predates vetting support (skip the image vetting check)
+                    self._vetting_config = False
+                    return
+                self._vetting_config = (vetting["value"], allowlist["value"])
+            except Exception as e:
+                self.messages.append(
+                    {
+                        "type": "warning",
+                        "message": f"Could not check if container images are authorised. Is the cluster running properly? Error: {e}",
+                    }
+                )
+                self._vetting_config = False
+                return
+
+        # Validate image against allowlist
+        vetting_enabled, allowlist = self._vetting_config
+        if vetting_enabled and image not in allowlist:
+            raise EnvironmentValidationError(
+                f"Environment image is not allowed: {image}"
+            )
+
     def _validate_environment_image(self, image, kubernetes_uid=None):
         """Validate image environment.
 
         :param image: Full image name with tag if specified.
         :param kubernetes_uid: Kubernetes UID defined in workflow spec.
         """
-
         if image not in self.validated_images:
             image_name, image_tag = self._validate_image_tag(image)
+            self.check_image_authorized(image)
             exists_locally, _ = self._image_exists(image_name, image_tag)
             if exists_locally or self.pull:
                 uid, gids = self._get_image_uid_gids(image_name, image_tag)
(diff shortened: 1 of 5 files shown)


reana_client/validation/environments.py
```diff
@@ -98,16 +105,53 @@ def display_messages(self):
                 indented=True,
             )
 
+    def check_image_authorized(self, image):
```
michaelbuchar, line 109: This function appends an error message but does not raise EnvironmentValidationError. The command prints the error and can still exit successfully. Maybe raise an exception or explicitly fail the command, and add a CLI exit code test as well.

reana_client/validation/environments.py
```diff
@@ -98,16 +105,53 @@ def display_messages(self):
...
+        """Checks if an image is authorized for use in workflows.
+
+        :param image: Full image name with tag.
+        """
+        if self.access_token:
+            try:
+                from reana_client.api.client import info
+
+                cluster_info = info(self.access_token)
```
michaelbuchar, line 132: This info request occurs inside the per-image method. Fetch cluster info once and cache the allowlist. Also clarify whether vetting should remain opt-in behind validate --environments, plain reana-client validate currently skips it.

reana_client/validation/environments.py
CameronMcClymont: Updated to raise EnvironmentValidationError in line with other validation checks.

reana_client/validation/environments.py
CameronMcClymont: Thanks, now fetching cluster info once and caching the allowlist.

Vetting should indeed remain opt-in as it is part of the `--environments` validation which was already opt-in before these changes (which I assume we want to leave alone).

reana_client/validation/environments.py
```diff
@@ -98,16 +106,61 @@ def display_messages(self):
...
         """Validate image environment.
 
         :param image: Full image name with tag if specified.
         :param kubernetes_uid: Kubernetes UID defined in workflow spec.
         """
-
         if image not in self.validated_images:
             image_name, image_tag = self._validate_image_tag(image)
             exists_locally, _ = self._image_exists(image_name, image_tag)
+            self.check_image_authorized(image)
```
tiborsimko, line 165: **🤖 Image-existence check runs before vetting check (LOW)**

`reana-client/reana_client/validation/environments.py:160-163`:

```python
if image not in self.validated_images:
    image_name, image_tag = self._validate_image_tag(image)
    exists_locally, _ = self._image_exists(image_name, image_tag)
    self.check_image_authorized(image)
```

For a disallowed image, the client hits DockerHub/GitLab first and only then checks the allowlist. If the image is *also* missing from the registry, `_image_exists` raises `EnvironmentValidationError("…does not exist locally or remotely.")` before the more actionable "image is not allowed" can fire — the user gets the less relevant error.

Vetting is one cached `info` call; gating the registry lookup behind it is strictly better for UX (and avoids a secondary information-leak concern of probing third-party registries for images the admin has already disallowed).

Suggested fix — reorder so the cluster's allowlist decides first:

```python
if image not in self.validated_images:
    image_name, image_tag = self._validate_image_tag(image)
    self.check_image_authorized(image)
    exists_locally, _ = self._image_exists(image_name, image_tag)
    ...
```

reana_client/validation/environments.py
```diff
@@ -98,16 +106,61 @@ def display_messages(self):
...
+            # Vetting config has not been cached yet - fetch and cache
+            try:
+                from reana_client.api.client import info
+
+                cluster_info = info(self.access_token)
+                self._vetting_config = (
+                    cluster_info["vetted_container_images_enabled"]["value"],
+                    cluster_info["vetted_container_images_allowlist"]["value"],
+                )
+            except Exception as e:
```
tiborsimko, line 140: **🤖 New client against old server reports a misleading cluster warning (LOW)**

`reana-client/reana_client/validation/environments.py:127-145`:

```python
cluster_info = info(self.access_token)
self._vetting_config = (
    cluster_info["vetted_container_images_enabled"]["value"],
    cluster_info["vetted_container_images_allowlist"]["value"],
)
...
except Exception as e:
    self.messages.append({
        "type": "warning",
        "message": f"Could not check if container images are authorised. "
                   f"Is the cluster running properly? Error: {e}",
    })
```

When a newer `reana-client` (which knows about vetting) talks to a server that hasn't yet shipped #739, the `info` response doesn't carry the new fields, the indexing raises `KeyError`, and the user sees "Is the cluster running properly?" — which sends them off chasing a phantom cluster outage. `reana-client` ships independently on PyPI, so this transitional skew is real.

Suggested fix — treat absent vetting fields as "server doesn't advertise this feature, skip the authorisation check without warning"; keep the broad `except` as a real network/auth fallback:

```python
vetting = cluster_info.get("vetted_container_images_enabled")
allowlist = cluster_info.get("vetted_container_images_allowlist")
if vetting is None or allowlist is None:
    self._vetting_config = False  # server doesn't advertise vetting
    return
self._vetting_config = (vetting["value"], allowlist["value"])
```

reana_client/validation/environments.py
tiborsimko, line 165: **🤖 Snakemake default-image vetting diverges between client and server (MEDIUM)**

A Snakemake rule without a `container:` directive surfaces as `step["environment"] == ""` in the loaded REANA spec. The two validators handle that empty string very differently.

Server — `reana-server/reana_server/validation.py:130-136` via `reana_commons.validation.images.extract_images`:

```python
for image in extract_images(reana_yaml):
    if image and image not in allowed_images:
        raise REANAValidationError(f"Image not allowed: {image}")
```

The empty string is filtered out, so the workflow passes (the runtime substitutes `REANA_DEFAULT_SNAKEMAKE_ENV_IMAGE`, which is admin-controlled).

Client — `reana-client/reana_client/validation/environments.py:520-536`:

```python
if not image:
    self.messages.append({"type": "warning", "message": f"Environment image not specified, using {REANA_DEFAULT_SNAKEMAKE_ENV_IMAGE}."})
    image = REANA_DEFAULT_SNAKEMAKE_ENV_IMAGE
...
self._validate_environment_image(image, kubernetes_uid=kubernetes_uid)
```

It substitutes the default image first, *then* runs it through `_validate_environment_image` → `check_image_authorized`. If the admin enables vetting but forgets to allowlist `docker.io/snakemake/snakemake:v7.32.4`, the user's Snakemake workflow passes server-side validation and submits cleanly, but `reana-client validate --environments -t …` fails with `Environment image is not allowed: docker.io/snakemake/snakemake:v7.32.4`. That asymmetry will confuse users — the typical assumption is that `reana-client validate` mirrors what the server will accept.

Suggested fix — pick one of:

1. (preferred) Skip the authorisation check when the image is the substituted default, matching the server's "empty == admin-controlled" treatment. The client can still emit its existing "image not specified" warning.
2. Extend the server's `extract_images` for Snakemake to substitute the default in the same way, and let it be vetted in both places — but then the chart needs `REANA_DEFAULT_SNAKEMAKE_ENV_IMAGE` to be allowlisted by default, which is a deployment-doc burden.

(1) is the smaller change and matches the "admin owns the default" intent that the new server tests already encode (`snakemake-no-container-uses-admin-default`).

CameronMcClymont: Ah that's a nice catch. Fixed using the first suggestion so the warning is still printed but the substituted default image is no longer validated (matching reana-server).

reana_client/validation/environments.py
CameronMcClymont: Fixed so that when either vetting field is absent, the function returns instead of printing the cluster warning (as suggested).

reana_client/validation/environments.py
CameronMcClymont: Nice, yeah fixed by reordering as suggested.

tiborsimko:
BTW a small UX observation: the vetted workflows are accepted for creation, but not for execution. So a user ends up with a workflow in a "created" state:

```console
$ reana-client run -w bsm
==> Creating a workflow...
==> Verifying REANA specification file... /Users/tibor/private/project/reana/src/reana-demo-bsm-search/reana.yaml
==> Verifying REANA specification parameters...
  -> SUCCESS: REANA specification parameters appear valid.
==> Verifying workflow parameters and commands...
  -> WARNING: Yadage parameter "workspace" found on step "makews" is not defined in input parameters.
==> Verifying dangerous workflow operations...
  -> SUCCESS: Workflow operations appear valid.
==> Verifying compute backends in REANA specification file...
  -> SUCCESS: Workflow compute backends appear to be valid.
bsm.1
==> SUCCESS: File /reana.yaml was successfully uploaded.
==> Uploading files...
==> SUCCESS: File /workflow/databkgmc.yml was successfully uploaded.
==> SUCCESS: File /workflow/allmc_input.yml was successfully uploaded.
==> SUCCESS: File /workflow/workflow_sig.yml was successfully uploaded.
==> SUCCESS: File /tests/log-messages.feature was successfully uploaded.
==> SUCCESS: File /workflow/workflow_mc.yml was successfully uploaded.
==> SUCCESS: File /workflow/steps.yml was successfully uploaded.
==> SUCCESS: File /tests/run-duration.feature was successfully uploaded.
==> SUCCESS: File /workflow/workflow_data.yml was successfully uploaded.
==> SUCCESS: File /workflow/inputsm.yml was successfully uploaded.
==> SUCCESS: File /tests/workspace-files.feature was successfully uploaded.
==> SUCCESS: File /workflow/workflow_select_shape.yml was successfully uploaded.
==> SUCCESS: File /workflow/wflow_all_mc.yml was successfully uploaded.
==> SUCCESS: File /workflow/inputsig.yml was successfully uploaded.
==> Starting workflow...
==> ERROR: Cannot start workflow bsm.1:
Image not allowed: docker.io/reanahub/reana-demo-bsm-search:1.0.0

$ rcg list --filter name=bsm
NAME   RUN_NUMBER   CREATED               STARTED   ENDED   STATUS
bsm    1            2026-06-05T10:12:25   -         -       created
```

Compare this to the behaviour of Dask, which will be refused upfront if REANA does not support it:

```console
$ reana-client run -w dask
==> Creating a workflow...
==> Verifying REANA specification file... /Users/tibor/private/project/reana/src/reana-demo-dask-coffea/reana.yaml
==> Verifying REANA specification parameters...
  -> SUCCESS: REANA specification parameters appear valid.
==> Verifying workflow parameters and commands...
  -> SUCCESS: Workflow parameters and commands appear valid.
==> Verifying dangerous workflow operations...
  -> SUCCESS: Workflow operations appear valid.
==> Verifying compute backends in REANA specification file...
  -> SUCCESS: Workflow compute backends appear to be valid.
==> ERROR: Cannot create workflow dask:
Dask workflows are not allowed in this cluster.

$ rcg list --filter name=dask
NAME   RUN_NUMBER   CREATED   STARTED   ENDED   STATUS
```

IOW, attempting to run an unsupported Dask workflow will not pollute the workflow list, whilst attempting to run an unsupported container image would.

We could parse the workflow DAG upfront to see all containers that the user is using, and refuse workflow creation in case a non-vetted container is being used.
 
(This could also be taken into a future improvement issue.)

CameronMcClymont:
Extended `_validate_server_capabilities` in `utils.py` to address the comment above.

Result:

```
reana-client run -w bitcoin
==> Creating a workflow...
==> Verifying REANA specification file... /Users/cmcclymont/Documents/REANA/reana-hub/reana-demo-root6-roofit/reana.yaml
==> Verifying REANA specification parameters... 
  -> SUCCESS: REANA specification parameters appear valid.
==> Verifying workflow parameters and commands... 
  -> SUCCESS: Workflow parameters and commands appear valid.
==> Verifying dangerous workflow operations... 
  -> SUCCESS: Workflow operations appear valid.
==> Verifying compute backends in REANA specification file...
  -> SUCCESS: Workflow compute backends appear to be valid.
==> Verifying container images in REANA specification file...
  -> ERROR: Environment image is not allowed: docker.io/reanahub/bitcoin:6.18.04
```

```
./reana-client-go list --filter name=bitcoin
NAME   RUN_NUMBER   CREATED   STARTED   ENDED   STATUS
```

setup.py
```diff
@@ -26,7 +26,7 @@
         "sphinx-click>=1.0.4",
     ],
     "tests": [
-        "reana-commons[kubernetes,tests]>=0.95.0a17,<0.96.0",
+        "reana-commons[kubernetes,tests]>=0.95.0a20,<0.96.0",
```
tiborsimko, line 29: FYI bumped `reana-commons` in order to fix tests and repushed.

PR #750, opened 2025-07-21
feat(workflow): add JSON output for sharing commands 

### LINKED ISSUE #712: cli: add JSON output for workflow sharing commands
Add JSON output for workflow sharing commands.

Issue was created after a comment by Marco: https://github.com/reanahub/reana-client/pull/683#issuecomment-1884487052

CameronMcClymont:
Add JSON output support to the share-add and share-remove commands.
Return an empty JSON array from share-status when no users are present.
Add tests for successful, failed, and partially successful operations.

Closes #712

### DIFF
--- a/reana_client/cli/workflow.py
+++ b/reana_client/cli/workflow.py
@@ -45,6 +45,7 @@ from reana_client.config import (
     CLI_LOGS_FOLLOW_DEFAULT_INTERVAL,
     MAX_RUN_LABELS_SHOWN,
     CLI_WORKFLOWS_LIST_MAX_RESULTS,
+    JSON,
 )
 from reana_client.printer import display_message
 from reana_client.utils import (
@@ -1751,19 +1752,24 @@ def workflow_close_interactive_session(workflow, access_token):  # noqa: D301
     type=click.DateTime(formats=["%Y-%m-%d"]),
     help="Optional date when access to the workflow will expire for the given user(s) (format: YYYY-MM-DD).",
 )
+@click.option(
+    "--json",
+    "output_format",
+    flag_value="json",
+    default=None,
+    help="Get output in JSON format.",
+)
 @click.pass_context
 def workflow_share_add(
-    ctx, workflow, access_token, users, message, valid_until
+    ctx, workflow, access_token, users, message, valid_until, output_format
 ):  # noqa D412
     """Share a workflow with other users (read-only).
 
     The `share-add` command allows sharing a workflow with other users. The
     users will be able to view the workflow but not modify it.
 
-    Examples:
-
-    \t $ reana-client share-add -w myanalysis.42 --user bob@example.org
-
+    Examples:\n
+    \t $ reana-client share-add -w myanalysis.42 --user bob@example.org\n
     \t $ reana-client share-add -w myanalysis.42 --user bob@example.org --user cecile@example.org --message "Please review my analysis" --valid-until 2025-12-31
     """
     from reana_client.api.client import share_workflow
@@ -1789,14 +1795,24 @@ def workflow_share_add(
             share_errors.append(f"Failed to share {workflow} with {user}: {str(e)}")
             logging.debug(traceback.format_exc())
 
-    if shared_users:
+    if output_format == JSON:
+        result = {
+            "workflow": workflow,
+            "shared_with": shared_users,
+            "errors": share_errors,
+        }
         display_message(
-            f"{workflow} is now read-only shared with {', '.join(shared_users)}",
-            msg_type="success",
+            json.dumps(result, indent=2),
         )
+    else:
+        if shared_users:
+            display_message(
+                f"{workflow} is now read-only shared with {', '.join(shared_users)}",
+                msg_type="success",
+            )
 
-    for error in share_errors:
-        display_message(error, msg_type="error")
+        for error in share_errors:
+            display_message(error, msg_type="error")
 
     if share_errors:
         sys.exit(1)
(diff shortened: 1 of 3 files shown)


michaelbuchar:
ReadTheDocs CI check is failing

michaelbuchar:
### `share-status --json` still has a non-JSON empty-state path
Issue #712 asks for JSON output for workflow-sharing commands. share-status already has --json, but when shared_with is empty it still prints human text via display_message(..., msg_type="info"), so automation expecting JSON gets plain text. This is pre-existing but the PR closing the issue should probably cover it [workflow.py#L1946-L1950](https://github.com/reanahub/reana-client/blob/af543c86b5eb3d64b395478c79851d0bac3ab5dd/reana_client/cli/workflow.py#L1946-L1950)

michaelbuchar:
No tests cover the new JSON output paths, maybe we can add some more covering JSON success, partial failure or exit code behavior? The existing tests only cover text output

michaelbuchar:
I removed this PR from the Project since we already have the issue tracking the status of this

reana_client/cli/workflow.py
michaelbuchar: share-status --json still leaks text-table formatting into the JSON output for workflows shared without an expiry date. In workflow.py (line 1941), valid_until=None is converted to "-" before the data is handed to the JSON formatter at workflow.py (line 1946). That means the machine readable path will emit "valid_until": "-" instead of JSON null, which is a lossy representation and awkward for automation.

CameronMcClymont: Sure, that's a nice improvement. Fixed so that when there's no expiry date, `None` is used instead of `-` in the json representation.

michaelbuchar: Thanks, looks good!

michaelbuchar (dismissed):

reana_client/cli/workflow.py
```diff
@@ -1910,6 +1945,9 @@ def share_workflow_status(
 
         display_formatted_output(data, headers, format_, output_format)
```
tiborsimko, line 1950: **🤖 `share-status --json` returns `"valid_until": "-"` instead of `null` (MEDIUM)**

`reana_client/cli/workflow.py:1938-1946` builds the data row passed to `display_formatted_output`:

```python
data = [
    [entry["user_email"],
     (entry["valid_until"] if entry["valid_until"] is not None else "-")]
    for entry in shared_with
]
display_formatted_output(data, headers, format_, output_format)
```

The `"-"` placeholder is a *table-display* convention for an empty cell. When `--json` is set, the same data flows into `tablib_data.export("json")`, so users with no expiration date come out as `{"user_email": "...", "valid_until": "-"}` in JSON. Machine consumers see the string `"-"` and have to special-case it — that's a footgun for the very scripts the flag is meant to enable. The new `test_share_status_workflow_json_success` includes a row with `valid_until: None` but never asserts on its value, so the bug slips through the test suite.

Suggested fix — localise the change to `share-status` rather than touching `display_formatted_output`, which serves many other commands. Branch on `output_format` when building the data row so the `"-"` substitution applies only to the table renderer:

```python
shared_with = sharing_status.get("shared_with", [])
if shared_with:
    headers = ["user_email", "valid_until"]
    if output_format == JSON:
        data = [[entry["user_email"], entry["valid_until"]] for entry in shared_with]
    else:
        data = [
            [entry["user_email"],
             entry["valid_until"] if entry["valid_until"] is not None else "-"]
            for entry in shared_with
        ]
    display_formatted_output(data, headers, format_, output_format)
```

Then strengthen the test to lock in the contract:

```python
alice = next(r for r in parsed if r["user_email"] == "alice@example.org")
assert alice["valid_until"] is None
```

tests/test_cli_workflows.py
```diff
@@ -1496,3 +1496,371 @@ def test_share_status_workflow():
...
+    import json
+
+    status_code = 200
+    response = {
+        "message": "is not shared with anyone",
+        "workflow_id": "string",
+        "workflow_name": "string",
+    }
+    env = {"REANA_SERVER_URL": "localhost"}
+    mock_http_response, mock_response = Mock(), Mock()
```
tiborsimko: **🤖 Redundant `import json` and dead `Mock()` in `test_share_status_workflow_json_empty` (NIT)**

`tests/test_cli_workflows.py` — the new test has:

```python
def test_share_status_workflow_json_empty():
    """Test share-status --json outputs valid JSON when workflow is not shared."""
    import json

    status_code = 200
    response = {
        "message": "is not shared with anyone",
        "workflow_id": "string",
        "workflow_name": "string",
    }
    env = {"REANA_SERVER_URL": "localhost"}
    mock_http_response, mock_response = Mock(), Mock()
    mock_http_response.status_code = status_code
    mock_response = response
```

`json` is already imported at the module level — the other new tests use `json.loads` without re-importing, so the local `import json` is dead. The `Mock()` assigned to `mock_response` is overwritten on the very next line by `mock_response = response`, so the constructor call also does nothing. Neither is harmful, but both make the test read like a half-finished copy-paste.

Suggested fix — drop the local `import json`, and split the dual-assignment so the unused `Mock()` doesn't appear at all:

```python
mock_http_response = Mock()
mock_http_response.status_code = status_code
mock_response = response
```

reana_client/cli/workflow.py
tiborsimko: Addressed in a new fixup commit, kept separate on top of the feat commit for review clarity. This fixup commit can be squashed into the feat commit before merge, preserving the first author with `Co-authored-by` for the fixup.

tests/test_cli_workflows.py
tiborsimko: Addressed in a new fixup commit, kept separate on top of the feat commit for review clarity. This fixup commit can be squashed into the feat commit before merge, preserving the first author with `Co-authored-by` for the fixup.

michaelbuchar (approved):
LGTM!

PR #759, opened 2026-02-03
feat(workflow): guard deletion of restarted runs

### LINKED ISSUE #448: Deleting a restarted workflow run removes the original workflow workspace
### **Description**

When restarting a workflow and deleting the restarted run via the UI, the workspace of the original workflow run is removed, leaving the original run in an inconsistent (out-of-sync) state.

This results in a finished workflow whose workspace no longer exists on disk, which is later detected by `reana-admin check-workflows` as an error.

---

### **Steps to reproduce**

1. Run a workflow:

   ```bash
   reana-client run -w helloworld-demo
   ```

   This creates workflow run `helloworld-demo.3`.

2. Restart the workflow:

   ```bash
   reana-client restart -w helloworld-demo
   ```

   This creates workflow run `helloworld-demo.3.1`.

3. Delete the restarted workflow run (`helloworld-demo.3.1`) using the REANA UI. Do not check the `Delete all the runs of the workflow` checkbox.

4. Run the admin consistency check:

   ```bash
   kubectl exec -i -t deployment/reana-server -- flask reana-admin check-workflows --date-start '2025-12-01' --admin-access-token $REANA_ADMIN_ACCESS_TOKEN   
   ```

---

### **Observed behavior**

The admin check reports the original workflow run (`helloworld-demo.3`) as out-of-sync:

```
ERROR: The workspace '/var/reana/users/.../workflows/<workflow-id>' does not exist.
```

The original workflow run:

* Is marked as `finished` in the database
* No longer has a corresponding workspace on disk

---

### **Expected behavior**

One of the following should happen instead:

1. The workflow workspace of the original run should **not** be deleted when deleting a restarted run
2. The user should be clearly warned that deleting the restarted run will also remove the workspace of the original run and result in data loss
3. All related runs (e.g. `3` and `3.1`) should have their statuses updated consistently (e.g. both marked as deleted).

In any case, deleting a restarted run should not leave the original run in a broken, out-of-sync state.

michaelbuchar:
Closes https://github.com/reanahub/reana-ui/issues/448

### DIFF
--- a/reana_client/cli/workflow.py
+++ b/reana_client/cli/workflow.py
@@ -33,12 +33,18 @@ from reana_client.cli.utils import (
     requires_environments,
     retrieve_workflow_logs,
     follow_workflow_logs,
+    parse_workflow_run_number,
+    get_run_number_major_key,
+    format_run_number_label,
+    format_run_label_list,
 )
 from reana_client.config import (
     ERROR_MESSAGES,
     RUN_STATUSES,
     TIMECHECK,
     CLI_LOGS_FOLLOW_DEFAULT_INTERVAL,
+    MAX_RUN_LABELS_SHOWN,
+    CLI_WORKFLOWS_LIST_MAX_RESULTS,
 )
 from reana_client.printer import display_message
 from reana_client.utils import (
@@ -95,7 +101,7 @@ def workflow_sharing_group(ctx):
     "_format",
     multiple=True,
     help="Format output according to column titles or column values. "
-    "Use `<columm_name>=<column_value>` format. "
+    "Use `<column_name>=<column_value>` format. "
     "E.g. display workflow with failed status and named test_workflow "
     "`--format status=failed,name=test_workflow`.",
 )
@@ -132,7 +138,7 @@ def workflow_sharing_group(ctx):
     "filters",
     multiple=True,
     help="Filter workflow that contains certain filtering criteria. "
-    "Use `--filter <columm_name>=<column_value>` pairs. "
+    "Use `--filter <column_name>=<column_value>` pairs. "
     "Available filters are ``name`` and ``status``.",
 )
 @click.option(
@@ -1262,6 +1268,15 @@ def workflow_run(
     help="Delete all runs of a given workflow.",
 )
 @click.option(
+    "--include-all-restarts",
+    "include_all_restarts",
+    is_flag=True,
+    help=(
+        "Delete all restarted runs that share the same workspace as the selected run. "
+        "Without this flag, deletion will fail if the run is part of a restart chain."
+    ),
+)
+@click.option(
     "--include-workspace",
     "should_delete_workspace",
     is_flag=True,
@@ -1271,8 +1286,13 @@ def workflow_run(
 @add_access_token_options
 @check_connection
 @click.pass_context
-def workflow_delete(
-    ctx, workflow: str, all_runs: bool, should_delete_workspace: bool, access_token: str
+def workflow_delete(  # noqa: C901
+    ctx,
+    workflow: str,
+    all_runs: bool,
+    include_all_restarts: bool,
+    should_delete_workspace: bool,
+    access_token: str,
 ):  # noqa: D301
     """Delete a workflow.
 
@@ -1285,7 +1305,11 @@ def workflow_delete(
     \t $ reana-client delete -w myanalysis.42\n
     \t $ reana-client delete -w myanalysis.42 --include-all-runs
     """
-    from reana_client.api.client import delete_workflow
+    from reana_client.api.client import (
+        delete_workflow,
+        get_workflow_status,
+        get_workflows,
+    )
     from reana_client.utils import get_api_url
 
     should_delete_workspace = True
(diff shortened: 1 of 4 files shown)


reana_client/cli/workflow.py
```diff
@@ -1296,15 +1319,120 @@ def workflow_delete(
 
     if workflow:
         try:
+
```
tiborsimko: Style: no black after `try`?

reana_client/cli/workflow.py
```diff
@@ -1297,14 +1321,119 @@ def workflow_delete(
...
+                    workflow_base
                 )
-            else:
-                message = get_workflow_status_change_msg(workflow, "deleted")
-            display_message(message, msg_type="success")
+                display_message(message, msg_type="success")
+                return
+
+            # Otherwise, detect whether this run has restarts
+            wf_status = get_workflow_status(workflow, access_token) or {}
```
tiborsimko: ### AI.1 - Redundant `get_workflow_status` call

In `workflow.py`, `get_workflow_status` is called inside the `if all_runs:` block (~line 1324) and then again unconditionally at ~line 1340 for the non-`all_runs` path. The early `return` inside `if all_runs:` makes this technically correct, but the structure is confusing to follow.

**Suggestion:** Restructure the function as an explicit `if/else` so each branch has exactly one `get_workflow_status` call, making the control flow immediately obvious.

```python
if all_runs:
    wf_status = get_workflow_status(workflow, access_token) or {}
    # ... all_runs logic ...
else:
    wf_status = get_workflow_status(workflow, access_token) or {}
    # ... single/restart logic ...
```

reana_client/cli/workflow.py
```diff
@@ -1297,14 +1321,119 @@ def workflow_delete(
     if workflow:
         try:
             logging.info("Connecting to {0}".format(get_api_url()))
-            delete_workflow(workflow, all_runs, should_delete_workspace, access_token)
+
+            # If deleting all runs, keep existing behaviour (including restarts)
             if all_runs:
+                # Check if any run for given workflow exists
```
tiborsimko: ### AI.2 - Extra API calls in `--include-all-runs` path

The `all_runs` branch now makes `get_workflow_status` + `get_workflows` calls just to check whether all runs are already deleted, before calling `delete_workflow`. Previously this was a single `delete_workflow` call. This adds two extra HTTP round-trips to every `--include-all-runs` invocation.

**Suggestion:** Check whether the server-side `delete_workflow` already handles the "nothing to delete" case gracefully (e.g. returns a 404 or a no-op). If so, remove the pre-check and let the server handle it. If the server returns an unhelpful error, consider handling it in the `except` block instead.

reana_client/cli/workflow.py
```diff
@@ -1297,14 +1321,119 @@ def workflow_delete(
...
+                wf_status = get_workflow_status(workflow, access_token) or {}
+                full_name = wf_status.get("name") or workflow
+                base_name, _, _ = parse_workflow_run_number(full_name)
```
tiborsimko: ### AI.3 - `workflow_base` parsing changes existing behavior

The old code used `workflow.split(".")[0]` for the success message in the `--include-all-runs` path. The new code uses `parse_workflow_run_number`, which handles dots in workflow names differently. For example, `my.flow.3` now yields base=`my.flow` instead of `my`.

This is arguably a bug fix, but it is a behavioral change visible to users in CLI output.

**Suggestion:** Note this explicitly in the PR description as an intentional fix so reviewers are aware it is not an unintended side effect.

reana_client/cli/workflow.py
```diff
@@ -1285,7 +1305,11 @@ def workflow_delete(
...
     """
-    from reana_client.api.client import delete_workflow
+    from reana_client.api.client import (
+        delete_workflow,
+        get_workflow_status,
+        get_workflows,
+    )
     from reana_client.utils import get_api_url
 
     should_delete_workspace = True
```
tiborsimko, line 1315: ### AI.4 - `should_delete_workspace` is always `True`

At `workflow.py` line ~1313, `should_delete_workspace` is unconditionally set to `True`:

```python
should_delete_workspace = True
```

This overrides whatever the user passed via `--include-workspace`, making that CLI flag dead code.

**Suggestion:** If this is intentional (workspace deletion is now always required), remove the `--include-workspace` option from the command definition to avoid confusing users. If it is not intentional, remove the override so the user's choice is respected.

reana_client/cli/workflow.py
```diff
@@ -1297,14 +1321,119 @@ def workflow_delete(
...
+                    "shared workspace and leave other runs in an inconsistent state.\n"
+                    "Related runs: {}\n"
+                    "Rerun the command with --include-all-restarts to delete this run and its restarts.".format(
+                        full_name, format_run_label_list(labels, MAX_RUN_LABELS_SHOWN)
+                    ),
+                    msg_type="error",
+                )
+                sys.exit(1)
+
+            if has_restart_series and include_all_restarts:
```
tiborsimko: ### AI.5 - Restart deletion loop has no partial-failure handling

In the `has_restart_series and include_all_restarts` branch (~line 1404):

1. The first run is deleted **with** workspace deletion.
2. Remaining runs are deleted in a loop **without** workspace deletion.

If the first deletion succeeds but a subsequent `delete_workflow` call fails, the loop aborts and some runs are left undeleted while their shared workspace is already gone. This is a worse state than before the command was run.

**Suggestion:** Either wrap each deletion in a try/except and collect failures to report at the end, or adopt a two-phase approach: first mark all runs as deleted (without workspace), then delete the workspace in a final step. This way a partial failure leaves runs marked but the workspace intact.

reana_client/cli/workflow.py
```diff
@@ -1297,14 +1321,119 @@ def workflow_delete(
...
+                # Delete workspace once, mark related runs deleted without deleting workspace again
+                workflow_id_or_name = wf_id or full_name
+                delete_workflow(workflow_id_or_name, False, True, access_token)
+                primary_id = wf_id
+                primary_name = full_name
+                for r in related_runs:
+                    rid = r.get("id")
+                    rname = r.get("name")
+                    if (primary_id and rid == primary_id) or (rname == primary_name):
```
tiborsimko: ### AI.6 - `related_runs` identity comparison may be fragile

At ~line 1411, the loop skips the already-deleted primary run by comparing IDs and names:

```python
if (primary_id and rid == primary_id) or (rname == primary_name):
    continue
```

`primary_id` comes from `get_workflow_status` while `rid` comes from `get_workflows`. If the two endpoints return IDs in different formats (e.g. with/without UUID dashes), the comparison could fail and the primary run would be deleted a second time (without workspace), potentially causing a server error.

**Suggestion:** Normalize both IDs before comparison (e.g. strip dashes), or rely solely on the name comparison which is a plain string from both endpoints.

reana_client/cli/workflow.py
```diff
@@ -1297,14 +1321,119 @@ def workflow_delete(
...
+                labels = [format_run_number_label(r.get("name")) for r in related_runs]
+                display_message(
+                    "Workflow run '{}' including its restarts have been deleted ({}).".format(
+                        full_name, format_run_label_list(labels, MAX_RUN_LABELS_SHOWN)
+                    ),
+                    msg_type="success",
+                )
+                return
+
+            # Normal single workflow run delete (no restarts)
```
tiborsimko: ### AI.7 - No CLI-level tests for `workflow_delete` with `--include-all-restarts`

The new tests in `test_cli_utils.py` thoroughly cover the utility functions, but there are no tests for the actual `workflow_delete` command exercising the new `--include-all-restarts` flag. The complex branching logic in `workflow_delete` (restart detection, multi-deletion loop, error messages) is the riskiest part of this PR.

**Suggestion:** Add mock-based CLI tests using `CliRunner` that cover at least:

- Single run with no restarts (normal delete).
- Run that is part of a restart chain without `--include-all-restarts` (expect error + exit 1).
- Run that is part of a restart chain with `--include-all-restarts` (expect all related runs deleted).
- Already-deleted workflow (expect info message, no deletion).
- `--include-all-runs` with all runs already deleted (expect info message).

reana_client/config.py
```diff
@@ -82,3 +82,9 @@
 
 CLI_LOGS_FOLLOW_DEFAULT_INTERVAL = 10
 """Default interval between log requests in seconds."""
+
+MAX_RUN_LABELS_SHOWN = 10
+"""Maximum number of run labels to print in CLI output, extra labels are collapsed as '+N more'."""
+
+CLI_WORKFLOWS_LIST_MAX_RESULTS = 1000
```
tiborsimko, line 89: ### AI.8 - `CLI_WORKFLOWS_LIST_MAX_RESULTS = 1000` may be slow for large accounts

When resolving restart chains, the code fetches up to 1000 workflow runs in a single API call. For users with many workflows sharing the same base name, this could be slow and return far more data than needed.

**Suggestion:** Consider whether a smaller page size with early termination would suffice, since the code only needs runs matching a specific `major_key`. Alternatively, if the server supports filtering by run number prefix, pass that filter to reduce the response size.

reana_client/cli/workflow.py
michaelbuchar: Fixed!

reana_client/cli/workflow.py
michaelbuchar: Pre-check is removed, thanks!

reana_client/cli/workflow.py
michaelbuchar: REANA does not allow `.` in workflow names

reana_client/cli/workflow.py
michaelbuchar: This is intentional and predates this PR. If we want to remove the redundant `--include-workspace` flag, that should be handled separately from this PR.

reana_client/cli/workflow.py
michaelbuchar: Added a try/except block to collect failures.

reana_client/cli/workflow.py
michaelbuchar: I don’t think this is an issue. The primary run is skipped on either matching ID or matching full workflow name, so even a hypothetical ID formatting difference would still be covered by the name comparison. Both `get_workflow_status` and `get_workflows` return the same `workflow.id_` and the same full name format, so I don't think there's evidence of a format mismatch in what we currently have.

reana_client/cli/workflow.py
michaelbuchar: No tests needed

reana_client/config.py
michaelbuchar: This has already been resolved by a previous comment: https://github.com/reanahub/reana-ui/pull/469#discussion_r2769128556 so no action needed.

tiborsimko: Some suggestions for a better commit log headline in order to have more readable release notes:

1. feat(workflow): handle restart series in delete command (#759)
2. feat(workflow): add --include-all-restarts to delete command (#759)
3. feat(workflow): guard deletion of restarted runs (#759)

and the commit log body providing explanation:

    Prevent deleting a single workflow run when it belongs to a
    restart series with a shared workspace, since that would leave 
    sibling runs in an inconsistent state. Add
    --include-all-restarts flag to explicitly delete all runs in
    the restart series together.

    Closes reanahub/reana-ui#448

I guess 2 or 3 work nicely?

(Note also the change of commit log scope.)

michaelbuchar: Done!

tiborsimko (approved):

PR #769, opened 2026-05-04
feat(utils): show Dask service logs

### LINKED ISSUE #768: feat(logs): show Dask service logs
### Current behaviour

For Dask workflows, the service logs are well exposed on the web interface:

Image

But not in Python or Go client:

```console
$ rc logs -w dask-coffea-serial-kubernetes
...
==> Job logs
==> Step: process
==> Workflow ID: 979a25e3-460b-4a29-bd92-680d08e566e6
==> Compute backend: Kubernetes
==> Job ID: reana-run-job-22af4bc4-4d0f-4cc9-a74f-219e6f8cf1d9
==> Docker image: docker.io/coffeateam/coffea-dask-cc7:0.7.22-py3.10-g7f049
==> Command: python analysis.py
==> Status: finished
==> Started: 2026-05-04T08:57:58
==> Finished: 2026-05-04T08:58:53
==> Logs:
job: :
 Matplotlib created a temporary cache directory at /tmp/matplotlib-8z22hpf3 because the default path (/.config/matplotlib) is not a writable directory; it is highly recommended to set the MPLCONFIGDIR environment variable to a writable directory, in particular to speed up the import of Matplotlib and to better support multiprocessing.
all events 53446198
number of chunks 534


Completed
```

### Expected behaviour

The CLI clients should show the service logs just as the Web UI does.

tiborsimko:
The REST API already exposes Dask scheduler and worker pod logs under the `service_logs` key of the workflow logs response, and the Web UI surfaces them in a dedicated tab. The CLI silently dropped the field, leaving Dask users without a terminal-side view of their cluster logs.

Render the captured pod logs in `reana-client logs` after the workflow engine logs and before the job logs.

Closes #768

### DIFF
--- a/reana_client/cli/utils.py
+++ b/reana_client/cli/utils.py
@@ -364,6 +364,34 @@ def output_user_friendly_logs(workflow_logs, steps):
         click.secho(f"{leading_mark} Engine internal logs", fg="yellow")
         click.secho(workflow_logs["engine_specific"])
 
+    # Service logs (e.g. Dask scheduler/worker pod logs)
+    service_logs = workflow_logs.get("service_logs", {})
+    service_names = sorted(
+        service_name for service_name, entries in service_logs.items() if entries
+    )
+    if service_names:
+        click.echo("\n")
+        click.secho(f"{leading_mark} Service logs", bold=True, fg="yellow")
+        for service_name in service_names:
+            entries = service_logs[service_name]
+            click.secho(
+                f"{leading_mark} Service: {service_name}",
+                bold=True,
+                fg="yellow",
+            )
+            for entry in entries:
+                component = entry.get("component", "")
+                content = entry.get("content", "")
+                click.secho(f"{leading_mark} Component: {component}", fg="yellow")
+                if content:
+                    click.secho(f"{leading_mark} Logs:", fg="yellow")
+                    click.secho(content)
+                else:
+                    display_message(
+                        f"Component {component} emitted no logs.",
+                        msg_type="info",
+                    )
+
     returned_step_names = set(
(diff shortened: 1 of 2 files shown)


reana_client/cli/utils.py
```diff
@@ -364,6 +364,29 @@ def output_user_friendly_logs(workflow_logs, steps):
         click.secho(f"{leading_mark} Engine internal logs", fg="yellow")
         click.secho(workflow_logs["engine_specific"])
 
+    # Service logs (e.g. Dask scheduler/worker pod logs)
+    if workflow_logs.get("service_logs"):
```
michaelbuchar: ### AI suggestion

The code checks only `if workflow_logs.get("service_logs"):` before printing the `Service logs` section. However, the server can return a payload like:

`{"service_logs": {"dask-service-123": []}}`

i.e. a service exists but it has no stored log entries. In that case, the CLI will print:

- `==> Service logs`
- `==> Service: dask-service-123`

and then nothing else.

This can happen for legacy/backfilled workflows or runs where no service logs were actually persisted. It would be better either to skip services with empty `entries`, or to print an explicit info message such as “Service dask-service-123 emitted no logs.”

tiborsimko: Thanks, addressed. The renderer now first builds the list of services that actually have log entries, and only prints the `Service logs` section when that list is non-empty. This avoids the empty section for payloads such as `{"service_logs": {"dask-service-123": []}}`.

I also added a regression test for this case. Entries with an existing component but empty `content` still keep the explicit `Component ... emitted no logs.` message.

michaelbuchar (approved):
LGTM

PR #180, opened 2018-09-18
 cwl-runner: upgrade to latest client API

diegodelemos:
* Updated the out of sync `cwl-runner` CLI to use latest API (closes #178).
* Removes unnecessary custom functions to discover file dependencies (closes #147).

PR #772, opened 2026-05-13
feat(quota): expose quota period in CLI commands

### LINKED ISSUE #771: cli: expose quota period and replenishment window
### Summary

Expose periodic quota information in the CLI so that users can understand which accounting window their current usage applies to and when their quota is expected to replenish.

At the moment, the Web UI shows the current CPU quota period window, but the CLI does not surface comparable information.

### Problem

For deployments using periodic CPU quota accounting, the current CLI output does not tell the user:

- whether a periodic quota policy is active
- which quota period the shown CPU usage applies to
- when the current accounting window ends / when quota is expected to replenish

Examples:

```console
$ reana-client quota-show --resource cpu -h
30s out of 16m 40s used (3%)
```

The output above does not indicate the active quota period.

Similarly, `reana-client info` does not expose server-defined quota period information.

### Expected behaviour

Achieve CLI feature parity with the Web UI for quota-period visibility.

Possible improvements:

- `reana-client info` should show whether a periodic CPU quota policy is active
- `reana-client quota-show --resource cpu -h` should display the active accounting window
- the shown period should make it clear when users can expect quota replenishment

For example, the CLI could show something like:

```console
$ reana-client quota-show --resource cpu -h
30s out of 16m 40s used (3% in the period from 2026-05-01 to 2026-08-31) 
```

### Notes

- The backend already exposes `quota_period_months` and `quota_period_start_at`
- The Web UI already renders this information
- A matching follow-up should likely be done for `reana-client-go` as well
- This is intentionally split from the main quota-period backend/UI PR set to keep review scope manageable

michaelbuchar:
Closes #771

### DIFF
--- a/reana_client/cli/ping.py
+++ b/reana_client/cli/ping.py
@@ -16,6 +16,7 @@ import click
 from reana_client.cli.utils import add_access_token_options, check_connection
 from reana_client.config import JSON
 from reana_client.printer import display_message
+from reana_client.utils import build_cpu_quota_period_info
 from reana_client.version import __version__
 
 
@@ -110,9 +111,15 @@ def info(ctx, access_token: str, output_format: str):  # noqa: D301
     \t $ reana-client info
     """
     try:
-        from reana_client.api.client import info
+        from reana_client.api.client import get_user_quota, info
 
         response = info(access_token)
+        try:
+            response.update(build_cpu_quota_period_info(get_user_quota(access_token)))
+        except (KeyError, ValueError) as e:
+            logging.debug(
+                "Could not enrich cluster info with quota period details: %s", str(e)
+            )
         if output_format == JSON:
             display_message(json.dumps(response))
         else:
--- a/reana_client/cli/quotas.py
+++ b/reana_client/cli/quotas.py
@@ -22,6 +22,7 @@ from reana_client.cli.utils import (
     human_readable_or_raw_option,
 )
 from reana_client.printer import display_message
+from reana_client.utils import format_quota_period_window
 
 
 @click.group(help="Quota commands")
@@ -97,6 +98,22 @@ def quota_show(
                 quota[resource], human_readable_or_raw
             )
             health_color = REANA_RESOURCE_HEALTH_COLORS.get(health)
+            period_window = (
+                format_quota_period_window(quota[resource])
+                if resource == "cpu"
+                else None
+            )
+            if period_window:
+                # NOTE: This splice relies on get_quota_resource_usage()
+                # rendering percentage output as a trailing parenthetical.
+                if resource_usage.endswith(")"):
+                    resource_usage = resource_usage[:-1] + (
+                        f" in the period from {period_window})"
+                    )
+                else:
+                    resource_usage = (
+                        f"{resource_usage} in the period from {period_window}"
+                    )
             return click.secho(resource_usage, fg=health_color)
 
         result = (
@@ -105,6 +122,14 @@ def quota_show(
             and quota[resource].get(report).get("raw", 0) > 0
             else "No {}.".format(report)
         )
+        if (
+            human_readable_or_raw == "human_readable"
+            and resource == "cpu"
+            and not result.startswith("No ")
+        ):
+            period_window = format_quota_period_window(quota[resource])
+            if period_window:
+                result = f"{result} in the period from {period_window}"
         return display_message(result)
 
     except Exception as e:
--- a/reana_client/utils.py
+++ b/reana_client/utils.py
@@ -8,13 +8,14 @@
 """REANA client utils."""
 
 import base64
-from datetime import datetime
+import calendar
+from datetime import datetime, timezone
 import logging
 import os
 import pathlib
 import sys
 import traceback
-from typing import Dict, Optional
+from typing import Dict, Optional, Tuple, Union
 from uuid import UUID
 
 from reana_commons.utils import get_workflow_status_change_verb
@@ -23,6 +24,100 @@ from reana_commons.specification import load_reana_spec
 from reana_client.config import reana_yaml_valid_file_names
 from reana_client.printer import display_message
 from reana_client.validation.utils import validate_reana_spec
+
+
+# NOTE: Keep this month-boundary arithmetic in sync with reana_db/utils.py
+# (_add_months), reana-client-go's addMonths helper, and reana-ui's quota
+# period window calculation.
+def _add_months(dt: datetime, months: int) -> datetime:
+    """Add whole months to a datetime while preserving the time of day."""
+    year = dt.year + (dt.month - 1 + months) // 12
+    month = (dt.month - 1 + months) % 12 + 1
+    day = min(dt.day, calendar.monthrange(year, month)[1])
+    return dt.replace(year=year, month=month, day=day)
+
+
+def _format_date_label(dt: datetime) -> str:
+    """Format a datetime as an ISO calendar date."""
+    return dt.date().isoformat()
+
+
+def parse_server_datetime(value: Union[str, datetime, None]) -> Optional[datetime]:
+    """Parse a server datetime string into a naive UTC datetime."""
+    if not value:
+        return None
+
+    if isinstance(value, datetime):
+        parsed = value
+    elif isinstance(value, str):
+        try:
+            parsed = datetime.fromisoformat(
+                value.replace("Z", "+00:00") if value.endswith("Z") else value
+            )
+        except ValueError:
+            logging.debug("Could not parse server datetime %r as ISO 8601", value)
+            return None
+    else:
+        logging.debug(
+            "Unexpected server datetime type %s for value %r",
+            type(value).__name__,
+            value,
+        )
+        return None
+
+    if parsed.tzinfo:
+        parsed = parsed.astimezone(timezone.utc).replace(tzinfo=None)
+    return parsed
+
+
+def format_quota_period_window(resource: Dict) -> Optional[str]:
+    """Format the active quota accounting window for a quota resource."""
+    start_label, end_label = get_quota_period_date_range(resource)
+    if not start_label or not end_label:
+        return None
+
+    return f"{start_label} to {end_label}"
+
+
+def get_quota_period_date_range(resource: Dict) -> Tuple[Optional[str], Optional[str]]:
+    """Return ISO start/end dates for the active quota accounting window."""
+    period_months = resource.get("quota_period_months")
+    period_start_at = parse_server_datetime(resource.get("quota_period_start_at"))
+    if not period_months or not period_start_at:
+        return None, None
+
+    period_end_at = _add_months(period_start_at, period_months)
+    return _format_date_label(period_start_at), _format_date_label(period_end_at)
+
+
+def build_cpu_quota_period_info(
+    quota: Dict,
+) -> Dict[str, Dict[str, Union[str, int]]]:
+    """Build CLI info entries describing the current CPU quota period."""
+    cpu_quota = quota.get("cpu") if quota else None
+    if not cpu_quota:
+        return {}
+
+    period_months = cpu_quota.get("quota_period_months") or 0
+    period_info = {
+        "cpu_quota_period_months": {
+            "title": "CPU quota period in months",
+            "value": period_months,
+        }
+    }
+
+    start_label, end_label = get_quota_period_date_range(cpu_quota)
+    if start_label and end_label:
+        period_info["current_cpu_quota_period_start"] = {
+            "title": "Current CPU quota period start",
+            "value": start_label,
+        }
+        period_info["current_cpu_quota_period_end"] = {
+            "title": "Current CPU quota period end",
+            "value": end_label,
+        }
+
+    return period_info
 
 
 def workflow_uuid_or_name(ctx, param, value):
--- a/tests/test_cli_quotas.py
+++ b/tests/test_cli_quotas.py
@@ -34,6 +34,162 @@ def _build_quota_response(quota_period_months=3, quota_period_start_at=None):
     }
 
 
+def test_quota_show_displays_current_cpu_period():
+    """Test quota-show prints the active CPU quota window."""
+    env = {"REANA_SERVER_URL": "localhost"}
+    runner = CliRunner(env=env)
+    quota = _build_quota_response(
+        quota_period_start_at="2026-06-04T00:00:00Z",
+    )
+
+    with patch("reana_client.api.client.get_user_quota", return_value=quota):
+        result = runner.invoke(
+            cli,
+            ["quota-show", "-t", "000000", "--resource", "cpu", "--human-readable"],
+        )
+
+    assert result.exit_code == 0
+    assert (
+        "30s out of 16m 40s used (3% in the period from 2026-06-04 to 2026-09-04)"
+        in result.output
+    )
+
+
+def test_quota_show_report_usage_human_displays_cpu_period():
+    """Test quota-show usage report prints the active CPU quota window."""
+    env = {"REANA_SERVER_URL": "localhost"}
+    runner = CliRunner(env=env)
+    quota = _build_quota_response(
+        quota_period_start_at="2026-06-04T00:00:00Z",
+    )
+
+    with patch("reana_client.api.client.get_user_quota", return_value=quota):
+        result = runner.invoke(
+            cli,
+            [
+                "quota-show",
+                "-t",
+                "000000",
+                "--resource",
+                "cpu",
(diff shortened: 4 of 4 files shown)


reana_client/cli/quotas.py
```diff
@@ -97,6 +98,20 @@ def quota_show(
...
+                    resource_usage = resource_usage[:-1] + (
+                        f" in the period from {period_window})"
+                    )
+                else:
+                    resource_usage = (
+                        f"{resource_usage} in the period from {period_window}"
+                    )
             return click.secho(resource_usage, fg=health_color)
 
         result = (
```
tiborsimko, line 119: 🤖 **Show CPU period suffix when `-h` is used with `--report usage|limit` (LOW)**

The default mode (`quota-show -r cpu`) now displays the active CPU accounting window, but the explicit `--report` paths drop it. A user invoking `reana-client quota-show --resource cpu --report usage -h` sees only `30s`, which can be mis-interpreted as lifetime usage.

The natural rule is *raw output is scriptable, `-h` is for humans* — so the period belongs in `-h` output for both `--report usage` and `--report limit` on CPU (the limit applies to one period, so the window is part of what the limit means).

Proposed behaviour for the `--resource cpu` path on a user with periodic CPU accounting:

| Flags | Today | Proposed |
|---|---|---|
| `--report usage` | `30000` (raw) | unchanged |
| `--report usage -h` | `30s` | `30s in the period from 2026-05-15 to 2026-07-15` |
| `--report limit` | `6789000` (raw) | unchanged |
| `--report limit -h` | `1885h 50m` | `1885h 50m in the period from 2026-05-15 to 2026-07-15` |

For disk and for CPU users without periodic accounting, the suffix should not appear (today's behaviour). Raw output stays untouched in all cases so scripts keep working.

Suggested fix — extend the `--report` branch at `reana_client/cli/quotas.py:117`:

```python
result = (
    quota[resource][report][human_readable_or_raw]
    if quota[resource].get(report)
    and quota[resource].get(report).get("raw", 0) > 0
    else "No {}.".format(report)
)
if (
    human_readable_or_raw == "human_readable"
    and resource == "cpu"
    and not result.startswith("No ")
):
    window = format_quota_period_window(quota[resource])
    if window:
        result = f"{result} in the period from {window}"
return display_message(result)
```

New test cases for `tests/test_cli_quotas.py`:

- `--report usage -h` for a periodic CPU user → expect the suffix
- `--report usage -h` for a non-periodic CPU user → no suffix (regression guard)
- `--report usage` (no `-h`) for a periodic CPU user → still just the raw number (regression guard)
- `--report limit -h` for a periodic CPU user → expect the suffix
- `--report usage -h` for disk → no suffix

The same change should land in `reana-client-go` PR #180 for symmetry — equivalent comment posted there.

reana_client/utils.py
```diff
@@ -25,6 +26,84 @@
 from reana_client.validation.utils import validate_reana_spec
 
 
+def _add_months(dt: datetime, months: int) -> datetime:
```
tiborsimko, line 32: 🤖 **Period-end-date arithmetic is now duplicated in a 4th place (MEDIUM)**

This `_add_months` is byte-for-byte equivalent to the same helper in `reana_db/utils.py:_add_months` (already referenced by `reana_server` for periodic-window advancement). The Go counterpart in `reana-client-go` PR #180 adds the 3rd copy in `cmd/quota_period.go:addMonths`, and `reana-ui` PR #494 reaches the 4th via `moment.add(months, "months")` in `Quota.js`. Any future fix to the clamping rule (e.g. how Jan 31 + 1 month should land) has to be applied in all four implementations, with no compile-time or test-time signal if one is forgotten.

The duplication exists only because the server returns `quota_period_start_at` + `quota_period_months` but not `quota_period_end_at`, so every client has to recompute the window end locally. The cleanest fix is to compute it once on the server and ship it to every consumer:

1. **reana-db** — expose a `get_current_quota_period_end_at(...)` helper alongside the existing `get_current_quota_period_start_at()`, reusing the same `_add_months` primitive.
2. **reana-server** — extend `_serialize_user_quota` and `_get_quota_period` to populate a new `quota_period_end_at` field in the response.
3. **reana-commons** — declare `quota_period_end_at: { type: string, format: date-time, x-nullable: true }` in the OpenAPI spec next to `quota_period_start_at`, plus an updated example.
4. **All four clients** — read the server-provided end date and skip the local arithmetic entirely. `reana_client/utils.py:_add_months` and `_format_date_label` can go away; `format_quota_period_window` becomes a one-liner that formats two server-provided dates. Same simplification in `reana-client-go/cmd/quota_period.go` and `reana-ui/src/pages/profile/components/Quota.js`.

This is a coordinated change touching 5 repos, so it's reasonable to file as a follow-up issue rather than block this PR. But at minimum, if you keep the local arithmetic now, add a comment like

```python
# NOTE: keep in sync with reana_db/utils.py:_add_months,
# reana-client-go's addMonths, and reana-ui's moment.add(...).
```

so the duplication is visible to future maintainers.

If you do want to do it now: the server-side bits are small (the math primitive already exists in `reana_db/utils.py`), and removing the client copies is net negative LOC.

Cross-references: same observation applies to `reana-client-go` PR #180 (`cmd/quota_period.go:addMonths`) — flagging here as the architectural decision point, will mirror with a one-liner on the Go PR.

reana_client/cli/quotas.py
```diff
@@ -97,6 +98,20 @@ def quota_show(
                 quota[resource], human_readable_or_raw
             )
             health_color = REANA_RESOURCE_HEALTH_COLORS.get(health)
+            period_window = (
+                format_quota_period_window(quota[resource])
+                if resource == "cpu"
+                else None
+            )
+            if period_window:
+                if resource_usage.endswith(")"):
```
tiborsimko, line 109: 🤖 **Brittle string rewriting to splice the period into the usage line (LOW)**

This block decides where to inject the period suffix based on whether the assembled usage line happens to end with `)` (because a percentage was rendered). The logic is fragile in two ways:

1. It depends on a string-level property of the output of `get_quota_resource_usage()` rather than on the underlying data. If that function ever appends a trailing unit, a colour code, a status badge, or any other character after the closing paren, the `endswith(")")` check silently falls into the wrong branch and the period suffix lands in an awkward position — without any test failing, because all current fixtures end in `)`.
2. The same hack is duplicated in `reana-client-go` PR #180 (`cmd/quota_show.go:displayQuotaResourceUsage` uses `strings.HasSuffix(usageMsg, ")")`). Two implementations of the same fragile rewrite to maintain.

Cleaner: compose the trailing parenthetical from its components in a single place rather than rewriting the assembled string. Roughly, push period-window awareness *into* the message builder so it can include the suffix alongside the percentage from the start:

```python
resource_usage, health = get_quota_resource_usage(
    quota[resource],
    human_readable_or_raw,
    period_window=(
        format_quota_period_window(quota[resource])
        if resource == "cpu"
        else None
    ),
)
```

and have `get_quota_resource_usage` emit, in one pass:

- `"30s out of 16m 40s used (3%)"` when no period
- `"30s out of 16m 40s used (3% in the period from X to Y)"` when period is set and a percentage is rendered
- `"30s used in the period from X to Y"` when no percentage is rendered

This keeps all the format decisions in one function, eliminates the `endswith` reverse-engineering, and gives the same primitive a single home that the Go client can mirror cleanly.

Not a blocker — the current code is correct for today's outputs. Worth either refactoring now (small change, both clients) or filing as a follow-up; either way, please drop a short comment here noting the splice's dependence on the upstream format.

Cross-reference: same observation applies to `reana-client-go` PR #180 at `cmd/quota_show.go:displayQuotaResourceUsage`.

reana_client/utils.py
```diff
@@ -25,6 +26,84 @@
...
+
+    if isinstance(value, datetime):
+        parsed = value
+    elif isinstance(value, str):
+        try:
+            parsed = datetime.fromisoformat(
+                value.replace("Z", "+00:00") if value.endswith("Z") else value
+            )
+        except ValueError:
+            return None
```
tiborsimko, line 59: 🤖 **`parse_server_datetime` silently swallows parse failures (LOW)**

Both `ValueError` and the unexpected-type fall-through return `None` without logging, and the only caller (`format_quota_period_window`) treats `None` as "don't display a period" — so if the server ever starts returning a malformed `quota_period_start_at`, the period info simply disappears from `quota-show` / `info` with no diagnostic. The user gets the surprising silent behaviour and you get no breadcrumb when debugging.

This is the same shape as the bravado-validation-swallowing path we hit earlier in this release cycle (reana-server #768 BLOCKER about the naive-ISO `quota_period_start_at`) — a malformed timestamp produced wrong-looking client output with no surfaced error. A `logging.debug` line on each failure path is enough to make the next instance debuggable:

```python
def parse_server_datetime(value):
    """Parse a server datetime string into a naive UTC datetime."""
    if not value:
        return None

    if isinstance(value, datetime):
        parsed = value
    elif isinstance(value, str):
        try:
            parsed = datetime.fromisoformat(
                value.replace("Z", "+00:00") if value.endswith("Z") else value
            )
        except ValueError:
            logging.debug(
                "Could not parse server datetime %r as ISO 8601", value
            )
            return None
    else:
        logging.debug(
            "Unexpected server datetime type %s for value %r",
            type(value).__name__,
            value,
        )
        return None
    ...
```

`debug` is the right level — it shouldn't be noisy in normal use, but it lights up under `reana-client --loglevel DEBUG` which is what a user would run when their period info goes missing.

Cross-reference: same observation for `reana-client-go` PR #180 — `formatQuotaPeriodWindow` in `cmd/quota_period.go` returns `""` on `time.Parse` failure with no log line. Worth a matching `log.Debugf` there.

reana_client/cli/ping.py
```diff
@@ -110,9 +111,15 @@ def info(ctx, access_token: str, output_format: str):  # noqa: D301
     \t $ reana-client info
     """
     try:
-        from reana_client.api.client import info
+        from reana_client.api.client import get_user_quota, info
 
         response = info(access_token)
+        try:
+            response.update(build_cpu_quota_period_info(get_user_quota(access_token)))
+        except Exception as e:
```
tiborsimko: 🤖 **Broad `except Exception` masks programmer errors as well as API failures (LOW)**

The intent here is right — `info` should keep working on clusters where quotas are disabled, the user has no `cpu` resource, or the server is older than reana-server #768 — so swallowing API-shape errors is correct. But `Exception` also catches `AttributeError`, `TypeError`, `KeyError` raised from inside `build_cpu_quota_period_info` itself, which would be programmer bugs that you actually want surfaced (and not buried at `logging.debug` level, where most users will never see them).

It's also worth noting that `get_user_quota` in `reana_client/api/client.py` already catches `HTTPError` internally and returns `None`, and `build_cpu_quota_period_info` already handles a `None` / empty input by returning `{}`. So the happy path and the "server rejected my token" path never raise here — what `Exception` is catching is the *unhappy*, shape-mismatch path (a response missing `quota`, a malformed `quota_period_start_at` not caught by `parse_server_datetime`, an outright bug in the new helper).

Two options, in order of how invasive:

1. **Minimal fix — narrow the catch to the shape we actually expect to handle.** `KeyError` covers a missing `quota` field; `ValueError` covers any future strict parse path; we *don't* catch `AttributeError`/`TypeError`, which would now be a real traceback rather than a silent miss:

   ```python
   try:
       response.update(build_cpu_quota_period_info(get_user_quota(access_token)))
   except (KeyError, ValueError) as e:
       logging.debug(
           "Could not enrich cluster info with quota period details: %s", str(e)
       )
   ```

2. **Slightly nicer — let `get_user_quota` carry the burden alone, drop the local try/except.** It already swallows `HTTPError` and `build_cpu_quota_period_info` handles `None`. The only thing the local `try` covers beyond that is exotic bravado validation errors, which would also fire on the un-instrumented `info(access_token)` call two lines above — so this `try` doesn't meaningfully improve robustness over what's already there. Removing it makes the failure mode visible.

Either way, please keep the explicit `logging.debug` text so a user running `--loglevel DEBUG` can correlate "where did my period info go?" with a logged reason.

Cross-reference: same shape in `reana-client-go` PR #180 (`cmd/info.go`) — two `log.Debugf` swallows, one for `api.Operations.GetYou` and one for `parseQuotaInfo`. The `GetYou` one is fine (it's an API call). The `parseQuotaInfo` one masks programmer errors in the same way and should similarly narrow.

reana_client/utils.py
michaelbuchar: Added a comment, thanks

reana_client/cli/quotas.py
michaelbuchar: Added note as well

reana_client/utils.py
michaelbuchar: Fixed!

reana_client/cli/ping.py
michaelbuchar: Fixed!

reana_client/cli/quotas.py
michaelbuchar: Fixed

reana_client/utils.py
```diff
@@ -25,6 +26,84 @@
...
+        return {}
+
+    period_months = cpu_quota.get("quota_period_months")
+    period_label = (
+        "Disabled"
+        if not period_months
+        else f"Every {period_months} {'month' if period_months == 1 else 'months'}"
+    )
+    period_info = {
+        "cpu_quota_period": {
```
tiborsimko: 🤖 **`info` quota fields should be machine-parseable scalars, not English prose (LOW)**

Today's `info` output for the quota period is human-readable but not scriptable:

```
$ reana-client info | grep -i quota
CPU quota period: Every 2 months
Current CPU quota period: 2026-06-04 to 2026-09-04

$ reana-client info --json | jq -S '.cpu_quota_period'
{
  "title": "CPU quota period",
  "value": "Every 2 months"
}
```

A user wanting to alert when their period changes, confirm a deployment-wide rollout, or feed the value into another tool has to regex-parse `"Every N months"` / `"Disabled"` — and the `"Disabled"` branch is a different shape than the numeric branch, so the parser has to handle two cases. Likewise, `"2026-06-04 to 2026-09-04"` requires splitting on `" to "` to extract either endpoint.

Suggested fix — break the structured fields out of the prose and use scalar values, matching the convention `quota.default_cpu_limit: 0` already uses (`0 = disabled / unlimited`):

```
$ reana-client info | grep -i quota
CPU quota period in months: 2
Current CPU quota period start: 2026-06-04
Current CPU quota period end: 2026-09-04

$ reana-client info --json | jq -S '.cpu_quota_period_months.value, .current_cpu_quota_period_start.value, .current_cpu_quota_period_end.value'
2
"2026-06-04"
"2026-09-04"
```

For a user with no periodic CPU accounting, `cpu_quota_period_months.value` is `0`, and the start/end fields are omitted (or `null` in JSON if you prefer always-present keys). The plain-text rendering for that case becomes just `CPU quota period in months: 0`, with no start/end lines — uniform parse rule, no special case.

This maps cleanly to existing `info` precedent: each field has a `title` + a scalar `value`. The current `"Every 2 months"` is a derived display string, which doesn't belong in the structured payload.

A concrete diff for `build_cpu_quota_period_info` would be roughly:

```python
def build_cpu_quota_period_info(quota):
    cpu_quota = quota.get("cpu") if quota else None
    if not cpu_quota:
        return {}

    period_months = cpu_quota.get("quota_period_months") or 0
    period_info = {
        "cpu_quota_period_months": {
            "title": "CPU quota period in months",
            "value": period_months,
        }
    }
    period_start_at = parse_server_datetime(
        cpu_quota.get("quota_period_start_at")
    )
    if period_months and period_start_at:
        period_end_at = _add_months(period_start_at, period_months)
        period_info["current_cpu_quota_period_start"] = {
            "title": "Current CPU quota period start",
            "value": _format_date_label(period_start_at),
        }
        period_info["current_cpu_quota_period_end"] = {
            "title": "Current CPU quota period end",
            "value": _format_date_label(period_end_at),
        }
    return period_info
```

(If you decide to address the period-end-date-duplication MEDIUM comment by having the server return `quota_period_end_at`, the local arithmetic above drops out entirely — these two concerns compose nicely.)

For `quota-show`, the existing `... in the period from X to Y` rendering already exposes both dates, so no change needed there — only `info` needs this.

Cross-reference: same fix applies to `reana-client-go` PR #180 in `cmd/quota_period.go:buildCPUQuotaPeriodInfo` and `cmd/info.go:buildInfoOutputMap` — equivalent comment posted there.

michaelbuchar: Fixed!

tiborsimko (approved):

PR #777, opened 2026-06-19
feat(auth): add OIDC login and credential management

### LINKED ISSUE #977: Improve SSO and use JWT for external authentication
Improve SSO allow only one provider and use JWT for external authentication.

Sohn123:
Add reana-client login and logout so users authenticate through the
REANA server's OIDC issuer and reuse locally stored credentials. Use
browser login with an authorisation code, PKCE and state validation by
default, and provide --headless device login for machines without a
local browser. Allow a fixed callback port through
REANA_CLIENT_LOGIN_LOOPBACK_PORT so deployments can register a stable
loopback redirect URI.

Discover issuer endpoints and the CLI client identifier through the
selected REANA server so users need not configure the identity provider
separately. Require HTTPS for issuer URLs and reject redirects on
discovery, token and revocation requests. Validate device prompts,
polling intervals and reported token lifetimes so malformed issuer
responses produce actionable authentication errors.

Store credentials separately by normalised server URL in
~/.config/reana/reana-client.json, with REANA_CLIENT_CONFIG available
for an alternative location. Write the store atomically with restrictive
permissions and protect updates with process locks so concurrent
commands do not corrupt it. Reject symlinked lock files and bound lock
acquisition so contention produces a useful error instead of an
indefinite wait.

Refresh expiring access tokens automatically and serialise refresh
attempts for each server so concurrent commands reuse rotated
credentials. Recheck stored credential revisions before writing back and
conditionally clear rejected tokens so an older request cannot overwrite
a later login or erase another process's replacement credentials. Revoke
superseded tokens outside the credential-store lock so a slow issuer
cannot stall unrelated commands. Preserve a returned refresh token when
other response fields are malformed so token rotation does not discard
the only usable replacement.

Revoke the selected server's refresh token on logout when the issuer
advertises a revocation endpoint, then clear its local token material
while retaining server metadata. Report remote revocation failures as
warnings so users can distinguish local logout from confirmed issuer
revocation.

Send bearer authentication headers through generated API calls,
specification uploads and direct file transfers so credentials no longer
travel in API query parameters. Resolve the active server consistently
across these paths, giving REANA_SERVER_URL precedence over the saved
selection, and bound direct file-transfer connection and read timeouts.

Use stored credentials unless -t/--access-token or REANA_ACCESS_TOKEN
explicitly overrides them so existing automation can supply
issuer-issued JWTs. Reject environment tokens without the JWT three-part
format and point users to login. Redact access-token command parameters
in debug logs and retain underlying errors when credential resolution
fails unexpectedly.

Enable TLS certificate verification by default and provide
REANA_SERVER_CA_CERTS for deployments using a private CA. Apply
REANA_SERVER_TLS_VERIFY only to the selected server's HTTPS hostname and
effective port, including bundled Keycloak endpoints, so local
self-signed deployments work while other identity-provider origins
remain verified.

Accept 1/true/yes/on and 0/false/no/off for REANA_SERVER_TLS_VERIFY,
ignoring case and surrounding whitespace, so Python and Go clients
interpret the setting consistently. Keep verification enabled for unset
or empty values, reject other values unless a CA bundle takes
precedence, and warn once when verification is disabled.

Fetch dedicated notebook session secrets for launch URLs and explicitly
request them for interactive-workflow listings so notebook access no
longer reuses API tokens. Preserve successful session creation when
fetching its secret or inactivity-policy metadata fails, reporting a
warning instead of a failed open operation.

Document interactive login, credential-file permissions and CI reuse so
unattended jobs can use previously obtained credentials safely. Explain
refresh-token expiry and rotation limits, and use private temporary
files with cleanup in the CI example so persisted credentials are not
exposed on shared runners.

Raise the reana-commons minimum to 0.95.0a23 for bearer API contracts
and per-client TLS settings, require Click 8 for explicit option-source
detection and Requests 2.31 for compatible HTTP dependencies on
supported Python versions. Package all discovered client modules so
installed distributions include authentication and command
implementations.

BREAKING CHANGE: Use this client with the OIDC-compatible REANA server
and run reana-client login, or supply an issuer-issued JWT. Legacy
opaque REANA tokens no longer authenticate. TLS certificate verification
is enabled by default; configure REANA_SERVER_CA_CERTS for private
certificates. Replace static-token automation with credentials whose
refresh-token lifetime and rotation policy fit the job. For a legacy
0.95 alpha cluster, pin reana-client==0.95.0a5 and
reana-commons==0.95.0a22.

Closes reanahub/reana#977

Co-authored-by: Tibor Šimko <tibor.simko@cern.ch>

### DIFF
--- a/reana_client/api/client.py
+++ b/reana_client/api/client.py
@@ -35,6 +35,19 @@ current_rs_api_client = LocalProxy(
 )
 
 
+def _auth_request_options(access_token):
+    """Return bravado request options carrying bearer authentication."""
+    return {"headers": {"Authorization": "Bearer {}".format(access_token)}}
+
+
+def _auth_headers(access_token, extra_headers=None):
+    """Return requests headers carrying bearer authentication."""
+    headers = {"Authorization": "Bearer {}".format(access_token)}
+    if extra_headers:
+        headers.update(extra_headers)
+    return headers
+
+
 def ping(access_token):
     """Check if the REANA server is reachable and the user is correctly authenticated.
 
@@ -46,7 +59,7 @@ def ping(access_token):
     """
     try:
         response, http_response = current_rs_api_client.api.get_you(
-            access_token=access_token
+            _request_options=_auth_request_options(access_token),
         ).result()
         if http_response.status_code == 200:
             response["status"] = "Connected"
@@ -82,7 +95,7 @@ def get_user_quota(access_token):
     """
     try:
         response, http_response = current_rs_api_client.api.get_you(
-            access_token=access_token
+            _request_options=_auth_request_options(access_token),
         ).result()
         if http_response.status_code == 200:
             return response["quota"]
@@ -144,7 +157,7 @@ def get_workflows(
     """
     try:
         response, http_response = current_rs_api_client.api.get_workflows(
-            access_token=access_token,
+            _request_options=_auth_request_options(access_token),
             verbose=verbose,
             type=type,
             page=page,
@@ -190,7 +203,8 @@ def get_workflow_status(workflow, access_token):
     """
     try:
         response, http_response = current_rs_api_client.api.get_workflow_status(
-            workflow_id_or_name=workflow, access_token=access_token
+            _request_options=_auth_request_options(access_token),
+            workflow_id_or_name=workflow
         ).result()
         if http_response.status_code == 200:
             return response
@@ -224,12 +238,12 @@ def create_workflow(reana_specification, name, access_token):
              the ``workflow_id`` and ``workflow_name``, along with a ``message`` of success.
     """
     try:
-        response, http_response = current_rs_api_client.api.create_workflow(
+        (response, http_response) = current_rs_api_client.api.create_workflow(
+            _request_options=_auth_request_options(access_token),
             reana_specification=json.loads(
                 json.dumps(reana_specification, sort_keys=True)
             ),
             workflow_name=name,
-            access_token=access_token,
         ).result()
         if http_response.status_code == 201:
             return response
@@ -325,10 +339,10 @@ def create_workflow_from_json(
         if input_params is not None:
             reana_yaml["inputs"]["parameters"] = input_params
         validate_reana_yaml(reana_yaml)
-        response, http_response = current_rs_api_client.api.create_workflow(
+        (response, http_response) = current_rs_api_client.api.create_workflow(
+            _request_options=_auth_request_options(access_token),
             reana_specification=json.loads(json.dumps(reana_yaml, sort_keys=True)),
             workflow_name=name,
-            access_token=access_token,
         ).result()
         if http_response.status_code == 201:
             return response
@@ -364,9 +378,9 @@ def start_workflow(workflow, access_token, parameters):
              along with a ``message`` of success.
     """
     try:
-        response, http_response = current_rs_api_client.api.start_workflow(
+        (response, http_response) = current_rs_api_client.api.start_workflow(
+            _request_options=_auth_request_options(access_token),
             workflow_id_or_name=workflow,
-            access_token=access_token,
             parameters=parameters,
         ).result()
         if http_response.status_code == 200:
@@ -405,13 +419,15 @@ def upload_file(workflow, file_, file_name, access_token):
 
     try:
         endpoint = current_rs_api_client.api.upload_file.operation.path_name.format(
-            workflow_id_or_name=workflow
+            workflow_id_or_name=workflow,
         )
         http_response = requests.post(
             urljoin(get_api_url(), endpoint),
             data=file_,
-            params={"file_name": file_name, "access_token": access_token},
-            headers={"Content-Type": "application/octet-stream"},
+            params={"file_name": file_name},
+            headers=_auth_headers(
+                access_token, {"Content-Type": "application/octet-stream"}
+            ),
             verify=False,
         )
         if http_response.ok:
@@ -448,10 +464,10 @@ def get_workflow_logs(workflow, access_token, steps=None, page=None, size=None):
              contains the requested logs.
     """
     try:
-        response, http_response = current_rs_api_client.api.get_workflow_logs(
+        (response, http_response) = current_rs_api_client.api.get_workflow_logs(
+            _request_options=_auth_request_options(access_token),
             workflow_id_or_name=workflow,
             steps=steps,
-            access_token=access_token,
             page=page,
             size=size,
         ).result()
@@ -492,11 +508,13 @@ def download_file(workflow, file_name, access_token):
 
         logging.getLogger("urllib3").setLevel(logging.CRITICAL)
         endpoint = current_rs_api_client.api.download_file.operation.path_name.format(
-            workflow_id_or_name=workflow, file_name=file_name
+            workflow_id_or_name=workflow,
+            file_name=file_name,
         )
         http_response = requests.get(
             urljoin(get_api_url(), endpoint),
-            params={"file_name": file_name, "access_token": access_token},
+            params={"file_name": file_name},
+            headers=_auth_headers(access_token),
             verify=False,
         )
         if "Content-Disposition" in http_response.headers:
@@ -545,10 +563,10 @@ def delete_file(workflow, file_name, access_token):
              name of the file as key and info about the size as value.
     """
     try:
-        response, http_response = current_rs_api_client.api.delete_file(
+        (response, http_response) = current_rs_api_client.api.delete_file(
+            _request_options=_auth_request_options(access_token),
             workflow_id_or_name=workflow,
             file_name=file_name,
-            access_token=access_token,
         ).result()
         if http_response.status_code == 200 and (
             response["deleted"] or response["failed"]
@@ -592,9 +610,9 @@ def list_files(
                 ``last-modified`` keys.
     """
     try:
-        response, http_response = current_rs_api_client.api.get_files(
+        (response, http_response) = current_rs_api_client.api.get_files(
+            _request_options=_auth_request_options(access_token),
             workflow_id_or_name=workflow,
-            access_token=access_token,
             file_name=file_name,
             page=page,
             size=size,
@@ -711,7 +729,8 @@ def get_workflow_parameters(workflow, access_token):
     """
     try:
         response, http_response = current_rs_api_client.api.get_workflow_parameters(
-            workflow_id_or_name=workflow, access_token=access_token
+            _request_options=_auth_request_options(access_token),
+            workflow_id_or_name=workflow,
         ).result()
         if http_response.status_code == 200:
             return response
@@ -745,7 +764,8 @@ def get_workflow_specification(workflow, access_token):
     """
     try:
         response, http_response = current_rs_api_client.api.get_workflow_specification(
-            workflow_id_or_name=workflow, access_token=access_token
+            _request_options=_auth_request_options(access_token),
+            workflow_id_or_name=workflow,
         ).result()
         if http_response.status_code == 200:
             return response
@@ -792,10 +812,10 @@ def delete_workflow(workflow, all_runs: bool, workspace: bool, access_token: str
             "all_runs": all_runs,
             "workspace": workspace,
         }
-        response, http_response = current_rs_api_client.api.set_workflow_status(
+        (response, http_response) = current_rs_api_client.api.set_workflow_status(
+            _request_options=_auth_request_options(access_token),
             workflow_id_or_name=workflow,
             status="deleted",
-            access_token=access_token,
             parameters=parameters,
         ).result()
         if http_response.status_code == 200:
@@ -832,10 +852,10 @@ def stop_workflow(workflow, force_stop, access_token):
     """
     try:
         parameters = {"force_stop": force_stop}
-        response, http_response = current_rs_api_client.api.set_workflow_status(
+        (response, http_response) = current_rs_api_client.api.set_workflow_status(
+            _request_options=_auth_request_options(access_token),
             workflow_id_or_name=workflow,
             status="stop",
-            access_token=access_token,
             parameters=parameters,
         ).result()
         if http_response.status_code == 200:
@@ -875,12 +895,12 @@ def diff_workflows(workflow_id_a, workflow_id_b, brief, access_token, context_li
         are the actual lines that differ.
     """
     try:
-        response, http_response = current_rs_api_client.api.get_workflow_diff(
+        (response, http_response) = current_rs_api_client.api.get_workflow_diff(
+            _request_options=_auth_request_options(access_token),
             workflow_id_or_name_a=workflow_id_a,
             workflow_id_or_name_b=workflow_id_b,
             brief=brief,
             context_lines=context_lines,
-            access_token=access_token,
         ).result()
 
         if http_response.status_code == 200:
@@ -919,9 +939,9 @@ def open_interactive_session(
     :return: the relative path to the interactive session.
     """
     try:
-        response, http_response = current_rs_api_client.api.open_interactive_session(
+        (response, http_response) = current_rs_api_client.api.open_interactive_session(
+            _request_options=_auth_request_options(access_token),
             workflow_id_or_name=workflow,
-            access_token=access_token,
             interactive_session_type=interactive_session_type,
             interactive_session_configuration=interactive_session_configuration,
         ).result()
@@ -954,9 +974,9 @@ def close_interactive_session(workflow, access_token):
     :return: the relative path to the interactive session.
     """
     try:
-        response, http_response = current_rs_api_client.api.close_interactive_session(
+        (response, http_response) = current_rs_api_client.api.close_interactive_session(
+            _request_options=_auth_request_options(access_token),
             workflow_id_or_name=workflow,
-            access_token=access_token,
         ).result()
         if http_response.status_code == 200:
             return response
@@ -990,11 +1010,11 @@ def mv_files(source, target, workflow, access_token):
              and a ``message`` about the success of the operation.
     """
     try:
-        response, http_response = current_rs_api_client.api.move_files(
+        (response, http_response) = current_rs_api_client.api.move_files(
+            _request_options=_auth_request_options(access_token),
             source=source,
             target=target,
             workflow_id_or_name=workflow,
-            access_token=access_token,
         ).result()
 
         if http_response.status_code == 200:
@@ -1036,10 +1056,10 @@ def get_workflow_disk_usage(workflow, parameters, access_token):
              to a file, with the ``name`` and ``size`` keys.
     """
     try:
-        response, http_response = current_rs_api_client.api.get_workflow_disk_usage(
+        (response, http_response) = current_rs_api_client.api.get_workflow_disk_usage(
+            _request_options=_auth_request_options(access_token),
             workflow_id_or_name=workflow,
             parameters=parameters,
-            access_token=access_token,
         ).result()
         if http_response.status_code == 200:
             return response
@@ -1079,8 +1099,10 @@ def add_secrets(secrets, overwrite, access_token):
     :return: a dictionary containing the ``message`` key with a success message.
     """
     try:
-        response, http_response = current_rs_api_client.api.add_secrets(
-            secrets=secrets, access_token=access_token, overwrite=overwrite
+        (response, http_response) = current_rs_api_client.api.add_secrets(
+            _request_options=_auth_request_options(access_token),
+            secrets=secrets,
+            overwrite=overwrite,
         ).result()
         if http_response.status_code == 201:
             return response
@@ -1115,8 +1137,9 @@ def delete_secrets(secrets, access_token):
     :return: a list with the names of the deleted secrets.
     """
     try:
-        response, http_response = current_rs_api_client.api.delete_secrets(
-            secrets=secrets, access_token=access_token
+        (response, http_response) = current_rs_api_client.api.delete_secrets(
+            _request_options=_auth_request_options(access_token),
+            secrets=secrets,
         ).result()
         if http_response.status_code == 200:
             return response
@@ -1153,8 +1176,8 @@ def list_secrets(access_token):
              ``name`` and ``type`` keys.
     """
     try:
-        response, http_response = current_rs_api_client.api.get_secrets(
-            access_token=access_token
+        (response, http_response) = current_rs_api_client.api.get_secrets(
+            _request_options=_auth_request_options(access_token),
         ).result()
         if http_response.status_code == 200:
             return response
@@ -1189,8 +1212,8 @@ def info(access_token):
              and ``maximum_interactive_session_inactivity_period``.
     """
     try:
-        response, http_response = current_rs_api_client.api.info(
-            access_token=access_token
+        (response, http_response) = current_rs_api_client.api.info(
+            _request_options=_auth_request_options(access_token),
         ).result()
         if http_response.status_code == 200:
             return response
@@ -1230,8 +1253,8 @@ def get_workflow_retention_rules(workflow, access_token):
             response,
             http_response,
         ) = current_rs_api_client.api.get_workflow_retention_rules(
+            _request_options=_auth_request_options(access_token),
             workflow_id_or_name=workflow,
-            access_token=access_token,
         ).result()
         if http_response.status_code == 200:
             return response
@@ -1264,11 +1287,11 @@ def prune_workspace(workflow, include_inputs, include_outputs, access_token):
              a ``message`` key with the result of the operation.
     """
     try:
-        response, http_response = current_rs_api_client.api.prune_workspace(
+        (response, http_response) = current_rs_api_client.api.prune_workspace(
+            _request_options=_auth_request_options(access_token),
             workflow_id_or_name=workflow,
             include_inputs=include_inputs,
             include_outputs=include_outputs,
-            access_token=access_token,
         ).result()
 
         if http_response.status_code == 200:
@@ -1315,10 +1338,10 @@ def share_workflow(
         if valid_until:
             share_params["valid_until"] = valid_until
 
-        response, http_response = current_rs_api_client.api.share_workflow(
+        (response, http_response) = current_rs_api_client.api.share_workflow(
+            _request_options=_auth_request_options(access_token),
             workflow_id_or_name=workflow,
             share_details=share_params,
-            access_token=access_token,
         ).result()
 
         if http_response.status_code == 200:
@@ -1352,11 +1375,11 @@ def unshare_workflow(workflow, user_email_to_unshare_with, access_token):
         unshare_params = {
             "workflow_id_or_name": workflow,
             "user_email_to_unshare_with": user_email_to_unshare_with,
-            "access_token": access_token,
         }
 
-        response, http_response = current_rs_api_client.api.unshare_workflow(
-            **unshare_params
+        (response, http_response) = current_rs_api_client.api.unshare_workflow(
+            _request_options=_auth_request_options(access_token),
+            **unshare_params,
         ).result()
 
         if http_response.status_code == 200:
@@ -1390,8 +1413,8 @@ def get_workflow_sharing_status(workflow, access_token):
             response,
             http_response,
         ) = current_rs_api_client.api.get_workflow_share_status(
+            _request_options=_auth_request_options(access_token),
             workflow_id_or_name=workflow,
-            access_token=access_token,
         ).result()
 
         if http_response.status_code == 200:
@@ -1409,3 +1432,30 @@ def get_workflow_sharing_status(workflow, access_token):
             f"Message: {e.response.json()['message']}"
         )
         raise Exception(e.response.json()["message"])
+
+def get_openid_configuration():
+    """Get OpenID configuration.
+
+    :return: a dictionary containing the OpenID configuration.
+    """
+    try:
+        (response, http_response) = current_rs_api_client.api.get_openid_configuration().result()
+        if http_response.status_code == 200:
+            return response
+        else:
+            raise Exception(
+                "Expected status code 200 but replied with "
+                "{status_code}".format(status_code=http_response.status_code)
+            )
+
+    except HTTPError as e:
+        logging.debug(
+            "OpenID configuration could not be retrieved: "
+            "\nStatus: {}\nReason: {}\n"
+            "Message: {}".format(
+                e.response.status_code, e.response.reason, e.response.json()["message"]
+            )
+        )
+        raise Exception(e.response.json()["message"])
+    except Exception as e:
+        raise e
--- a/reana_client/auth/oidc.py
+++ b/reana_client/auth/oidc.py
@@ -5,16 +5,18 @@
 #
 # REANA is free software; you can redistribute it and/or modify it
 # under the terms of the MIT License; see LICENSE file for more details.
-"""OIDC device-flow and token refresh helpers."""
+"""OIDC login (loopback PKCE and device flow) and token refresh helpers."""
 
 import base64
 import hashlib
 import json
 import secrets
 import time
+import webbrowser
 from datetime import datetime, timedelta, timezone
-from typing import Callable, Dict, Optional
-from urllib.parse import urljoin
+from http.server import BaseHTTPRequestHandler, HTTPServer
+from typing import Callable, Dict, Optional, Tuple
+from urllib.parse import parse_qs, urlencode, urljoin, urlparse
 
 import requests
 
@@ -26,12 +28,23 @@ from reana_client.auth.storage import (
     upsert_server_entry,
 )
 
-
 DEFAULT_SCOPES = "openid profile email offline_access"
-FALLBACK_SCOPES = "openid profile email"
 EXPIRY_LEEWAY_SECONDS = 60
 DISCOVERY_PATH = "/api/.well-known/openid-configuration"
 PKCE_CODE_CHALLENGE_METHOD = "S256"
+
+LOOPBACK_HOST = "127.0.0.1"
+LOOPBACK_CALLBACK_PATH = "/callback"
+LOOPBACK_TIMEOUT_SECONDS = 300
+
+_CALLBACK_SUCCESS_HTML = (
+    b"<html><body><h1>REANA login complete.</h1>"
+    b"<p>You can close this tab and return to the terminal.</p></body></html>"
+)
+_CALLBACK_ERROR_HTML = (
+    b"<html><body><h1>REANA login failed.</h1>"
+    b"<p>Return to the terminal for details.</p></body></html>"
+)
 
 
 class AuthenticationError(Exception):
@@ -132,8 +145,8 @@ def discover(server_url: str) -> Dict:
     metadata = _response_json(response)
     required_fields = [
         "issuer",
+        "authorization_endpoint",
         "token_endpoint",
-        "device_authorization_endpoint",
         "reana_cli_client_id",
     ]
     missing_fields = [field for field in required_fields if not metadata.get(field)]
@@ -145,40 +158,9 @@ def discover(server_url: str) -> Dict:
     return metadata
 
 
-def _start_device_authorization(metadata: Dict, scopes: str, pkce: Dict) -> Dict:
-    """Start OIDC device authorization flow."""
-    response = requests.post(
-        metadata["device_authorization_endpoint"],
-        data={
-            "client_id": metadata["reana_cli_client_id"],
-            "scope": scopes,
-            "code_challenge": pkce["code_challenge"],
-            "code_challenge_method": pkce["code_challenge_method"],
-        },
-        timeout=30,
-        verify=False,
-    )
-    payload = _response_json(response)
-    if response.ok:
-        return payload
-    if payload.get("error") == "invalid_scope" and scopes == DEFAULT_SCOPES:
-        return _start_device_authorization(metadata, FALLBACK_SCOPES, pkce)
-    raise AuthenticationError(
-        "Could not start device login: "
-        f"{payload.get('error_description') or payload.get('error') or response.text}"
-    )
-
-
-def _offline_access_not_allowed(payload: Dict) -> bool:
-    """Return whether token response refused offline access."""
-    error = (payload.get("error") or "").lower()
-    description = (payload.get("error_description") or "").lower()
-    return "offline" in description and (
-        "not allowed" in description or error in {"invalid_grant", "invalid_scope"}
-    )
-
-
-def _store_token_response(server_url: str, metadata: Dict, token_response: Dict) -> Dict:
+def _store_token_response(
+    server_url: str, metadata: Dict, token_response: Dict
+) -> Dict:
     """Persist token response for a server."""
     refresh_token = token_response.get("refresh_token")
     if not refresh_token:
@@ -190,7 +172,8 @@ def _store_token_response(server_url: str, metadata: Dict, token_response: Dict)
         "issuer": metadata["issuer"],
         "client_id": metadata["reana_cli_client_id"],
         "token_endpoint": metadata["token_endpoint"],
-        "device_authorization_endpoint": metadata["device_authorization_endpoint"],
+        "authorization_endpoint": metadata.get("authorization_endpoint"),
+        "device_authorization_endpoint": metadata.get("device_authorization_endpoint"),
         "revocation_endpoint": metadata.get("revocation_endpoint"),
         "access_token": token_response.get("access_token"),
         "access_token_expires_at": _token_expires_at(token_response),
@@ -200,58 +183,210 @@ def _store_token_response(server_url: str, metadata: Dict, token_response: Dict)
     return upsert_server_entry(server_url, entry)
 
 
+def _build_authorization_url(
+    metadata: Dict, scopes: str, pkce: Dict, state: str, redirect_uri: str
+) -> str:
+    """Build the OIDC authorization endpoint URL for the loopback flow."""
+    params = {
+        "response_type": "code",
+        "client_id": metadata["reana_cli_client_id"],
+        "redirect_uri": redirect_uri,
+        "scope": scopes,
+        "state": state,
+        "code_challenge": pkce["code_challenge"],
+        "code_challenge_method": pkce["code_challenge_method"],
+    }
+    return metadata["authorization_endpoint"] + "?" + urlencode(params)
+
+
+class _CallbackHandler(BaseHTTPRequestHandler):
+    """Capture the single authorization-code redirect on the loopback server."""
+
+    def do_GET(self):  # noqa: N802
+        """Record callback query parameters and acknowledge the browser."""
+        parsed = urlparse(self.path)
+        if parsed.path != LOOPBACK_CALLBACK_PATH:
+            self.send_response(404)
+            self.end_headers()
+            return
+        query = {key: values[0] for key, values in parse_qs(parsed.query).items()}
+        self.server.callback_query = query
+        body = _CALLBACK_SUCCESS_HTML if "code" in query else _CALLBACK_ERROR_HTML
+        self.send_response(200)
+        self.send_header("Content-Type", "text/html; charset=utf-8")
+        self.end_headers()
+        self.wfile.write(body)
+
+    def log_message(self, *args):
+        """Silence the default stderr request logging."""
+
+
+def _start_callback_server() -> Tuple[HTTPServer, str]:
+    """Start a loopback HTTP server and return it with its redirect URI."""
+    httpd = HTTPServer((LOOPBACK_HOST, 0), _CallbackHandler)
+    httpd.callback_query = None
+    port = httpd.server_address[1]
+    redirect_uri = f"http://{LOOPBACK_HOST}:{port}{LOOPBACK_CALLBACK_PATH}"
+    return httpd, redirect_uri
+
+
+def _wait_for_callback(httpd: HTTPServer, timeout: int) -> Optional[Dict]:
+    """Serve requests until the authorization callback arrives or times out."""
+    deadline = time.monotonic() + timeout
+    httpd.callback_query = None
+    while httpd.callback_query is None:
+        remaining = deadline - time.monotonic()
+        if remaining <= 0:
+            break
+        httpd.timeout = remaining
+        httpd.handle_request()
+    return httpd.callback_query
+
+
+def _exchange_authorization_code(
+    metadata: Dict, code: str, pkce: Dict, redirect_uri: str
+) -> Dict:
+    """Exchange an authorization code for tokens using the PKCE verifier."""
+    response = requests.post(
+        metadata["token_endpoint"],
+        data={
+            "grant_type": "authorization_code",
+            "client_id": metadata["reana_cli_client_id"],
+            "code": code,
+            "redirect_uri": redirect_uri,
+            "code_verifier": pkce["code_verifier"],
+        },
+        timeout=30,
+        verify=False,
+    )
+    payload = _response_json(response)
+    if not response.ok:
+        raise AuthenticationError(
+            "Browser login failed: "
+            f"{payload.get('error_description') or payload.get('error') or response.text}"
+        )
+    return payload
+
+
+def login_with_loopback(
+    server_url: str,
+    display_url: Callable[[str], None],
+    open_browser: Callable[[str], bool] = webbrowser.open,
+    timeout: int = LOOPBACK_TIMEOUT_SECONDS,
+) -> Dict:
+    """Perform the loopback authorization-code + PKCE flow and store credentials."""
+    normalized_url = normalize_server_url(server_url)
+    metadata = discover(normalized_url)
+    pkce = generate_pkce_pair()
+    state = secrets.token_urlsafe(32)
+
+    httpd, redirect_uri = _start_callback_server()
+    try:
+        authorization_url = _build_authorization_url(
+            metadata, DEFAULT_SCOPES, pkce, state, redirect_uri
+        )
+        display_url(authorization_url)
+        try:
+            open_browser(authorization_url)
+        except Exception:
+            pass
+        query = _wait_for_callback(httpd, timeout)
+    finally:
+        httpd.server_close()
+
+    if query is None:
+        raise AuthenticationError("Browser login timed out. Please run login again.")
+    if query.get("error"):
+        raise AuthenticationError(
+            "Browser login failed: "
+            f"{query.get('error_description') or query.get('error')}"
+        )
+    returned_state = query.get("state")
+    if not returned_state or not secrets.compare_digest(returned_state, state):
+        raise AuthenticationError(
+            "Browser login failed: state parameter mismatch (possible CSRF)."
+        )
+    code = query.get("code")
+    if not code:
+        raise AuthenticationError(
+            "Browser login failed: no authorization code was returned."
+        )
+
+    token_response = _exchange_authorization_code(metadata, code, pkce, redirect_uri)
+    return _store_token_response(normalized_url, metadata, token_response)
+
+
+def _start_device_authorization(metadata: Dict, pkce: Dict) -> Dict:
+    """Start OIDC device authorization flow."""
+    response = requests.post(
+        metadata["device_authorization_endpoint"],
+        data={
+            "client_id": metadata["reana_cli_client_id"],
+            "scope": DEFAULT_SCOPES,
+            "code_challenge": pkce["code_challenge"],
+            "code_challenge_method": pkce["code_challenge_method"],
+        },
+        timeout=30,
+        verify=False,
+    )
+    payload = _response_json(response)
+    if response.ok:
+        return payload
+    raise AuthenticationError(
+        "Could not start device login: "
+        f"{payload.get('error_description') or payload.get('error') or response.text}"
+    )
+
+
 def login_with_device_flow(
     server_url: str,
     display_callback: Callable[[Dict], None],
     sleep: Callable[[int], None] = time.sleep,
 ) -> Dict:
-    """Perform OIDC device flow and store resulting credentials."""
+    """Perform OIDC device flow (headless fallback) and store credentials."""
     normalized_url = normalize_server_url(server_url)
     metadata = discover(normalized_url)
-    scopes_to_try = [DEFAULT_SCOPES, FALLBACK_SCOPES]
-    for scopes in scopes_to_try:
-        pkce = generate_pkce_pair()
-        device_response = _start_device_authorization(metadata, scopes, pkce)
-        display_callback(device_response)
+    if not metadata.get("device_authorization_endpoint"):
+        raise AuthenticationError(
+            "This REANA server does not advertise a device authorization endpoint."
+        )
+    pkce = generate_pkce_pair()
+    device_response = _start_device_authorization(metadata, pkce)
+    display_callback(device_response)
 
-        interval = int(device_response.get("interval", 5))
-        device_code = device_response["device_code"]
-        while True:
-            sleep(interval)
-            response = requests.post(
-                metadata["token_endpoint"],
-                data={
-                    "grant_type": "urn:ietf:params:oauth:grant-type:device_code",
-                    "device_code": device_code,
-                    "client_id": metadata["reana_cli_client_id"],
-                    "code_verifier": pkce["code_verifier"],
-                },
-                timeout=30,
-                verify=False,
-            )
-            payload = _response_json(response)
-            if response.ok:
-                return _store_token_response(normalized_url, metadata, payload)
+    interval = int(device_response.get("interval", 5))
+    device_code = device_response["device_code"]
+    while True:
+        sleep(interval)
+        response = requests.post(
+            metadata["token_endpoint"],
+            data={
+                "grant_type": "urn:ietf:params:oauth:grant-type:device_code",
+                "device_code": device_code,
+                "client_id": metadata["reana_cli_client_id"],
+                "code_verifier": pkce["code_verifier"],
+            },
+            timeout=30,
+            verify=False,
+        )
+        payload = _response_json(response)
+        if response.ok:
+            return _store_token_response(normalized_url, metadata, payload)
 
-            error = payload.get("error")
-            if error == "authorization_pending":
-                continue
-            if error == "slow_down":
-                interval += 5
-                continue
-            if error == "expired_token":
-                raise AuthenticationError(
-                    "Device login expired. Please run login again."
-                )
-            if error == "access_denied":
-                raise AuthenticationError("Device login was denied.")
-            if scopes == DEFAULT_SCOPES and _offline_access_not_allowed(payload):
-                break
-            raise AuthenticationError(
-                "Device login failed: "
-                f"{payload.get('error_description') or error or response.text}"
-            )
-    raise AuthenticationError("Device login failed: offline access is not allowed.")
+        error = payload.get("error")
+        if error == "authorization_pending":
+            continue
+        if error == "slow_down":
+            interval += 5
+            continue
+        if error == "expired_token":
+            raise AuthenticationError("Device login expired. Please run login again.")
+        if error == "access_denied":
+            raise AuthenticationError("Device login was denied.")
+        raise AuthenticationError(
+            "Device login failed: "
+            f"{payload.get('error_description') or error or response.text}"
+        )
 
 
 def _access_token_valid(server_entry: Dict) -> bool:
@@ -290,9 +425,10 @@ def refresh_credentials(server_url: str, server_entry: Optional[Dict] = None) ->
         "issuer": server_entry["issuer"],
         "reana_cli_client_id": server_entry["client_id"],
         "token_endpoint": server_entry["token_endpoint"],
-        "device_authorization_endpoint": server_entry[
+        "authorization_endpoint": server_entry.get("authorization_endpoint"),
+        "device_authorization_endpoint": server_entry.get(
             "device_authorization_endpoint"
-        ],
+        ),
         "revocation_endpoint": server_entry.get("revocation_endpoint"),
     }
     if "refresh_token" not in payload:
--- a/reana_client/auth/oidc.py
+++ b/reana_client/auth/oidc.py
@@ -1,0 +1,343 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of REANA.
+# Copyright (C) 2026 CERN.
+#
+# REANA is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+"""OIDC device-flow and token refresh helpers."""
+
+import base64
+import hashlib
+import json
+import secrets
+import time
+from datetime import datetime, timedelta, timezone
+from typing import Callable, Dict, Optional
+from urllib.parse import urljoin
+
+import requests
+
+from reana_client.auth.storage import (
+    clear_token_material,
+    get_active_server,
+    get_server_entry,
+    normalize_server_url,
+    upsert_server_entry,
+)
+
+
+DEFAULT_SCOPES = "openid profile email offline_access"
+FALLBACK_SCOPES = "openid profile email"
+EXPIRY_LEEWAY_SECONDS = 60
+DISCOVERY_PATH = "/api/.well-known/openid-configuration"
+PKCE_CODE_CHALLENGE_METHOD = "S256"
+
+
+class AuthenticationError(Exception):
+    """Authentication failure visible to CLI users."""
+
+
+def utcnow() -> datetime:
+    """Return timezone-aware current UTC time."""
+    return datetime.now(timezone.utc)
+
+
+def parse_timestamp(value: Optional[str]) -> Optional[datetime]:
+    """Parse ISO timestamp stored in credential file."""
+    if not value:
+        return None
+    if value.endswith("Z"):
+        value = value[:-1] + "+00:00"
+    return datetime.fromisoformat(value)
+
+
+def format_timestamp(value: datetime) -> str:
+    """Format timestamp for credential file."""
+    return value.astimezone(timezone.utc).isoformat().replace("+00:00", "Z")
+
+
+def _decode_jwt_exp(token: Optional[str]) -> Optional[datetime]:
+    """Decode JWT exp claim without validating the token."""
+    if not token or token.count(".") < 2:
+        return None
+    payload = token.split(".")[1]
+    payload += "=" * (-len(payload) % 4)
+    try:
+        claims = json.loads(base64.urlsafe_b64decode(payload.encode("ascii")))
+        exp = claims.get("exp")
+        return datetime.fromtimestamp(exp, timezone.utc) if exp else None
+    except Exception:
+        return None
+
+
+def _token_expires_at(token_response: Dict) -> Optional[str]:
+    """Return access token expiry timestamp from token response."""
+    expires_in = token_response.get("expires_in")
+    if expires_in is not None:
+        return format_timestamp(utcnow() + timedelta(seconds=int(expires_in)))
+    jwt_exp = _decode_jwt_exp(token_response.get("access_token"))
+    return format_timestamp(jwt_exp) if jwt_exp else None
+
+
+def _refresh_token_expires_at(token_response: Dict) -> Optional[str]:
+    """Return refresh token expiry timestamp if issuer exposes it."""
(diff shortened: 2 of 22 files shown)


reana_client/auth/storage.py
```diff
@@ -0,0 +1,130 @@
...
+
+def get_server_entry(server_url: str, config: Optional[Dict] = None) -> Dict:
+    """Return credential store entry for a server."""
+    config = config or load_config()
+    normalized_url = normalize_server_url(server_url)
+    return config.get("servers", {}).get(normalized_url, {})
+
+
+def upsert_server_entry(server_url: str, entry: Dict) -> Dict:
+    """Update credential store entry for a server and mark it active."""
```
tiborsimko, line 193: **🤖 PR777-01: CLI refresh-token rotation is not serialised (MEDIUM)**

Atomic file replacement prevents partial writes but not read-modify-write races. Two CLI processes can refresh the same rotating token concurrently: the winner stores the new token, then the loser receives `invalid_grant` and calls `clear_token_material()`, removing the winner's newly stored credentials and forcing an unnecessary login.

The refresh and clearing path is in `reana_client/auth/oidc.py:402-448`; the unlocked read-modify-write helpers are in `reana_client/auth/storage.py:110-130`.

Suggested fix — use an inter-process lock per credential file/server around reload, expiry recheck, refresh, and write. After acquiring the lock, re-read the entry and reuse a token already refreshed by another process. Apply the same lock to every credential-store read-modify-write operation.

reana_client/auth/storage.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+def normalize_server_url(server_url: str) -> str:
+    """Normalize server URL used as credential store key."""
+    if not server_url:
+        raise ValueError("REANA server URL is not set")
+    server_url = server_url.strip()
+    if "://" not in server_url:
+        server_url = f"http://{server_url}"
+    parsed = urlparse(server_url)
```
tiborsimko, line 111: **🤖 PR777-02: Missing URL schemes silently downgrade the client to HTTP (MEDIUM)**

`normalize_server_url()` prefixes a scheme-less hostname with `http://`. Subsequent API requests can then send access JWTs over cleartext without warning, and initial discovery is also exposed to network tampering. `verify=True` provides no protection for an HTTP URL.

The insecure default is in `reana_client/auth/storage.py:25-45`.

Suggested fix — default scheme-less server names to HTTPS. Permit HTTP only for explicit loopback URLs, or behind a clearly named insecure-development option.

reana_client/cli/workflow.py
```diff
@@ -328,10 +329,15 @@ def workflows_list(  # noqa: C901
             workflow["run_number"] = run_number
             workflow["duration"] = get_workflow_duration(workflow)
             if type == "interactive":
+                from reana_client.api.client import get_interactive_session_secret
+
+                session_secret = get_interactive_session_secret(
```
tiborsimko: **🤖 PR777-04: Interactive-session listing makes N+1 secret requests and one failure aborts it (MEDIUM)**

`reana-client list --type interactive` now fetches the per-session notebook secret separately for every workflow inside the rendering loop (`reana_client/cli/workflow.py:331-341`):

```python
session_secret = get_interactive_session_secret(
    workflow["id"], access_token
)["session_secret"]
```

N sessions therefore add N serial HTTP round trips after the list request. More importantly, the call sits inside the command's outer `try`, so one owner-only 403 while listing a workflow shared with the caller, one transient failure, or one missing secret aborts the entire listing instead of degrading only that row.

Suggested fix — isolate failures per row and render the secret-less session URI when a secret is unavailable. Preferably, return an opaque owner-only launch URL or the authorised session secret with each applicable list item so the command does not need an N+1 request pattern at all.

reana_client/auth/oidc.py
```diff
@@ -0,0 +1,491 @@
...
+        raise AuthenticationError(
+            "This REANA server does not advertise a device authorization endpoint."
+        )
+    pkce = generate_pkce_pair()
+    device_response = _start_device_authorization(metadata, pkce)
+    display_callback(device_response)
+
+    interval = int(device_response.get("interval", 5))
+    device_code = device_response["device_code"]
+    while True:
```
tiborsimko, line 458: **🤖 PR777-05: Device-flow polling has no client-side deadline (LOW)**

The device-authorisation polling loop is `while True` (`reana_client/auth/oidc.py:358`) and terminates only on the terminal responses the identity provider chooses to send (`authorization_pending`/`slow_down` continue; `expired_token` and other errors stop, `reana_client/auth/oidc.py:376-381`). The device-authorisation response's `expires_in` field, which the provider supplies precisely so clients can bound their own polling, is not used as a local deadline.

A misbehaving or misconfigured provider that keeps answering `authorization_pending` therefore pins `reana-client login` in an infinite poll, and the user has no indication that the code they were shown expired long ago.

Suggested fix — record `expires_in` from the device-authorisation response and exit the loop with a clear "code expired, please retry" message once the deadline passes, regardless of what the token endpoint keeps returning.

reana_client/api/client.py
```diff
@@ -945,6 +965,29 @@ def open_interactive_session(
         raise e
 
 
+def get_interactive_session_secret(workflow, access_token):
+    """Return the per-session notebook access secret for a workflow owner."""
+    from reana_client.utils import get_api_url
+
+    endpoint = "/api/workflows/{}/interactive-session-secret".format(workflow)
```
tiborsimko: **🤖 PR777-06: The new session-secret helper bypasses the generated API transport (LOW)**

`get_interactive_session_secret()` manually constructs and executes a raw request even though the operation is part of the OpenAPI contract (`reana_client/api/client.py:968-988`). This bespoke path interpolates `workflow` without URL-encoding the path segment, has no request timeout, and unconditionally parses non-200 responses as JSON:

```python
endpoint = "/api/workflows/{}/interactive-session-secret".format(workflow)
...
raise Exception(http_response.json().get("message"))
```

A valid workflow name containing URL-reserved characters can therefore target the wrong path; a stalled ingress can block the CLI indefinitely; and an HTML proxy error masks the original failure with `JSONDecodeError`. These are all avoided elsewhere by keeping operation construction and HTTP error handling in one transport layer.

Suggested fix — call the generated `get_interactive_session_secret` operation with `_auth_request_options(access_token)` and `workflow_id_or_name=workflow`. If a direct request is retained, quote the path segment, set a timeout, and tolerate non-JSON error bodies explicitly.

reana_client/auth/oidc.py
```diff
@@ -0,0 +1,516 @@
...
+            data={
+                "grant_type": "urn:ietf:params:oauth:grant-type:device_code",
+                "device_code": device_code,
+                "client_id": metadata["reana_cli_client_id"],
+                "code_verifier": pkce["code_verifier"],
+            },
+            timeout=max(0.1, min(30, remaining)),
+            verify=tls_verify(),
+        )
+        if monotonic() >= deadline:
```
tiborsimko: **🤖 PR777-07: Successful device-flow token response is discarded after the local deadline (LOW)**

After the token POST, the new deadline check runs before checking `response.ok` (`reana_client/auth/oidc.py:386-387`):

```python
if monotonic() >= deadline:
    raise AuthenticationError("Device login expired. Please run login again.")
```

If the user authorises in the final seconds and the token request completes just past the deadline, the IdP has already issued valid access and refresh tokens, but the client throws them away — unstored and unrevoked — and forces a full re-login. The dedicated test `test_device_flow_rejects_success_returned_after_local_expiry` shows this is intentional, but there is no security benefit: a genuinely expired device code is rejected by the IdP with `expired_token` anyway, so the local check only needs to bound *polling*, not veto server-accepted grants. Meanwhile a live refresh token has been minted and abandoned.

Suggested fix — move the post-request deadline check after the `response.ok` early return (or gate it on `not response.ok`), so a successful exchange is always stored; keep the pre-request deadline checks as the polling bound. Alternatively, if the rejection is kept, revoke the discarded refresh token.

reana_client/config.py
```diff
@@ -7,13 +7,53 @@
...
+
+    ``REANA_SERVER_CA_CERTS`` takes precedence over ``REANA_INSECURE``. When
+    neither is set the standard ``REQUESTS_CA_BUNDLE`` environment variable is
+    still honoured by ``requests``.
+    """
+    ca_certs = os.getenv(CA_CERTS_ENV)
+    if ca_certs:
+        return ca_certs
+    if os.getenv(INSECURE_ENV, "").strip().lower() in _INSECURE_TRUE_VALUES:
+        urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
```
tiborsimko, line 55: **🤖 PR777-08: REANA_INSECURE disables TLS verification with zero user-visible signal (LOW)**

When `REANA_INSECURE` is truthy, `tls_verify()` returns `False` *and* suppresses the only runtime indication that certificate verification is off (`reana_client/config.py:45-46`):

```python
urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
```

A stale `export REANA_INSECURE=true` in a shell profile then silently disables TLS verification for every server in the credential store, including production, in a CLI whose whole purpose is handling bearer tokens. The escape hatch itself is fine — it is the requested single insecure setting from PR777-02 — but it should not be invisible.

Suggested fix — emit a one-time warning instead of relying on the suppressed urllib3 warning, e.g. `display_message("TLS verification is disabled (REANA_INSECURE is set).", msg_type="warning")` or a `logging.warning` guarded by a module-level flag.

reana_client/auth/storage.py
tiborsimko: Addressed in commit `d19a835`, kept as a separate fixup on top of the original commit for review clarity. This fixup commit can be squashed into the original commit before merge, preserving the first author with `Co-authored-by` for the fixup.

reana_client/auth/storage.py
tiborsimko: Addressed in commit `d19a835`, kept as a separate fixup on top of the original commit for review clarity. This fixup commit can be squashed into the original commit before merge, preserving the first author with `Co-authored-by` for the fixup.

reana_client/cli/workflow.py
tiborsimko: Addressed in commit `d19a835`, kept as a separate fixup on top of the original commit for review clarity. This fixup commit can be squashed into the original commit before merge, preserving the first author with `Co-authored-by` for the fixup.

reana_client/auth/oidc.py
tiborsimko: Addressed in commit `d19a835`, kept as a separate fixup on top of the original commit for review clarity. This fixup commit can be squashed into the original commit before merge, preserving the first author with `Co-authored-by` for the fixup.

reana_client/api/client.py
tiborsimko: Addressed in commit `d19a835`, kept as a separate fixup on top of the original commit for review clarity. This fixup commit can be squashed into the original commit before merge, preserving the first author with `Co-authored-by` for the fixup.

reana_client/auth/storage.py
```diff
@@ -0,0 +1,212 @@
...
+
+def normalize_server_url(server_url: str) -> str:
+    """Normalize server URL used as credential store key."""
+    if not server_url:
+        raise ValueError("REANA server URL is not set")
+    server_url = server_url.strip()
+    if "://" not in server_url:
+        server_url = f"https://{server_url}"
+    parsed = urlparse(server_url)
+    if not parsed.scheme or not parsed.netloc:
```
tiborsimko, line 112: **🤖 PR777-11: Authentication accepts cleartext and non-HTTP server URLs (MEDIUM)**

`normalize_server_url()` adds `https://` only when a scheme is missing. An
explicit `http://` URL is accepted without warning, and any other scheme with
a netloc also passes validation. The discovery request, authorization-code
exchange, device polling, refresh, revocation, and file-transfer calls then use
those URLs directly. With HTTP, access and long-lived refresh credentials can
cross the network in cleartext; with another scheme, Requests fails later with
an opaque `InvalidSchema` error.

This is inconsistent with the new TLS policy: `REANA_INSECURE` emits a warning
for disabled certificate verification, while the more dangerous cleartext
mode is silent. The issuer metadata returned by REANA Server is currently
validated server-side, but the client should still reject a relayed
credential-bearing endpoint that violates its own transport policy rather
than relying on that implementation detail.

Suggested fix — require `https` for REANA Server and all non-loopback OIDC
endpoints. `REANA_INSECURE` should continue to mean HTTPS without certificate
verification, not permission to use HTTP. If a loopback-only HTTP exception is
truly needed, make it narrow and explicit. Add tests for missing scheme,
HTTPS, rejected HTTP/FTP, and rejected cleartext discovered endpoints.

reana_client/api/client.py
```diff
@@ -405,14 +428,16 @@ def upload_file(workflow, file_, file_name, access_token):
 
     try:
         endpoint = current_rs_api_client.api.upload_file.operation.path_name.format(
-            workflow_id_or_name=workflow
+            workflow_id_or_name=workflow,
         )
         http_response = requests.post(
```
tiborsimko, line 446: **🤖 PR777-12: Raw upload and download requests have no timeout (LOW)**

`upload_file()` and `download_file()` call `requests.post()`/`requests.get()`
without `timeout=`. A stalled connection or response can therefore hang a CLI
process indefinitely. The upload handler already catches
`requests.exceptions.Timeout`, but that branch is unreachable without a
Requests timeout, which makes the apparent recovery misleading.

These two calls bypass the generated Bravado transport because they stream
file content, so the timeouts added to OIDC requests do not protect them.

Suggested fix — define one shared `(connect, read)` timeout policy appropriate
for streaming transfers and pass it to both requests. Keep the read timeout
long enough for large files but finite/configurable, make the existing timeout
message reachable, and add focused timeout tests for upload and download.

reana_client/cli/utils.py
```diff
@@ -35,16 +37,14 @@
...
 
     @click.option(
         "-t",
         "--access-token",
-        default=os.getenv("REANA_ACCESS_TOKEN"),
-        callback=lambda ctx, _, access_token: access_token_check(
-            ctx, _, access_token, required
-        ),
-        help="Access token of the current user.",
+        envvar="REANA_ACCESS_TOKEN",
```
tiborsimko, line 47: **🤖 PR777-13: A stale opaque-token override silently defeats a valid OIDC login (LOW)**

The amended access-token decorator deliberately keeps `-t/--access-token` and
`REANA_ACCESS_TOKEN` as manual bearer-token overrides, with precedence over
the new credential store. Keeping an explicit JWT override can be useful for
automation, but the old environment variable commonly contains a legacy
opaque REANA token. After the no-grace-period cut-over, such a stale export
silently overrides a valid OIDC login and turns every command into a confusing
401; the help/error text continues to recommend the same variable without
explaining that opaque tokens are retired.

Suggested fix — retain the manual override if it is a supported JWT automation
interface, but fail early with an actionable migration error when its value is
not a signed-JWT shape accepted by the new server. Consider renaming the
environment variable to make the JWT-only contract unambiguous, and document
that it overrides stored OIDC credentials. Add a test proving a legacy opaque
value cannot silently shadow a valid stored login.

reana_client/api/client.py
```diff
@@ -30,9 +29,27 @@
 from reana_commons.validation.utils import validate_reana_yaml, validate_workflow_name
 from werkzeug.local import LocalProxy
 
-current_rs_api_client = LocalProxy(
-    partial(get_current_api_client, component="reana-server")
-)
+def _get_current_reana_server_api_client():
+    """Return a generated API client using the shared TLS policy."""
+    return get_current_api_client(
```
tiborsimko, line 45: **🤖 PR777-14: Stored authentication and generated API calls resolve different server URLs (MEDIUM)**

The new credential store records a normalised `active_server`, and auth/file
paths use `get_active_server()`. Generated API calls instead construct the
commons client, whose `_load_config_from_env()` reads raw
`REANA_SERVER_URL` directly and knows nothing about the stored active server.
This creates two concrete failures:

- `reana-client login --server-url https://reana.example.org` succeeds and
  stores credentials, but a later command with no exported environment fails
  in commons with `MissingAPIClientConfiguration`;
- `REANA_SERVER_URL=localhost` is normalised to `https://localhost` for login
  and raw transfers, while commons receives `localhost` and fails generated
  calls with a missing-schema error.

The active-server store is intended to make the selected deployment durable;
requiring a second, byte-identical environment configuration defeats that
design and can send authentication and API traffic to different origins.

Suggested fix — resolve one normalised active URL in reana-client and pass it
explicitly into `BaseAPIClient`/`get_current_api_client`; do not make commons
independently reread the raw environment. Couple this with PR545-08 and add an
integration test covering login with `--server-url`, no environment on the
next invocation, and a generated authenticated API call.

reana_client/auth/oidc.py
```diff
@@ -0,0 +1,514 @@
...
+        "code_verifier": code_verifier,
+        "code_challenge": code_challenge,
+        "code_challenge_method": PKCE_CODE_CHALLENGE_METHOD,
+    }
+
+
+def discover(server_url: str) -> Dict:
+    """Discover OIDC endpoints relayed by REANA server."""
+    normalized_url = normalize_server_url(server_url)
+    response = requests.get(
```
tiborsimko: **🤖 PR777-15: Login network failures escape as raw tracebacks (LOW)**

The new login flow makes four direct request sites before credentials are
stored: REANA discovery, authorization-code exchange, device authorization,
and device-token polling. These calls do not wrap `requests.RequestException`.
The `login` command catches only `AuthenticationError` and `ValueError`, so DNS,
connection, TLS, proxy, and timeout failures can escape through Click as raw
tracebacks. `refresh_credentials()` already demonstrates the intended pattern
by converting request failures into an actionable `AuthenticationError`.

Suggested fix — wrap each login-side request failure consistently, preserving
the cause for debug logging while showing a concise phase-specific error to
the user. Add tests for a discovery connection error, token-exchange timeout,
and device-polling transport failure and assert the command exits cleanly
without a traceback.

reana_client/auth/storage.py
tiborsimko: **🤖 PR777-01 follow-up (2026-07-28): verified fixed**

*Component:* reana-client — credential refresh persistence

Credential read/refresh/write is now protected by a cross-process file lock,
and the winner's rotated credential state is re-read before another process
refreshes. The atomic replacement path is retained, closing the refresh-token
reuse race between concurrent CLI processes.

Verified fixed at `9f08fe4`. No residual for the rotation race.

reana_client/auth/storage.py
tiborsimko: **🤖 PR777-02 follow-up (2026-07-28): verified fixed**

*Component:* reana-client — server URL normalisation

A schemeless server value is normalised to HTTPS rather than HTTP, and the
client validates the resulting URL before using it for OIDC discovery or API
calls. The accidental downgrade caused by simply omitting the scheme is gone.

Verified fixed at `9f08fe4`. No residual.

reana_client/cli/workflow.py
tiborsimko: **🤖 PR777-04 follow-up (2026-07-28): verified fixed**

*Component:* reana-client — interactive-session listing

The client uses the server's bulk `include_session_secrets` response instead of
issuing one secret request per listed session. Secret retrieval is therefore
one bounded API operation and an individual enrichment request can no longer
abort the entire listing.

Verified fixed at `9f08fe4`. No residual for the N+1/fail-all concern.

reana_client/auth/oidc.py
tiborsimko: **🤖 PR777-05 follow-up (2026-07-28): verified fixed**

*Component:* reana-client — device authorisation polling

Device polling now derives a local deadline from the issuer's `expires_in` and
uses bounded HTTP calls. It stops with a deterministic expiry error rather
than polling indefinitely when the issuer never returns a terminal response.

Verified fixed at `9f08fe4`. No residual.

reana_client/api/client.py
tiborsimko: **🤖 PR777-06 follow-up (2026-07-28): verified fixed**

*Component:* reana-client — interactive-session-secret API call

The session-secret operation is now part of the regenerated/shared API
contract and uses the same configured transport/authentication machinery as
the other generated calls. The ad-hoc helper that bypassed those policies is
gone.

Verified fixed at `9f08fe4`. No residual.

reana_client/auth/oidc.py
tiborsimko: **🤖 PR777-07 follow-up (2026-07-28): verified fixed**

*Component:* reana-client — device-flow deadline boundary

The polling loop now accepts and persists a successful token response returned
by the final bounded request; it does not discard that response merely because
the local clock passed the deadline while the request was in flight.

Verified fixed at `9f08fe4`. No residual.

reana_client/config.py
tiborsimko: **🤖 PR777-08 follow-up (2026-07-28): verified fixed**

*Component:* reana-client — insecure TLS mode

When `REANA_INSECURE` disables certificate verification, the CLI now emits an
explicit user-visible warning. Normal operation keeps TLS verification
enabled; the warning is covered by the amended tests.

Verified fixed at `9f08fe4`. No residual for the silent downgrade concern.

reana_client/auth/storage.py
tiborsimko: **🤖 PR777-11 follow-up (2026-08-11): verified fixed at 16adf24**

Stored server URLs are now normalised through a validator which rejects
non-HTTPS schemes, and discovered OIDC authorization, token, device and
end-session endpoints receive the same HTTPS check. Regression tests cover
plain HTTP and unrelated schemes, so neither a stored server nor issuer
metadata can silently downgrade the credential exchange.

reana_client/api/client.py
tiborsimko: **🤖 PR777-12 follow-up (2026-08-11): verified fixed at 16adf24**

File transfers now use an explicit `(30, 300)` connect/read timeout for both
upload and download requests. Tests assert that both paths pass the bounded
policy. A stalled file-transfer connection can therefore no longer block the
CLI indefinitely while still allowing a longer read window for large files.

reana_client/cli/utils.py
tiborsimko: **🤖 PR777-13 follow-up (2026-08-11): verified fixed at 16adf24**

An access token sourced from `REANA_ACCESS_TOKEN` is now checked for the
three-segment JWT compact form and rejected with an actionable migration
message when it is a legacy opaque credential. Explicit command-line token
input remains supported. The environment compatibility trap is therefore
detected at its source instead of failing later with an opaque API error.

reana_client/api/client.py
tiborsimko: **🤖 PR777-14 follow-up (2026-08-11): verified fixed at 16adf24**

The current-server API client now passes `get_active_server()` explicitly into
reana-commons, whose resolution order gives that explicit URL precedence over
the environment and bundled mapping without mutating shared specification
state. Tests cover a stored active server with no environment variable and the
precedence rules, so commands consistently target the selected server.

reana_client/auth/oidc.py
tiborsimko: **🤖 PR777-15 follow-up (2026-08-11): verified fixed at 16adf24**

OIDC discovery, authorization-code exchange, device authorization and device
polling now catch `requests.RequestException` and raise the CLI's
`AuthenticationError` with endpoint context. Regression tests exercise these
transport failures, replacing raw tracebacks with the intended user-facing
authentication error across all login flows.

reana_client/auth/oidc.py
```diff
@@ -0,0 +1,571 @@
...
+        httpd.handle_request()
+    return httpd.callback_query
+
+
+def _exchange_authorization_code(
+    metadata: Dict, code: str, pkce: Dict, redirect_uri: str
+) -> Dict:
+    """Exchange an authorization code for tokens using the PKCE verifier."""
+    try:
+        response = requests.post(
```
tiborsimko, line 329: **🤖 PR777-16: HTTPS validation is bypassed by automatic redirects (MEDIUM)**

`_validate_oidc_https_urls()` checks that the initially advertised OIDC URLs
use HTTPS, but the six outbound discovery and OIDC requests still use Requests'
default `allow_redirects=True`. An HTTPS endpoint can therefore redirect the
client to an HTTP destination after validation has succeeded. In particular,
Requests preserves the method and form body across 307/308 redirects, so an
authorization code, device code, refresh token, or revocation token can be
resent over cleartext. Requests rewrites ordinary 301/302/303 POST redirects
to GET; the credential-disclosure path here is specifically the 307/308
behaviour that retains the request body.

This leaves the new HTTPS policy dependent on every intermediate server never
returning a downgrade redirect. REANA Server already avoids this ambiguity:
`get_issuer_request_kwargs()` disables redirects and
`raise_for_issuer_status()` rejects 3xx responses explicitly.

Suggested fix — pass `allow_redirects=False` on every discovery, token, device
authorization, refresh, and revocation request, and report a controlled
authentication error for any 3xx response. Add regression coverage for an
HTTPS token endpoint returning a 307 redirect to HTTP, verifying that the
credential-bearing request is not followed.

Sohn123: Fixed — added allow_redirects=False to all six request sites (discovery, authorization-code exchange, device authorization, device-token polling, refresh, revocation) plus a shared _reject_redirect() helper that raises AuthenticationError on any 3xx response instead of following it. Added regression tests: test_discover_rejects_redirect_response (parametrized over 301/302/303/307/308) and test_refresh_credentials_rejects_redirect_response (a 307 on the token endpoint must not be followed with the refresh token in the body).

reana_client/auth/oidc.py
```diff
@@ -0,0 +1,658 @@
...
+        "token_endpoint": server_entry["token_endpoint"],
+        "authorization_endpoint": server_entry.get("authorization_endpoint"),
+        "device_authorization_endpoint": server_entry.get(
+            "device_authorization_endpoint"
+        ),
+        "revocation_endpoint": server_entry.get("revocation_endpoint"),
+    }
+    if "refresh_token" not in payload:
+        payload["refresh_token"] = refresh_token
+    return _store_token_response(normalized_url, metadata, payload)
```
tiborsimko, line 586: **🤖 PR777-17: An in-flight refresh can restore credentials after logout (MEDIUM)**

`refresh_credentials()` releases the credential lock before calling the token
endpoint and later stores a successful response unconditionally. This permits
the following interleaving:

1. refresh reads token T, releases the lock and starts the token request;
2. logout acquires the lock, revokes T, clears the entry and reports success;
3. the refresh response arrives and `_store_token_response()` recreates the
   cleared credential entry.

Holding the lock throughout `logout()` does not cover a refresh that was
already outside it. The CLI can therefore report a successful logout while
local credentials reappear immediately afterwards; whether the returned token
is still accepted then depends on the issuer's revocation semantics.

Suggested fix — coordinate refresh and logout through the same per-server
single-flight/generation state. Before storing a response, compare-and-swap
against the exact credential generation used to start the refresh; if logout
removed or invalidated that generation, discard the response. Add a regression
test that pauses an in-flight refresh, completes logout, then releases the
refresh and verifies that the credential entry remains absent.

reana_client/auth/oidc.py
```diff
@@ -0,0 +1,658 @@
...
+        if payload.get("error") == "invalid_grant":
+            # A concurrent process may have already rotated this exact
+            # refresh token and stored a new, valid one while this request
+            # was in flight -- only clear if the stored token is still the
+            # one that was just rejected, so a losing process's failure
+            # here can never destroy a winning process's fresh credentials.
+            if clear_token_material_if_matches(normalized_url, refresh_token):
+                raise AuthenticationError("Please run `reana-client login`.")
+            raise AuthenticationError(
+                "Credentials were refreshed by another concurrent "
```
tiborsimko, line 565: **🤖 PR777-18: Misleading concurrent-refresh message (NIT)**

`clear_token_material_if_matches()` returning `False` does not necessarily mean
that another process refreshed the credentials. It also returns `False` when
the stored token is absent, for example because a concurrent `logout` cleared
the entry while this refresh request was in flight. In that case “Credentials
were refreshed by another concurrent process. Please retry.” is incorrect, and
retrying cannot work without logging in again.

Suggested fix — reload and distinguish the two states: reuse/report a newer
stored credential when one exists, but report that the credentials were
removed and request a new login when the entry is absent.

reana_client/auth/storage.py
tiborsimko: **🤖 PR777-01 follow-up (2026-08-18): reopened after 920af80**

The credential-lock rewrite deliberately releases the lock before the token
request. Two CLI processes can therefore read and submit the same refresh token
concurrently again. `clear_token_material_if_matches()` prevents a losing
`invalid_grant` response from deleting a winner's stored credentials, but it
does not prevent duplicate refreshes or issuer-side refresh-token-family
revocation when rotation/reuse detection is enabled.

The objection to holding a global credential-file lock across a slow network
call is reasonable; the replacement needs per-server single-flight rather than
no in-flight coordination. One process should refresh while waiters block and
then re-read/reuse its stored result. The same coordination must cover logout,
as described in PR777-17.

Suggested fix — implement per-server refresh single-flight with waiters reusing
the winner's token, coordinate logout with the same generation/state, and add
multi-process tests for a rotating/reuse-detecting issuer.

reana_client/auth/storage.py
```diff
@@ -0,0 +1,345 @@
...
+def _acquire_file_lock(lock_file):
+    """Take a best-effort exclusive advisory lock across processes.
+
+    POSIX uses ``fcntl`` and Windows uses ``msvcrt``. When neither primitive is
+    available (or the OS lock cannot be taken) the in-process reentrancy guard
+    in :func:`credential_store_lock` still serialises access within one process,
+    and atomic file replacement still prevents partial writes.
+    """
+    if fcntl is not None:
+        fcntl.flock(lock_file.fileno(), fcntl.LOCK_EX)
```
tiborsimko: **🤖 PR777-09: Credential lock waits unboundedly and silently degrades on Windows (NIT)**

Moving the refresh-token network request outside the general credential-store
lock and adding a separate bounded refresh lock materially reduces contention.
A slow token endpoint no longer holds the general lock for the full HTTP round
trip.

The two platform-level residuals remain in `credential_store_lock()` itself.
POSIX acquisition still uses blocking `fcntl.flock(..., LOCK_EX)` with no
deadline or progress message, so a wedged holder can block login, logout, and
other credential-store updates indefinitely. On Windows,
`msvcrt.locking(..., LK_LOCK, 1)` still catches `OSError` and silently continues
without inter-process exclusion.

Suggested fix — use a bounded non-blocking retry loop for the general lock,
with one user-facing wait message, and do not silently proceed unlocked after a
Windows acquisition failure. Report a controlled error or an explicit logged
fallback instead.

reana_client/auth/oidc.py
```diff
@@ -0,0 +1,739 @@
...
+        }
+        if "refresh_token" not in payload:
+            payload["refresh_token"] = refresh_token
+
+        with credential_store_lock():
+            current_entry = get_server_entry(normalized_url)
+            if int(current_entry.get(CREDENTIAL_EPOCH_FIELD, 0)) != started_epoch:
+                _revoke_discarded_tokens(metadata, payload)
+                raise AuthenticationError("Please run `reana-client login`.")
+            return _store_token_response(normalized_url, metadata, payload)
```
tiborsimko: **🤖 PR777-19: An in-flight refresh can undo a concurrent server switch (MEDIUM)**

`refresh_credentials()` protects write-back with a per-server credential epoch,
but `_store_token_response()` ultimately calls `upsert_server_entry()`, which
always assigns the updated server to the global `active_server` field.

This leaves a cross-server race:

1. A refresh for server A records A's epoch and performs its token request
   outside the credential-store lock.
2. The user successfully logs in to server B, which changes B's epoch and makes
   B active; A's epoch is unchanged.
3. A's refresh completes, passes its per-A epoch comparison, stores the new A
   token, and changes `active_server` back to A.

This requires two concurrent `reana-client` processes—for example, a
long-running command refreshing A's token while `reana-client login` runs
against B in another terminal.

The CAS correctly prevents a refresh from undoing a logout or login affecting
the same server, but it cannot see this change to the global selection. The
next command can consequently run against A even though the user just selected
B, which is particularly risky when similarly named workflows exist on both
clusters.

Suggested fix — make token refresh update only the existing server entry and
preserve the current `active_server`; only an explicit login/server-selection
operation should make a server active. For example, add a `make_active` flag to
`upsert_server_entry()` and pass `False` from refresh, or use a separate update
primitive. Add a two-server concurrency test in which login to B completes
while refresh of A is in flight and assert that B remains active.

reana_client/auth/storage.py
```diff
@@ -0,0 +1,345 @@
...
+            return False
+    # Neither locking primitive is available: pessimistically report
+    # "already held" so callers fall back to their own refresh rather than
+    # risk two writers overlapping with no mutual exclusion at all.
+    return False
+
+
+def _refresh_lock_path() -> str:
+    """Return the path of the advisory lock scoped to refresh attempts."""
+    return "{}.refresh.lock".format(os.path.abspath(get_config_path()))
```
tiborsimko: **🤖 PR777-20: Refresh single-flight serialises unrelated REANA servers (LOW)**

The new refresh advisory lock is named only from the credential configuration
path (`<config>.refresh.lock`). It is therefore shared by every REANA server in
that file, even though refresh tokens and credential epochs are stored and
validated per server.

A slow refresh against server A makes a command using server B wait up to the
full refresh-lock timeout, currently 35 seconds. Once A releases the lock, B's
entry is still unrefreshed, so each B waiter that finds no valid access token
can fall back to an unlocked refresh. The global lock thus introduces
unnecessary cross-cluster latency and can still permit a burst of redundant
refreshes for B.

Suggested fix — scope the refresh-lock filename to a stable hash of the
normalised server URL, without placing credentials or the raw URL in the
filename. Add tests showing that two refreshes for the same server are
single-flight while refreshes for different servers can proceed independently.

reana_client/cli/utils.py
```diff
@@ -9,18 +9,22 @@
...
 import shlex
 import sys
 import time
 import re
 from typing import Callable, NoReturn, Optional, List, Tuple, Union, Iterable
+from urllib.parse import parse_qsl, urlencode, urlsplit, urlunsplit
 
 import click
 import tablib
+from click.core import ParameterSource
```
tiborsimko, line 23: **🤖 PR777-21: The CLI allows Click 7 but imports a Click 8-only API (MEDIUM)**

`reana_client/cli/utils.py` now imports `ParameterSource` from `click.core` and
uses it while resolving authentication options. `ParameterSource` was added in
Click 8, but `setup.py` still declares `click>=7`. Installing the declared
lowest dependency therefore fails during module import with:

```text
ImportError: cannot import name 'ParameterSource' from 'click.core'
```

The same function also calls `Context.get_parameter_source()`, which Click 7
likewise does not provide, so the incompatibility is not confined to the
import.

This prevents the CLI and its tests from starting under an otherwise permitted
dependency set; the current lowest-dependency CI matrix exercises that failure.
It is a functional lower-bound incompatibility introduced by this PR, not a
future shared-package release concern.

Suggested fix — either require `click>=8` in `install_requires`, or replace
both Click 8-only APIs with a Click 7-compatible way to distinguish an
explicitly supplied token from a default; removing only the import is not
sufficient. Keep a lowest-dependency test that imports the CLI authentication
option code so the declared bound and runtime API use cannot diverge again.

reana_client/auth/storage.py
tiborsimko: **🤖 PR777-01 follow-up (2026-08-21): verified fixed at d5544e9**

Refresh is now coordinated by a per-server inter-process lock. A contender
waits for the in-flight refresh, reloads the credential entry, and reuses the
new token instead of issuing a second rotating-token request. Epoch checks
also prevent a stale failure or response from clearing or replacing newer
credentials. The tests cover same-server single-flight behaviour. No residual
concern remains for this finding.

reana_client/auth/storage.py
tiborsimko: **🤖 PR777-09 follow-up (2026-08-21): partially addressed at d5544e9**

The credential-store lock now uses a bounded non-blocking retry loop on both
POSIX and Windows, and Windows acquisition failures no longer silently fall
back to an unlocked critical section. Those changes resolve the locking and
platform-correctness parts of the finding.

On timeout, `_acquire_file_lock()` raises a bare `TimeoutError`. The `login`
and `logout` commands catch `AuthenticationError` and `ValueError`, and there
is no top-level CLI translation for `TimeoutError`, so those commands terminate
with a traceback. Other authenticated commands reach the same lock through
`access_token_check()`: its catch-all logs the traceback and then displays the
generic missing-access-token message. A stale lock is therefore misdiagnosed
as "not signed in", directing the user towards another login attempt instead
of the stuck credential-lock holder.

Suggested fix — translate lock timeout into the CLI's normal authentication or
usage error at the credential-store boundary, preserving the actionable lock
message for every caller and avoiding both a traceback and the misleading
missing-token fallback. Add `CliRunner` coverage for login, logout, and an
ordinary authenticated command with a timed-out general credential lock.

reana_client/auth/oidc.py
tiborsimko: **🤖 PR777-16 follow-up (2026-08-21): verified fixed at d5544e9**

All discovery, device-authorisation, token, refresh, and revocation requests
now disable automatic redirects and classify 3xx responses as controlled
authentication failures. Regression coverage verifies that a credential-
bearing POST is not followed through a 307 downgrade. No residual concern
remains for this finding.

reana_client/auth/oidc.py
tiborsimko: **🤖 PR777-17 follow-up (2026-08-21): verified fixed at d5544e9**

Refresh write-back now compares the credential epoch captured before the
network request with the current entry under the store lock. Logout removes
or advances that state, so a response from an older in-flight refresh is
discarded instead of recreating credentials. The interleaving is covered by a
regression test. No residual concern remains for this finding.

reana_client/auth/oidc.py
tiborsimko: **🤖 PR777-18 follow-up (2026-08-21): verified fixed at d5544e9**

The concurrent-change path now reloads state and distinguishes a newer stored
credential from an entry removed by login/logout. It no longer reports every
failed compare-and-set as another successful refresh, and the user receives
the appropriate retry or login guidance. No residual concern remains for this
finding.

reana_client/auth/oidc.py
tiborsimko: **🤖 PR777-19 follow-up (2026-08-21): verified fixed at d5544e9**

Credential refresh now writes the server entry with `make_active=False`.
Only explicit login/server-selection operations change `active_server`, so a
late refresh for server A cannot undo a concurrent switch to server B. The
two-server interleaving is covered by tests. No residual concern remains for
this finding.

reana_client/auth/storage.py
tiborsimko: **🤖 PR777-20 follow-up (2026-08-21): verified fixed at d5544e9**

Refresh lock files are now scoped by a stable hash of the normalised server
URL. The raw URL and credentials are not exposed in the filename, same-server
refreshes remain single-flight, and independent servers can refresh without
blocking one another. Tests cover both lock-path separation and concurrent
acquisition. No residual concern remains for this finding.

reana_client/cli/utils.py
tiborsimko: **🤖 PR777-21 follow-up (2026-08-21): verified fixed at d5544e9**

`install_requires` now declares `click>=8`, matching the imported
`ParameterSource` and `Context.get_parameter_source()` APIs. The declared
lower bound no longer permits an installation that fails while importing the
CLI. No residual concern remains for this finding.

reana_client/auth/storage.py
Sohn123: Fixed, including the residual your follow-up called out. New `CredentialStoreError` exception replaces the bare `TimeoutError` `_acquire_file_lock` used to raise. It's now caught at the CLI boundary in `cli/utils.py` (alongside `AuthenticationError`/`ValueError`) and explicitly in `cli/ping.py`'s `login`/`logout` — the two commands your follow-up specifically named as still terminating with a raw traceback or a misleading "not signed in" message instead of the actionable lock message.

reana_client/auth/oidc.py
```diff
@@ -0,0 +1,788 @@
...
+        self.end_headers()
+        self.wfile.write(body)
+
+    def log_message(self, *args):
+        """Silence the default stderr request logging."""
+
+
+def _start_callback_server() -> Tuple[HTTPServer, str]:
+    """Start a loopback HTTP server and return it with its redirect URI."""
+    requested_port = int(os.getenv(LOOPBACK_PORT_ENV, "0"))
```
tiborsimko: **🤖 PR777-22: The loopback-port override is unvalidated and undocumented (LOW)**

`_start_callback_server()` parses the new environment variable before its
exception handling and does not validate the port range:

```python
requested_port = int(os.getenv(LOOPBACK_PORT_ENV, "0"))
try:
    httpd = HTTPServer((LOOPBACK_HOST, requested_port), _CallbackHandler)
except OSError as error:
```

A non-numeric value becomes a bare `ValueError` message such as `invalid
literal for int()`, which is not actionable for a user who may not know which
setting caused it. A value outside the valid port range can make `HTTPServer`
raise `OverflowError`; neither this block nor the CLI wrapper catches that
exception, so login terminates with a traceback. This reintroduces the kind of
uncontrolled authentication failure that the surrounding error handling is
intended to avoid.

`REANA_CLIENT_LOGIN_LOOPBACK_PORT` is also absent from the client and operator
documentation, even though the commit introduces it specifically for identity
providers which require an exact callback URI. Operators are not told to
register `http://127.0.0.1:<port>/callback`, and users are not told to set the
matching value.

Suggested fix — parse and range-check the value, translate malformed and
out-of-range input into `AuthenticationError` which names the environment
variable, and add regression tests for both cases. Document the setting and
the exact redirect URI alongside the loopback login and external-provider
registration instructions.

tiborsimko: **🤖 PR777-22 follow-up (2026-08-25): verified fixed at 2125727**

`REANA_CLIENT_LOGIN_LOOPBACK_PORT` is now parsed and range-checked before
`HTTPServer` is constructed. Non-numeric and out-of-range values raise a
controlled `AuthenticationError` which names the setting, while bind failures
for a valid fixed port explain why that port is required and how to recover.
The regression tests cover all three cases, including the former raw
`OverflowError` path.

The operator access guide in reanahub/docs.reana.io#231 now documents the
ephemeral default, the fixed-port trade-off, and the exact
`http://127.0.0.1:<port>/callback` redirect URI to register with providers
which require port matching. No residual concern remains for this finding.

setup.py
```diff
@@ -37,11 +37,12 @@
     extras_require["all"].extend(reqs)
 
 install_requires = [
-    "click>=7",
+    "click>=8",
     "pathspec>=0.9.0,<1.0 ; python_version<'3.9'",
     "pathspec>=0.9.0 ; python_version>='3.9'",
     "jsonpointer>=2.0",
     "reana-commons[yadage,snakemake,cwl]>=0.95.0a22,<0.96.0",
+    "requests>=2.20.0",
```
tiborsimko: **🤖 PR777-23: The declared requests floor breaks a supported lowest-dependency environment (LOW)**

The new direct dependency is declared as:

```python
"requests>=2.20.0",
```

The lowest-dependency job consequently pins `requests==2.20.0`, whose
transitive constraint selects `urllib3==1.24.3`. On the PR's declared Python
3.13 environment, pytest then fails before collection:

```text
ModuleNotFoundError: No module named 'urllib3.packages.six.moves'
```

The failing import is reached through `reana_commons.api_client` importing
Bravado, then `bravado_core.spec` importing Requests, which imports urllib3.
This is the only lowest-version matrix job that ran to completion before
fail-fast cancelled the others, so the branch currently has a demonstrated
unsupported combination inside its advertised Python range. This is separate
from the deferred reana-commons release ordering: the failure is in the public
dependencies selected from the new direct floor.

Suggested fix — raise the Requests lower bound to a version whose transitive
dependency range imports on every Python version declared by reana-client,
and re-run the full lowest-dependency matrix. Choose the floor based on the
resolved urllib3 range rather than assuming that any increment over 2.20 is
sufficient. The lowest-dependency set is built with
`requirements-builder --level=min`, which pins declared requirements only and
leaves transitive ones at their highest compatible version. The effective
urllib3 is therefore the highest version permitted by the chosen Requests
release's cap, which is the range the matrix needs to verify.

README.md
```diff
@@ -77,6 +77,48 @@ and reports effective runtime identities. Add `--pull` to verify availability
...
+   $ REANA_CLIENT_CONFIG=./reana-client.json reana-client login --server-url <your-server>
+   ```
+
+2. Store the resulting `reana-client.json` file's contents as a CI secret (e.g.
+   a masked/protected variable), not in the repository.
+3. In the CI job, write that secret out to a file and point
+   `REANA_CLIENT_CONFIG` at it before running `reana-client`:
+
+   ```console
+   $ echo "$REANA_CREDENTIALS_SECRET" > /tmp/reana-client.json
```
tiborsimko: **🤖 PR777-24: The CI credential recipe discards the store's 0600 protection (MEDIUM)**

The authentication section promises that the credential file containing the
access and refresh tokens is permission-restricted to `0600`, but the generic
CI recipe creates it through ordinary shell redirection at a fixed path:

```console
$ echo "$REANA_CREDENTIALS_SECRET" > /tmp/reana-client.json
```

Redirection creates a new file with mode `0666` filtered only by the process
umask. With the common `022` umask this is `0644`, making the refresh token
readable by other local users on a shared or multi-user self-hosted runner.
The predictable name also permits collisions and, depending on runner and OS
hardening, pre-created-link attacks in the world-writable `/tmp` directory.

`save_config()` eventually replaces files with mode `0600`, but only when the
client writes the credential store. A job whose access token remains valid
can read the exposed file without ever correcting its permissions, and the
example does not remove it afterwards.

Suggested fix — create a unique credential file with `mktemp`, explicitly
restrict it to `0600`, write with `printf`, and remove it from an `EXIT` trap.
For example:

```console
$ credential_file="$(mktemp)"
$ trap 'rm -f "$credential_file"' EXIT
$ chmod 600 "$credential_file"
$ printf '%s' "$REANA_CREDENTIALS_SECRET" > "$credential_file"
$ export REANA_CLIENT_CONFIG="$credential_file"
```

README.md
```diff
@@ -77,6 +77,48 @@ and reports effective runtime identities. Add `--pull` to verify availability
...
+   $ export REANA_CLIENT_CONFIG=/tmp/reana-client.json
+   $ reana-client ping
+   ```
+
+The stored refresh token has the lifetime your identity provider issues it with;
```
tiborsimko, line 122: **🤖 PR777-25: An immutable CI secret cannot retain rotated refresh tokens (LOW)**

The documented CI pattern stores one credential JSON document as a CI secret
and recreates the same starting file for every job. The client correctly
persists a replacement refresh token returned by the issuer, but only into
that job-local copy; the CI secret remains unchanged.

For an issuer configured to rotate refresh tokens and reject reuse, the
sequence is therefore:

1. A job restores refresh token R1 from the CI secret.
2. Its first refresh succeeds, invalidates R1, and writes R2 locally.
3. The job ends without updating the CI secret.
4. The next job restores R1 again and receives `invalid_grant`.

Concurrent jobs can fail similarly because each starts with its own copy of
R1; reana-client's per-file advisory lock cannot coordinate separate runners
or credential files. Requesting `offline_access` does not remove this issue:
providers, including Keycloak when refresh-token revocation is enabled, can
rotate offline tokens too.

The current text discusses only the token's eventual lifetime and presents
the copied-secret recipe as the supported automation pattern, so operators
are not warned that issuer rotation policy determines whether it survives
even the first refresh.

Suggested fix — document the no-rotation prerequisite if this remains the
recommended interim pattern. For provider-independent automation, use a
mutable and serialised credential store which retains refresh write-backs, or
provide a non-interactive workload credential flow; an immutable per-job copy
cannot safely preserve rotating credentials.

setup.py
tiborsimko: **🤖 PR777-23 follow-up (2026-08-26): verified fixed at 975f2e0**

The Requests floor is now `requests>=2.31.0`. Its `urllib3<3` upper bound lets
the lowest-dependency resolver select a modern urllib3 instead of the
Python-3.13-incompatible 1.24.3 selected under Requests 2.20. The specific
pre-collection import failure reported here is therefore no longer present in
the supported dependency set.

The commit also records reproduction with the real
`requirements-builder --level=min` installation and the Python 3.13 test
environment. No residual concern remains for this finding.

README.md
tiborsimko: **🤖 PR777-25 follow-up (2026-08-26): verified fixed at 03c9b6e**

The CI guidance now states the non-rotation prerequisite explicitly. It
explains that a rotated replacement is written only to the job-local copy,
that subsequent and concurrent jobs can restore an invalidated token and fail
with `invalid_grant`, and that `offline_access` does not remove this issuer-
policy dependency.

For deployments which rotate refresh tokens, the documentation points to a
mutable, serialised secret store (or a provider-supported non-rotating service
credential) instead of presenting the immutable-secret recipe as generally
safe. No residual concern remains.

README.md
tiborsimko: **🤖 PR777-24 follow-up (2026-08-26): verified fixed at 03c9b6e**

The CI recipe now creates a unique path with `mktemp`, installs an `EXIT` trap,
explicitly applies mode `0600`, writes with `printf`, and points
`REANA_CLIENT_CONFIG` at that private file. This removes both the permissive-
umask window and the predictable `/tmp/reana-client.json` target, and ensures
the credential copy is removed when the shell exits.

The surrounding explanation also states why ordinary redirection was unsafe.
No residual concern remains.

reana_client/auth/oidc.py
```diff
@@ -0,0 +1,853 @@
...
+        if seconds < 0:
+            raise ValueError
+        return format_timestamp(utcnow() + timedelta(seconds=seconds))
+    except (TypeError, ValueError, OverflowError) as exc:
+        raise AuthenticationError(
+            "Authentication server returned an invalid 'refresh_expires_in' value."
+        ) from exc
+
+
+def _device_login_parameters(device_response: Dict) -> Tuple[str, int, int]:
```
tiborsimko, line 188: **🤖 PR777-26: Device-flow response validation accepts unusable prompts and unbounded expiry (LOW)**

`_device_login_parameters()` validates the values consumed by the polling
loop (`device_code`, `expires_in`, and `interval`), but it does not validate the
values the user needs to authorise that code. A successful response without
the relevant fields reaches the display callback, which prints `None` for the
URL and code and then polls until the device code expires.

A response is actionable when it supplies either a non-empty
`verification_uri_complete`, or both a non-empty `verification_uri` and
`user_code`. RFC 8628 also defines `user_code` as required, so accepting its
absence should at least produce an explicit compatibility warning rather than
a silent unusable prompt.

The numeric validation is incomplete as well. Python accepts arbitrary-size
integers, so `expires_in = 10**400` passes `int()` and the positive-value check.
The subsequent calculation is outside the helper's exception handling:

```python
deadline = monotonic() + expires_in
```

It raises a raw `OverflowError` instead of the controlled `AuthenticationError`
used for the other malformed response values.

Suggested fix — validate the actionable prompt-field combination and reject
unusable values before displaying anything. Bound the device lifetime to a
representable/practical value (or construct the deadline inside the controlled
validation block), and add tests for absent prompt fields and an oversized
integer expiry. The equivalent Go-client validation gap is raised separately
as PR201-07.

reana_client/auth/oidc.py
```diff
@@ -0,0 +1,853 @@
...
+        )
+    entry = {
+        "issuer": metadata["issuer"],
+        "client_id": metadata["reana_cli_client_id"],
+        "token_endpoint": metadata["token_endpoint"],
+        "authorization_endpoint": metadata.get("authorization_endpoint"),
+        "device_authorization_endpoint": metadata.get("device_authorization_endpoint"),
+        "revocation_endpoint": metadata.get("revocation_endpoint"),
+        "access_token": access_token,
+        "access_token_expires_at": _token_expires_at(token_response),
```
tiborsimko: **🤖 PR777-27: Rejecting a token response can orphan its rotated refresh token (LOW)**

The refresh endpoint may already have consumed R1 and returned its replacement
R2 before `_store_token_response()` validates the response lifetimes. Building
the new entry calls `_token_expires_at()` and
`_refresh_token_expires_at()`:

```python
"access_token_expires_at": _token_expires_at(token_response),
"refresh_token_expires_at": _refresh_token_expires_at(token_response),
```

Either helper now raises `AuthenticationError` for a malformed lifetime. That
exception leaves the old, consumed R1 in the local credential store and drops
R2 without persisting or revoking it. The next command eventually receives
`invalid_grant` and asks the user to log in again, while R2 can remain live at
the issuer despite never being represented locally.

The adjacent credential-epoch mismatch path calls
`_revoke_discarded_tokens()` for exactly this discarded-token hazard, so the
validation-failure path is an inconsistent gap.

Suggested fix — validate/construct the replacement entry before entering the
credential write-back section and, if it cannot be stored, best-effort revoke
the returned refresh token before raising. For malformed access-token expiry,
falling back to the access JWT's `exp` as `_token_expires_at()` already does
when `expires_in` is absent can preserve a trustworthy expiry; revoke and fail
when no safe fallback exists. Add a refresh-rotation regression test which
returns R2 with a malformed lifetime and asserts that R2 is not left live and
untracked.

reana_client/auth/oidc.py
```diff
@@ -0,0 +1,853 @@
...
+        self.end_headers()
+        self.wfile.write(body)
+
+    def log_message(self, *args):
+        """Silence the default stderr request logging."""
+
+
+def _start_callback_server() -> Tuple[HTTPServer, str]:
+    """Start a loopback HTTP server and return it with its redirect URI."""
+    raw_port = os.getenv(LOOPBACK_PORT_ENV, "0")
```
tiborsimko: **🤖 PR777-28: An empty loopback-port override differs between the two clients (NIT)**

Python reads the environment value without trimming it:

```python
raw_port = os.getenv(LOOPBACK_PORT_ENV, "0")
requested_port = int(raw_port)
```

An exported-but-empty value, or one containing only whitespace, is therefore a
hard configuration error. The Go client trims the same variable and treats an
empty result as unset, selecting an ephemeral port. Shell and CI configuration
commonly use forms such as
`REANA_CLIENT_LOGIN_LOOPBACK_PORT="${PORT:-}"`, so otherwise interchangeable
clients behave differently for the same environment.

Suggested fix — trim the Python value and treat an empty result as `0`/unset,
matching the Go client. Retain the current controlled error for non-empty
malformed values.

reana_client/auth/storage.py
tiborsimko: **🤖 PR777-01 follow-up (2026-08-27): still partial at 522f270**

The per-server refresh lock, conditional `invalid_grant` clearing, and
credential-epoch write-back now protect the normal successful-leader path and
the tested login/logout races. A failed refresh leader still releases every
waiter into concurrent token rotation, however.

`wait_for_refresh_lock()` acquires the newly available lock only momentarily,
releases it, and returns `True`. If the failed leader did not leave a valid
access token, the caller then falls through to `requests.post()` with
`refresh_lock_file is None`; every other waiter follows the same path. An
issuer 5xx or timeout can therefore wake multiple processes into simultaneous
use of the same rotating refresh token.

One waiter can rotate R1 to R2 while another receives `invalid_grant`, clears
the still-current R1, and increments `credential_epoch`. The successful
waiter's write-back then sees a stale epoch and revokes R2, leaving neither
token usable and forcing login again. A leader which rejects a newly rotated
response during `_store_token_response()` has the same wake-up effect.

The timeout fallback is unsafe for the same reason: it starts an unlocked
request while a live process still owns the descriptor-bound file lock. A dead
process cannot leave a stale `flock` behind. The one platform caveat is the
existing no-locking-primitive fallback: when neither `fcntl` nor `msvcrt` is
available, refresh currently degrades to unlocked operation and should remain
distinguishable from genuine contention.

Suggested fix — distinguish “locking unsupported” from “lock held”. For real
contention, return the acquired lock handle from the wait operation and keep it
held across rereading credentials and any replacement refresh. If the overall
deadline expires, return a controlled retryable error without issuing another
token request; preserve today's unlocked fallback only on platforms with no
locking primitive. Add deterministic failed-leader/multiple-waiter and
timeout-with-no-extra-request tests.

reana_client/auth/oidc.py
```diff
@@ -0,0 +1,905 @@
...
+            ),
+            "revocation_endpoint": server_entry.get("revocation_endpoint"),
+        }
+        if "refresh_token" not in payload:
+            payload["refresh_token"] = refresh_token
+
+        with credential_store_lock():
+            current_entry = get_server_entry(normalized_url)
+            if int(current_entry.get(CREDENTIAL_EPOCH_FIELD, 0)) != started_epoch:
+                _revoke_discarded_tokens(metadata, payload)
```
tiborsimko: **🤖 PR777-29: Discarded-token revocation runs under the credential-store lock (MEDIUM)**

The refresh request itself correctly runs outside `credential_store_lock()`,
but the epoch-mismatch cleanup puts a network request back inside that global
lock:

```python
with credential_store_lock():
    current_entry = get_server_entry(normalized_url)
    if int(current_entry.get(CREDENTIAL_EPOCH_FIELD, 0)) != started_epoch:
        _revoke_discarded_tokens(metadata, payload)
        raise AuthenticationError("Please run `reana-client login`.")
```

`_revoke_discarded_tokens()` performs an HTTPS POST with a 30-second timeout.
The credential-store lock also has a 30-second acquisition timeout. If a
concurrent login or logout changes the epoch while a refresh is in flight and
the issuer's revocation endpoint is slow, every other `reana-client` process
using this credential file is serialised behind that request and can fail with
a credential-lock timeout. This violates the surrounding single-flight design,
which deliberately avoids holding the store-wide lock across issuer I/O.

The Go client already keeps these phases separate: `PutIfEpoch()` returns
before its best-effort revocation call.

Suggested fix — decide the epoch mismatch and copy the discarded token data
while holding the lock, then release it before calling
`_revoke_discarded_tokens()` and raising the controlled error. Add a regression
test which blocks the revocation POST and proves an unrelated credential-store
operation can still complete promptly.

reana_client/cli/workflow.py
```diff
@@ -1713,28 +1701,55 @@ def workflow_open_interactive_session(
...
+                logging.debug(traceback.format_exc())
+                logging.debug(str(secret_error))
+                session_secret = None
             display_message(
                 "Interactive session opened successfully", msg_type="success"
             )
+            if session_secret is None:
+                display_message(
+                    "Could not retrieve the session's access token; "
+                    "run `reana-client open` again to get the full URL.",
```
tiborsimko, line 1720: **🤖 PR777-30: The session-open warning recommends a command that cannot work (LOW)**

When session creation succeeds but the follow-up secret fetch fails, the new
warning says:

```text
Could not retrieve the session's access token; run `reana-client open` again
to get the full URL.
```

The session is already open at this point. Repeating `reana-client open`
reaches the workflow-controller's existing-session guard and returns
`Interactive session is already open` with HTTP 404, so the advertised
recovery path cannot retrieve the URL.

`reana-client list --sessions` is the working recovery path: the list endpoint
can return the per-session secret for owned interactive workflows, and the
client already uses it to render the full URI.

Suggested fix — direct the user to `reana-client list --sessions` (optionally
with the relevant workflow filter) instead of asking them to repeat `open`.
Update the regression test to assert the usable recovery command.

reana_client/auth/storage.py
```diff
@@ -0,0 +1,453 @@
...
+        lock_file.seek(0)
+        try:
+            msvcrt.locking(lock_file.fileno(), msvcrt.LK_NBLCK, 1)
+            return True
+        except OSError:
+            return False
+    # Neither locking primitive is available: pessimistically report
+    # "already held" so callers fail after the bounded wait rather than risk
+    # two refresh-token rotations overlapping with no mutual exclusion.
+    return False
```
tiborsimko, line 151: **🤖 PR777-31: Unsupported refresh locking waits before reporting nonexistent contention (NIT)**

When neither `fcntl` nor `msvcrt` is available,
`_try_acquire_file_lock_nb()` deliberately returns `False` to keep refresh
token rotation fail-closed:

```python
# Neither locking primitive is available: pessimistically report
# "already held" ...
return False
```

That condition is deterministic, not contention. Every refresh nevertheless
enters `wait_for_refresh_lock()`, polls for the full 35-second deadline, and
then reports `Timed out waiting for another credential refresh`, even though
no process can acquire the lock on that platform. The store lock in the same
module handles the condition differently by immediately degrading to its
documented no-op fallback.

Failing closed for refresh-token rotation is the right policy; the issue is
the delayed and inaccurate failure shape on the explicitly handled fallback
branch.

Suggested fix — detect the absence of both locking primitives before entering
the contention loop and fail immediately with a controlled message explaining
that safe cross-process refresh locking is unavailable on this platform. Keep
unlocked refresh disabled.

reana_client/auth/oidc.py
```diff
@@ -0,0 +1,905 @@
...
+    access_token = token_response.get("access_token")
+    try:
+        if not isinstance(access_token, str) or not access_token:
+            raise AuthenticationError(
+                "Authentication server did not return an access token."
+            )
+        access_token_expires_at = _token_expires_at(token_response)
+        refresh_token_expires_at = _refresh_token_expires_at(token_response)
+    except AuthenticationError:
+        if refresh_token:
```
tiborsimko, line 362: **🤖 PR777-32: A failed login recovery switches the active server (LOW)**

`_store_token_response()` preserves a returned refresh token when another
response field is malformed, which avoids losing a token that the issuer has
already rotated. On an explicit login, however, the recovery write forwards
the normal `make_active=True` value:

```python
except AuthenticationError:
    if refresh_token:
        upsert_server_entry(server_url, recovery_entry, make_active=make_active)
    raise
```

If server A is currently active and a login to server B receives a refresh
token but an invalid access token or lifetime, the command reports that login
failed while silently changing `active_server` to B. The next unrelated
command therefore targets a server whose login was rejected instead of the
still-working session on A.

Preserving the refresh token and committing the user's server switch are two
different operations. The former is needed even on a rejected response; the
latter should happen only after the response has been fully validated.

Suggested fix — make the recovery write preserve the replacement refresh
token with `make_active=False`, and mark the server active only on the normal
successful store path. Add a two-server test proving a malformed login to B
does not displace an active session on A.

reana_client/auth/oidc.py
```diff
@@ -0,0 +1,905 @@
...
+
+def _device_login_parameters(device_response: Dict) -> Tuple[str, int, int]:
+    """Validate and return device code, lifetime, and polling interval."""
+    try:
+        device_code = device_response.get("device_code")
+        if not isinstance(device_code, str) or not device_code:
+            raise ValueError
+        expires_in = device_response.get("expires_in")
+        if not isinstance(expires_in, int) or isinstance(expires_in, bool):
+            raise ValueError
```
tiborsimko, line 196: **🤖 PR777-33: An over-limit device lifetime is reported as a malformed response (NIT)**

The one-hour local polling bound is a reasonable defence against overflowing
or effectively unbounded issuer values, but the validation folds a value over
that policy limit into the missing/malformed response error:

```python
if expires_in <= 0 or expires_in > DEVICE_FLOW_MAX_SECONDS:
    raise ValueError
...
raise AuthenticationError(
    "Device login response did not contain a valid code and expiry."
)
```

Device-code lifetime is an administrator-configurable realm/client value in
Keycloak. A well-formed response with `expires_in=3601` therefore fails with a
message suggesting that the issuer omitted or malformed required fields,
rather than explaining REANA's local one-hour limit. The oversized polling
interval path has the same diagnostic ambiguity.

Suggested fix — keep the bound, but validate policy-limit violations
separately and report the supported maximum explicitly, for example `Device
login lifetime exceeds REANA's 3600-second polling limit`. Do the same for an
over-limit polling interval and cover both messages in the existing malformed
device-response tests.

reana_client/auth/oidc.py
```diff
@@ -0,0 +1,905 @@
...
+        raise AuthenticationError(
+            "Device login failed: "
+            f"{payload.get('error_description') or error or response.text}"
+        )
+
+
+def _access_token_valid(server_entry: Dict) -> bool:
+    """Return whether stored access token can be used now."""
+    access_token = server_entry.get("access_token")
+    expires_at = parse_timestamp(server_entry.get("access_token_expires_at"))
```
tiborsimko, line 701: **🤖 PR777-34: A malformed stored expiry crashes token validation (LOW)**

`_access_token_valid()` parses the credential-file timestamp before applying
any controlled error handling:

```python
access_token = server_entry.get("access_token")
expires_at = parse_timestamp(server_entry.get("access_token_expires_at"))
...
return expires_at - timedelta(seconds=EXPIRY_LEEWAY_SECONDS) > utcnow()
```

`datetime.fromisoformat()` raises `ValueError` for malformed text. A
syntactically valid timestamp without a timezone produces a naive `datetime`
instead, and comparing it with the timezone-aware `utcnow()` raises
`TypeError`. Either form in a damaged, manually edited, or legacy credential
file escapes the authentication error boundary and makes commands fail with a
raw traceback until the file is repaired manually.

Suggested fix — treat an unparseable or timezone-less stored expiry as an
invalid access token and enter the normal refresh/login recovery path, or
raise a controlled `AuthenticationError` which identifies the damaged
credential entry. Add tests for both malformed text and a naive ISO timestamp.

reana_client/auth/oidc.py
tiborsimko: **🤖 PR777-27 follow-up (2026-08-28): verified fixed at 8ac7e19**

`_store_token_response()` now constructs a refresh-token-only recovery entry
before validating the access token and lifetime fields. If validation rejects
the response after the issuer has rotated the refresh token, the replacement
is persisted without accepting the rejected access token, so the consumed
predecessor is no longer left as the only local credential.

The regression test returns a rotated refresh token with a malformed
`expires_in`, verifies the controlled error, and confirms that the replacement
was retained while the rejected access token was not. The orphaned-token
failure described in this finding is resolved.

reana_client/auth/storage.py
tiborsimko: **🤖 PR777-01 follow-up (2026-08-28): verified fixed at 8ac7e19**

`refresh_credentials()` now loops back through
`try_acquire_refresh_lock()` after every completed wait. If a leader failed
without storing a usable access token, exactly one waiter acquires leadership
and performs the replacement refresh while the remaining waiters continue to
contend. They can no longer all fall through into concurrent use of the same
rotating refresh token.

The overall deadline is retained across re-contention, and expiry returns a
controlled error without issuing an unlocked token request. The new
failed-leader regression test asserts that the successor performs exactly one
refresh. The deadline-expiry branch is verified by inspection only: the
requested test asserting that an expired wait issues no token request is not
present, and would be worth adding to lock the behaviour in. No residual
concern remains for the refresh-token serialisation race itself.

reana_client/auth/oidc.py
tiborsimko: **🤖 PR777-26 follow-up (2026-08-28): verified fixed at 8ac7e19**

Device-flow success responses now require a non-empty device code, a bounded
positive integer lifetime, and a usable HTTPS verification prompt. A
verification URI without a complete embedded code additionally requires a
non-empty user code. Polling intervals are type-checked and bounded before
they are converted or used in the polling loop.

The parameterised regression coverage exercises missing prompts and user
codes, insecure verification URLs, oversized lifetimes, and invalid polling
intervals before anything is displayed to the user. The unusable-prompt and
unbounded-duration paths identified in this finding are closed.

reana_client/auth/oidc.py
tiborsimko: **🤖 PR777-28 follow-up (2026-08-28): verified fixed at 8ac7e19**

The Python callback-server path now trims
`REANA_CLIENT_LOGIN_LOOPBACK_PORT` and maps an absent, empty, or whitespace-only
value to port zero. It therefore retains ephemeral-port behaviour under the
same empty override that the Go client treats as unset.

The parameterised regression test covers both an empty string and whitespace.
The cross-client environment-variable discrepancy is resolved.

reana_client/auth/oidc.py
tiborsimko: Addressed in `fc63653` (original fixup `c8ae30a`, now squashed). Discarded-token revocation now runs outside the credential-store lock, with a regression test proving that unrelated credential updates can proceed while revocation is blocked.

tiborsimko (approved):

PR #778, opened 2026-06-26
feat(workflow): scope interactive session secrets

### LINKED ISSUE #978: Allow restricting which user secrets are exposed to workflow jobs and interactive sessions
After #805, user secrets can be propagated to interactive sessions. Today, secrets are generally exposed broadly to workflow job pods, Dask workers, and interactive sessions. This is convenient, but it also means pods receive secrets they do not actually need.

## Problem

There is currently no least-privilege way for a workflow author or an interactive-session caller to restrict which user secrets are exposed to a given pod/session.

We should keep the current default behaviour for backward compatibility, but provide an explicit opt-in mechanism to reduce the exposed secret set when desired.

## Proposed solution

Introduce an optional `secret_names` allowlist.

Behaviour:

- If `secret_names` is omitted: keep current behaviour and expose all user secrets.
- If `secret_names` is an empty list: expose no user secrets.
- If `secret_names` is provided: expose only the named secrets.
- If a requested secret name does not exist: reject the request with a clear validation error.

Implementation approach:

- Filter by secret name once in `UserSecrets`, then derive env refs and mounted file items from the filtered subset.
- Preserve the current behaviour for Kerberos/VOMS/Rucio-enabled paths for now, i.e. do not scope secrets there yet.

## User-facing scope

Interactive sessions:

- `reana-client open ... --secret-name NAME` (repeatable)
- `reana-client open ... --no-secrets`
- `reana-client-go open ... --secret-name NAME` (repeatable)
- `reana-client-go open ... --no-secrets`

Workflow authoring:

- Serial: step-level `secret_names: ["alpha", "beta"]` in `reana.yaml`
- Dask: `workflow.resources.dask.secret_names: ["alpha", "beta"]` in `reana.yaml`
- Snakemake: rule-level `resources: secret_names="alpha,beta"`
- Yadage: step `environment.resources` entry such as `- secret_names: ["alpha", "beta"]`

## Out of scope for this issue

- CWL support
- Secret scoping for Kerberos/VOMS/Rucio-enabled jobs

These can be addressed in follow-up work once the base scoping model is agreed.

## Acceptance criteria

- Existing workflows and interactive-session flows remain unchanged when `secret_names` is not provided.
- Interactive sessions can request either an explicit secret allowlist or no secrets at all.
- Serial, Snakemake, Yadage, and Dask authors can specify the secret subset for their jobs/workers.
- Empty allowlists result in no secrets being exposed.
- Unknown secret names produce a clear user-facing error.
- Kerberos/VOMS/Rucio-enabled paths preserve current behaviour.

## Affected components

- `reana-commons`
- `reana-job-controller`
- `reana-workflow-controller`
- `reana-client`
- `reana-client-go`
- `reana-workflow-engine-serial`
- `reana-workflow-engine-snakemake`
- `reana-workflow-engine-yadage`
- `reana-server`

michaelbuchar:
Closes reanahub/reana#978

### DIFF
--- a/reana_client/cli/workflow.py
+++ b/reana_client/cli/workflow.py
@@ -1627,11 +1627,31 @@ def interactive_group():
     help="Docker image which will be used to spawn the interactive session. "
     "Overrides the default image for the selected type.",
 )
+@click.option(
+    "--secret-name",
+    "secret_names",
+    multiple=True,
+    help="Name of a user secret to expose in the interactive session. "
+    "Repeat the option to expose multiple secrets. If omitted, the session "
+    "inherits workflow.resources.secret_names when present; otherwise, all "
+    "user secrets are exposed.",
+)
+@click.option(
+    "--no-secrets",
+    is_flag=True,
+    help="Do not expose any user secrets in the interactive session.",
+)
 @add_access_token_options
 @check_connection
 @click.pass_context
 def workflow_open_interactive_session(
-    ctx, workflow, interactive_session_type, image, access_token
+    ctx,
+    workflow,
+    interactive_session_type,
+    image,
+    secret_names,
+    no_secrets,
+    access_token,
 ):  # noqa: D301
     """Open an interactive session inside the workspace.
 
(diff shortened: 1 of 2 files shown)


reana_client/cli/workflow.py
```diff
@@ -1683,11 +1683,31 @@ def interactive_group():
     help="Docker image which will be used to spawn the interactive session. "
     "Overrides the default image for the selected type.",
 )
+@click.option(
+    "--secret-name",
```
tiborsimko, line 1687: **🤖 PR778-01: The public CLI reference is missing the new options (LOW)**

The `--secret-name` and `--no-secrets` options are added to `reana-client open`, but `docs.reana.io/docs/reference/reana-client-cli-api/index.md` has not been regenerated and the `docs-cli-api` check fails on the resulting diff. There is no companion documentation change in this PR set, so merging the client as-is leaves both CI and the public command reference out of sync.

Suggested fix — regenerate and commit the CLI API reference in the documentation repository, or add the corresponding companion PR and link it here, then verify that `./run-tests.sh --docs-cli-api` is green.

PR #779, opened 2026-06-29
feat(validation): move workflow loading and validation to server

### LINKED ISSUE #8: Move workflow specification building and validation to the server
## Background

Today, validating a workflow specification requires the client to call the corresponding workflow engine (CWL, Serial, Snakemake, Yadage) to build the specification, then validate it locally. The server receives this specification and executes it as-is.

## Problem

This client-side approach has two consequences:

- Client language lock-in. Supporting a new client requires reimplementing every supported workflow engine in that language. This is why reana-client-go currently cannot create or submit workflows — porting all engines to Go is impractical.
- No security boundary. The server trusts a specification generated on the client without re-checking it. Client-side validation is a quality-of-life feature that catches mistakes early, not a guarantee the server can rely on.

## Proposal

Move specification building and validation to the server, so any client only needs to send a workflow snapshot (source files) and let the server do the rest.

Because building a specification means evaluating the workflow file, which can execute arbitrary code (e.g. Snakemake run: blocks, Jinja templating), this must happen inside a sandbox with least-privilege access — otherwise the server itself becomes remotely exploitable by a malicious specification.

## Acceptance Criteria

- [ ] Specification building and validation happen server-side, so a client only needs to submit workflow files, not run the corresponding engine
- [ ] Specification building and validation run in a sandboxed, least-privilege environment isolated from the rest of the server
- [ ] The sandbox has network access to fetch external dependencies referenced by the workflow (e.g. git+https CWL tools)

Sohn123:
Make the REANA server authoritative for workflow loading and validation.

Upload bounded specification bundles for create, validate, run, and
replacement restart operations.

Use generated Bravado operations and refuse servers that do not
advertise the specification-bundle protocol before upload.

Preserve legacy workflow-source declarations and the documented inline
serial create_workflow_from_json() API.

Reject unsafe paths, symbolic links, mutable ancestors, and oversized
snapshots.

Stream replacement specifications and workspace uploads.

Keep container registry checks local to explicit pull operations.

Closes reanahub/reana-workflow-validator#8

BREAKING CHANGE: workflow creation now uploads a raw specification
bundle instead of a client-serialized reana_specification JSON body.
The lower-level create_workflow(reana_specification, ...) Python API
is removed, and direct calls to the generated
client.api.create_workflow(reana_specification=...) operation must
migrate. Write a raw reana.yaml and call
create_workflow_from_bundle().


tiborsimko (commented):
**🤖 SNDBX-14: Orphaned validation modules and registry constants left behind after the client-side deletion (LOW)**

*Component:* reana-client — `reana_client/validation/parameters.py`,
`reana_client/validation/compute_backends.py`,
`tests/test_validate_parameters.py`, `tests/test_validate_compute_backends.py`,
`reana_client/config.py:40-57`

The PR correctly removed `reana_client/validation/server_capabilities.py` and its
test, but `reana_client/validation/parameters.py` (`validate_parameters`) and
`reana_client/validation/compute_backends.py` (`validate_compute_backends`) remain
with no production importer — `workflow_validate` no longer calls them (it uses
only `check_images_locally` and `validation/utils.py`). Their tests
(`tests/test_validate_parameters.py`, `tests/test_validate_compute_backends.py`)
remain, exercising dead code. Several registry constants in
`reana_client/config.py:40-57` (`ENVIRONMENT_IMAGE_SUSPECTED_TAGS_VALIDATOR`,
`DOCKER_REGISTRY_INDEX_URL`, `GITLAB_CERN_REGISTRY_INDEX_URL`/`_PREFIX`) backed the
now-removed local environment/registry checks and appear orphaned too.

Suggested fix — after a repo-wide grep to confirm no remaining importers, remove
`validation/parameters.py`, `validation/compute_backends.py`, their tests, and the
unused registry constants — matching how `server_capabilities` was handled.

reana_client/api/client.py
```diff
@@ -213,142 +205,268 @@ def get_workflow_status(workflow, access_token):
...
+    if not os.path.isfile(abs_path):
+        return
+    member = os.path.relpath(abs_path, base_dir)
+    if member in REANA_SPEC_FILENAMES:
+        # Keep the explicitly selected spec canonical; never clobber it.
+        return
+    members[member] = abs_path
+
+
+def _add_workflow_file_members(members, base_dir, workflow_file):
```
tiborsimko: **🤖 SNDBX-04: Client spec-bundle assembly has no path-escape or symlink containment (MEDIUM)**

*Component:* reana-client — `reana_client/api/client.py:249-258,261-291,369-374`;
reana-client-go — `pkg/specbundle/specbundle.go:112-127,137-168,180-200`

Both clients build the upload bundle from `workflow.file` (and
`inputs.parameters.input`) in the user's `reana.yaml` without validating that the
referenced paths stay inside the spec directory, and without skipping symlinks.

In `_add_workflow_file_members` (`reana_client/api/client.py:261-291`), a
`workflow.file` that escapes the base dir does not just read one file — it walks
the **parent tree**. For `workflow.file: ../secret.txt`, `workflow_dir` becomes
the parent of `base_dir`, the `same_dir or not isdir` guard is False, and control
falls into `os.walk(workflow_dir)`, bundling the whole parent directory.
`workflow.file: /etc/passwd` walks `/etc`; `workflow.file: /` walks the entire
filesystem. `_add_file_member` (`client.py:249-258`) gates on `os.path.isfile`,
which **follows symlinks**, so symlink targets are read and uploaded too. The
files are opened and sent at `client.py:369-374`. The Go client is an exact port
with the same behaviour: `os.Stat` follows symlinks (`specbundle.go:114`),
`filepath.Rel` is computed with no `..` check (`:118`), and
`filepath.Walk(workflowDir)` runs on the parent on escape (`:146`).

**Severity / threat model.** Under the normal model — a client talking to its own
trusted REANA server — this is MEDIUM, not HIGH. The server rejects any
`..`/absolute member name at staging *and* `rmtree`s the staging dir in `finally`
(`reana_server/rest/workflows.py:113-144`), and an escaping path necessarily
produces a `..`-prefixed member name, so escaping members are never persisted to
a workspace nor returned to the client — the "malicious spec author receives the
victim's file" path does not complete. What remains is still block-worthy: the
client **reads and transmits** out-of-project files (including symlink targets)
over the wire before rejection (a privacy/over-share leak, and a real exposure if
the server logs member content or an operator can read the transient upload
spool); `validate`, which users reasonably expect to be a *read-only local lint*,
silently uploads; and a crafted `workflow.file` can make the client walk an
enormous tree (parent dir, `/`) — a client-side hang/DoS footgun. It rises to HIGH
only when the client is pointed at an untrusted/multi-tenant server.

Suggested fix — validate bundle member paths **client-side before opening**, in
both clients: relative only, normalized, no `..` segment, not absolute, and no
symlink escape from the spec base directory (reuse `is_regular_path` in the Python
client, mirror it in Go). Consider a per-file/total size cap and an explicit error
on rejected members so the user understands why.

reana_client/api/client.py
```diff
@@ -213,142 +205,268 @@ def get_workflow_status(workflow, access_token):
...
+
+    A workflow file in a sub-directory (e.g. ``workflow/snakemake/Snakefile``)
+    typically imports co-located files (``include:``/``$import``/stage files), so
+    its whole sub-tree is bundled. A top-level workflow file is bundled on its
+    own — we do not fall back to gathering the entire spec directory.
+    """
+    abs_path = os.path.normpath(os.path.join(base_dir, workflow_file))
+    workflow_dir = os.path.dirname(abs_path)
+
+    same_dir = os.path.realpath(workflow_dir) == os.path.realpath(base_dir)
```
tiborsimko: **🤖 SNDBX-06: Top-level workflow file with relative includes is under-bundled — common layouts fail validation (MEDIUM)**

*Component:* reana-client — `reana_client/api/client.py:261-291`,
`tests/test_spec_bundle.py:83-92`; reana-client-go —
`pkg/specbundle/specbundle.go:133-144`, `pkg/specbundle/specbundle_test.go:138`

`_add_workflow_file_members` only walks a sub-tree when `workflow.file` lives in a
sub-directory; for a top-level workflow file it bundles the file alone:

```python
same_dir = os.path.realpath(workflow_dir) == os.path.realpath(base_dir)
if same_dir or not os.path.isdir(workflow_dir):
    _add_file_member(members, base_dir, workflow_file)
    return
```
(`reana_client/api/client.py:272-275`; the Go client mirrors this at
`specbundle.go:141-144`.)

So the very common Snakemake layout — a root-level `Snakefile` with
`include: "rules/common.smk"` — uploads only `Snakefile`, omitting
`rules/common.smk`. The same applies to a root-level `main.cwl` with
`$import: tools/tool.cwl`. The sandbox then cannot expand the DAG / resolve the
imports, so `reana-client validate` and `create` fail for a layout that worked
under the old client-side validation — while the sub-directory case
(`workflow/Snakefile`) bundles everything. The asymmetry is even asserted in
`tests/test_spec_bundle.py:83-92` (top-level file → only `{reana.yaml, Snakefile}`),
but no test covers a top-level file *with* an include, so the gap is invisible.

This was found independently by two reviewers, from opposite directions — a strong
signal it is real and will surprise users.

Suggested fix — fix the bundling rather than documenting a layout restriction:
for a top-level workflow file, still gather co-located includes (parse
`include:`/`$import` targets, or bundle sibling files referenced relatively). If a
restriction is unavoidable, surface it as an explicit client-side error naming the
missing referenced file, not a server-side load failure. Add a test for a
top-level workflow file with an include in both clients.

reana_client/api/client.py
```diff
@@ -213,142 +206,291 @@ def get_workflow_status(workflow, access_token):
...
+
+
+def _add_workflow_file_members(members, base_dir, workflow_file):
+    """Add ``workflow.file`` to the bundle, with its sub-tree if it lives in one.
+
+    A workflow file in a sub-directory (e.g. ``workflow/snakemake/Snakefile``)
+    typically imports co-located files (``include:``/``$import``/stage files), so
+    its whole sub-tree is bundled. A top-level workflow file is bundled on its
+    own — we do not fall back to gathering the entire spec directory.
+    """
```
michaelbuchar: This bundling rule looks too narrow for top-level workflows. When workflow.file lives at the bundle root, valid specs can still depend on sibling files, for example Snakemake include: "rules/common.smk" or CWL run: references next to the main workflow file. With the current behavior we only upload the root workflow file in that case, so validate/create regress on workflows that previously worked because the loader cannot find those sibling files in /validation/work/.... I reproduced this with a top-level Snakefile that includes rules/common.smk. Can we widen the bundle assembly here so root-level workflows also include the files they depend on, instead of treating “top-level” as “single-file”?

reana_client/cli/workflow.py
```diff
@@ -686,9 +694,13 @@ def workflow_restart(
...
-        )
+        # The server re-loads and validates the uploaded specification from the
+        # workspace at start (the binding gate), so a lightweight raw parse is
+        # enough here: the dict is only used locally afterwards to read the
+        # workflow type and input parameters for the input-parameter pre-check,
+        # and server-side for the restart type.
+        with open(specification_filename) as f:
+            parsed_parameters["reana_specification"] = yaml.safe_load(f)
         # upload new specification
         ctx.invoke(
```
michaelbuchar: This restart path still uploads only the replacement `reana.yaml`, but after the server-side validation change the server reloads non-serial replacement specs from the workspace, not from the JSON payload alone. That means a restart whose new spec points to a different workflow file, or changes sibling files next to the workflow, can fail or silently reuse stale workspace contents because the rest of the raw bundle was never updated. I reproduced this with a Snakemake restart whose replacement spec referenced `Snakefile2`; `/start` failed because `Snakefile2` was missing from `/validation/work`. Can we upload the full replacement spec bundle here, using the same bundle-scoping logic as `create`/`validate`, instead of uploading only the YAML?

reana_client/api/client.py
Sohn123: Hey, I think this is a design choice we need to make if all files that are used for workflow creation should be part of the input definition or if we apply smart globbing etc. I personally think it is nicer to make it an explicit choice which files should be included. Do our input files of a specification already support globbing? But if we think that is nicer to try to be smart, we can also do that.

reana_client/cli/workflow.py
Sohn123: Ok here the question for me is what is the semantic of a restart? Because isn't the point of a restart kind of to reuse the information already present in the workspace? Because what else would be the difference to just starting another workflow with the same name? I agree it might be confusing, but I think just uploading and overwriting/selectively overwriting would also be weird.

reana_client/api/client.py
Sohn123: See also: https://github.com/reanahub/reana-client-go/pull/185#discussion_r3499721231

reana_client/api/client.py
Sohn123: Is fixed in the latest version.

reana_client/api/client.py
Sohn123: We decided that this is the users task to include all relevant files as input files, because we don't want to upload too many files to the server. The validation should fail with an error clearly stating that a file is missing for a successful validation.

reana_client/api/client.py
Sohn123: Ok in an offline discussion we decided that we want users to include all snakemake files etc. by hand in the input files and don't try to be smart about globbing/including.

reana_client/cli/workflow.py
Sohn123: We discussed that this is intended behavior. A restart should not change the workspace files, it should just replace the reana.yaml if specified. Updating other files etc. is the users responsibility for a successful run/validation.

reana_client/cli/cwl_runner.py
```diff
@@ -99,6 +99,60 @@ def upload_files(files, basedir, workflow_id, access_token):
...
+    if jobfile:
+        with open(jobfile) as f:
+            job = yaml.load(f, Loader=yaml.FullLoader)
+        job_rel = os.path.relpath(jobfile, basedir)
+
+    cwl_dependencies = findfiles([get_file_dependencies_obj(processfile, basedir)])
+    bundle_dir = tempfile.mkdtemp()
+    try:
+
+        def stage(location):
```
tiborsimko: **🤖 PR779-01: CWL bundle staging can write outside its temporary directory (MEDIUM)**

The new CWL runner builds its server-side-validation bundle with a separate
`stage` helper that does not use the containment checks added to the main bundle
assembler:

```python
def stage(location):
    source = os.path.join(basedir, location)
    if not os.path.isfile(source):
        return
    destination = os.path.join(bundle_dir, location)
    os.makedirs(os.path.dirname(destination) or bundle_dir, exist_ok=True)
    shutil.copyfile(source, destination)
```

`workflow_rel` and `job_rel` come from `os.path.relpath`, and dependency
locations come from the CWL document. A `../` location escapes `bundle_dir` and
can copy to or overwrite a path outside `mkdtemp()` before the `finally` block
removes only the intended bundle directory. An absolute location replaces both
the source and destination prefixes in `os.path.join`, making them the same file
and crashing with `SameFileError`; symlink sources are followed. Even benign
workflows whose process/job sits outside a user-supplied `--basedir` are staged
outside the bundle and then fail validation because the referenced file was
never uploaded. The adjacent job-file parse also uses `yaml.FullLoader` instead
of the project's `safe_load` convention for untrusted specification inputs.

Suggested fix — route every CWL member through the same realpath, symlink, and
regular-file containment rules as `_gather_spec_members`: reject absolute and
escaping paths before opening them, verify both source containment under
`basedir` and destination containment under `bundle_dir`, and surface a clear
error rather than silently skipping. Add regression tests for absolute,
`../`-escaping, and symlinked process/job/dependency paths; use `yaml.safe_load`
for the job file.

reana_client/cli/workflow.py
tiborsimko: The restart `-f` semantics discussion has one remaining residual, tracked as [PR794-03 on reana-server#794](https://github.com/reanahub/reana-server/pull/794#discussion_r3590823593): a non-serial replacement uploaded under a non-canonical filename can be silently ignored in favour of the existing `reana.yaml`.

reana_client/cli/cwl_runner.py
Sohn123: Fixed in the latest version.

reana_client/api/client.py
```diff
@@ -213,142 +207,401 @@ def get_workflow_status(workflow, access_token):
...
+    base_dir = os.path.dirname(os.path.abspath(reana_file))
+    reana_file = os.path.abspath(reana_file)
+
+    # Preserve compatibility with callers passing ``-f myreana.yaml``: the
+    # server still receives it under the canonical top-level name it searches.
+    members = {"reana.yaml": reana_file}
+
+    spec = _load_spec_mapping(reana_file)
+    workflow = spec.get("workflow") or {}
```
tiborsimko: **🤖 PR779-02: Wrong-shaped YAML crashes bundle discovery before server validation (LOW)**

`_load_spec_mapping` deliberately treats parsing as best-effort so malformed
specifications can still be uploaded for the authoritative server-side error.
That contract is only enforced for invalid YAML and a non-mapping top level.
Syntactically valid nested values are used without type checks:

```python
workflow = spec.get("workflow") or {}
inputs = spec.get("inputs") or {}
workflow_file = workflow.get("file")
...
parameters = inputs.get("parameters") or {}
param_input = parameters.get("input")
```

Consequently `workflow: [value]`, `inputs: [value]`, or a non-mapping
`inputs.parameters` raises `AttributeError` in the client. The canonical
`reana.yaml` is never uploaded and the user does not receive the structured
server validation report that the best-effort parser is intended to preserve.

Suggested fix — guard `workflow`, `inputs`, and `inputs.parameters` with
`isinstance(value, dict)` before inspecting bundle hints. For every wrong shape,
fall back to the canonical `reana.yaml` member only and let the server report
the schema error. Add tests for list, scalar, and string values at each nested
level.

reana_client/api/client.py
```diff
@@ -213,142 +207,407 @@ def get_workflow_status(workflow, access_token):
...
+
+    open_files = {}
+    try:
+        for member, local_path in members.items():
+            if not os.path.isfile(local_path):
+                raise FileUploadError(
+                    "Specification bundle file not found: {}".format(member)
+                )
+            open_files[member] = open(local_path, "rb")
+        return requests.post(
```
tiborsimko: **🤖 PR779-03: Python bundle upload buffers the whole body in memory and sets no timeout (LOW)**

The Python client posts the specification bundle by handing open file objects to
`requests`:

```python
def _post_spec_members(path, members, params):
    ...
    open_files[member] = open(local_path, "rb")
    return requests.post(
        urljoin(get_api_url(), path),
        params=params,
        files=open_files,
        verify=False,
    )
```

`requests.post(files=...)` does not stream: it reads each file with `fp.read()`
and encodes the complete multipart body in memory before the request is sent
(`urllib3.filepost.encode_multipart_formdata`). At the supported 100 MiB bundle
size the client therefore holds the whole encoded body at once. The call also
passes no `timeout=`, so a stalled or half-open connection hangs the command
indefinitely.

This is the Python counterpart of the Go client's PR185-01 (same buffering, same
missing timeout); the two should be fixed together for parity.

Suggested fix — set an explicit `timeout=` on the bundle `requests.post`, and
avoid buffering the whole body in memory — stream it with a multipart encoder
that reads from the file handles lazily (for example
`requests_toolbelt.MultipartEncoder`, if it can be added as a dependency) or
chunk the upload. At minimum add the timeout.

reana_client/api/client.py
tiborsimko: **🤖 PR779-02 follow-up (2026-07-27): verified fixed**

*Component:* reana-client — `reana_client/api/client.py` (`_gather_spec_members`)

The best-effort bundle discovery now guards each nested field before reading
bundle hints:

```python
workflow = spec.get("workflow") or {}
if not isinstance(workflow, dict):
    workflow = {}
inputs = spec.get("inputs") or {}
if not isinstance(inputs, dict):
    inputs = {}
...
parameters = inputs.get("parameters") or {}
if not isinstance(parameters, dict):
    parameters = {}
```

Wrong-shaped `workflow`/`inputs`/`inputs.parameters` (list, scalar, string) no
longer raise `AttributeError`; the client falls back to bundling `reana.yaml`
alone so the canonical upload and the authoritative server-side report are
preserved. `tests/test_spec_bundle.py::test_gather_spec_members_ignores_wrong_shaped_bundle_hints`
parametrises list/scalar values at each nested level and asserts only
`reana.yaml` is bundled.

Verified fixed at `68c66b1`. No residual.

Sohn123: Confirmed fixed. Nested `workflow`, `inputs`, and `inputs.parameters` values are type-checked; invalid shapes fall back to uploading only canonical `reana.yaml` so the server returns the structured validation result. Regression tests cover wrong-shaped values at each level.

reana_client/api/client.py
Sohn123: Partially addressed. The custom `requests.post` path is gone; archive creation is disk-backed and capped, and the generated Bravado operation uses 10-second connect and 300-second request timeouts. Bravado's Requests transport may still buffer multipart encoding in memory, so this residual remains open.

tiborsimko: **🤖 PR779-03 follow-up (2026-07-29): verified fully fixed at the current heads**

The earlier developer reply correctly noted that the custom `requests.post`
path had been removed and the archive was disk-backed, byte/file capped, and
given explicit connect/request timeouts, but it left Bravado multipart buffering
as an open residual.

That residual is now fixed by the accompanying reana-commons changes.
`StreamingRequestsClient` is installed as Bravado's HTTP transport, replaces the
eager `files=` encoder with `StreamingMultipartBody`, precomputes the exact
`Content-Length`, and yields file parts in 1 MiB reads. The test reader rejects
any unbounded `read()`, confirming that the encoder consumes the archive in
bounded chunks. `_post_spec_members()` passes the disk-backed temporary archive
through this generated operation with a ten-second connect and 300-second
request timeout.

Verified against reana-client `fbed110` and reana-commons `a6d1254`. PR779-03 is
fully addressed; no upload-buffering or missing-timeout residual remains.

reana_client/cli/workflow.py
```diff
@@ -580,6 +602,7 @@ def display_status(workflow: str, current_status: str):
         try:
             logging.info("Connecting to {0}".format(get_api_url()))
             response = start_workflow(workflow, access_token, parsed_parameters)
+            _display_validation_warnings(response)
```
tiborsimko, line 604: **🤖 PR779-04: Start and restart requests can wait indefinitely during validation (LOW)**

Bundle creation/validation and workspace uploads now pass explicit ten-second
connect and 300-second response timeouts, but `start_workflow()` invokes the
generated operation without `_request_options` and calls `.result()` without a
timeout. Bravado therefore passes `timeout=None` to Requests.

Start and restart now synchronously reload and validate the workspace, which
can legitimately take close to the sandbox deadline. A network, server, or
controller stall can consequently leave both CLI commands blocked forever.
Calling generated operations without timeouts is a pre-existing client pattern,
not newly introduced by this PR. The earlier PR779-03 upload fix is sound; this
PR makes the existing start path materially slower by adding synchronous
validation, which is why the missing timeout matters to this change.

Suggested fix — define bounded control-request options and pass them to
`start_workflow()`, with enough read time for the server and sandbox validation
budgets. Add tests asserting both the connect and response timeout values for
ordinary start and restart.

reana_client/validation/environments.py
```diff
@@ -6,534 +6,154 @@
...
+    # authority on whether the image is usable.
+    subprocess.run(
+        [cli, "pull", image],
+        capture_output=True,
+        text=True,
+        timeout=timeout,
+    )
+    # Override the entrypoint so the image's own code does not run -- only ``id``.
+    inspect = subprocess.run(
+        [cli, "run", "--rm", "--entrypoint", "/bin/sh", image, "-c", "id -u && id -G"],
```
tiborsimko: **🤖 PR779-05: Local image inspection runs with network access and no resource limits (LOW)**

The `--pull` environment check runs `/bin/sh -c "id -u && id -G"` from the
inspected image with Docker/Podman's default network, capability set, PID
allowance, writable filesystem, and CPU/memory limits. Overriding ENTRYPOINT
does not prevent image code from running: the shell, `id` executable, dynamic
loader, and libraries are supplied by the untrusted image.

Container namespaces, the default seccomp profile, and Docker/Podman's reduced
default capability set still provide meaningful isolation. However, a
malicious image can execute arbitrary image-controlled code with access to
services reachable from the developer machine, or consume unbounded local
resources. When `subprocess.run()` times out it terminates the container-engine
client, but does not guarantee that the daemon-managed container stops; `--rm`
only takes effect after the container exits. The command is opt-in and this
execution behaviour predates the PR, although the amended code retains it and
adds the inaccurate claim that the image's own code does not run. The Go client
has the same issue in PR185-06, where the implementation is new.

Suggested fix — inspect a created-but-stopped container or exported filesystem
without executing image binaries. If execution is retained, add at least
`--network none`, a read-only root, dropped capabilities, no-new-privileges,
PID/CPU/memory limits, and a unique container/CID that is forcibly removed in
a `finally` block. Document the execution explicitly and test both the
hardening flags and timeout cleanup.

reana_client/cli/workflow.py
```diff
@@ -684,29 +718,28 @@ def workflow_restart(
...
-            filenames=(specification_filename,),
-            access_token=access_token,
-        )
+        # The server re-loads and validates the uploaded specification from the
+        # workspace at start (the binding gate), so a lightweight raw parse is
+        # enough here: the dict is only used locally afterwards to read the
+        # workflow type and input parameters for the input-parameter pre-check.
+        # The server binds and loads the raw uploaded file using its digest.
+        with open(specification_filename) as f:
+            replacement_specification = yaml.safe_load(f)
```
tiborsimko: **🤖 PR779-06: Restart reads the replacement outside the bundle byte budget (LOW)**

The `restart -f` path parses the complete replacement before checking whether
the parsed object is needed:

```python
with open(specification_filename) as f:
    replacement_specification = yaml.safe_load(f)
```

The object is only consumed inside `if parameters or options`, but an ordinary
replacement restart with no overrides still loads the whole file. This bypasses
the configured `REANA_SPEC_BUNDLE_MAX_BYTES` budget and the descriptor-based
regular-file checks used by specification bundle creation. The subsequent
SHA-256 loop is streaming, but it likewise has no byte ceiling, and the full
file is then uploaded before the server-side bundle limit rejects it.

The file is explicitly selected by the local user, so this is primarily a
robustness problem: an accidentally oversized replacement can consume
unbounded client memory and I/O before the authoritative server validation has
a chance to reject it.

Suggested fix — skip YAML parsing entirely when no parameter or option
overrides were supplied. When local override validation does need the object,
use the bounded `reana_commons.specification_paths.load_raw_reana_spec()`
helper. Enforce the same byte ceiling before hashing and uploading, and add
restart tests for an oversized replacement both with and without overrides.

reana_client/cli/workflow.py
tiborsimko: **🤖 PR779-04 follow-up (2026-08-07): verified fixed at 228fbb6**

`start_workflow()` now passes `_CONTROL_REQUEST_OPTIONS` to the generated
operation. The shared options provide a ten-second connect timeout and a
300-second response timeout, which covers both ordinary start and restart
because both command paths use this helper. The timeout is long enough for the
server's synchronous sandbox validation budget while preventing an indefinite
network wait.

The original finding is fully addressed.

reana_client/validation/environments.py
Sohn123: This is accepted intended behavior for now. `--pull` is an explicit opt-in runtime compatibility probe for an image the user intends to execute on REANA; it deliberately runs the image-provided shell and `id` binaries to observe runtime UID/supplementary groups. Static metadata inspection would not exercise that same runtime path. I corrected the misleading comment that claimed no image code runs, but I am not changing the execution model or adding local sandbox flags in this PR. The residual local-engine risk is therefore acknowledged and accepted.

reana_client/cli/workflow.py
Sohn123: Replacement YAML is not parsed at all when there are no parameter/option overrides. When local override validation needs the mapping, it now uses `reana_commons.specification_paths.load_raw_reana_spec()`, which performs descriptor-anchored regular-file checks and enforces the shared byte ceiling before bounded parsing. The generated multipart helper independently checks the securely opened file size before upload. The new regression test verifies that a bounded-loader rejection prevents the restart request.

tiborsimko: **🤖 PR779-06 follow-up (2026-08-10): verified fixed at ea094f4**

The restart path no longer parses replacement YAML when no parameter or
operational-option overrides were supplied. When local override validation
does need the mapping, it uses the shared descriptor-anchored,
byte-bounded `load_raw_reana_spec()` helper.

The multipart API helper also securely opens the replacement and checks its
size before upload, and the regression test verifies that a bounded-loader
failure prevents the restart request. This addresses the original unbounded
parse/hash/upload path.

reana_client/validation/environments.py
tiborsimko: **🤖 PR779-05 follow-up (2026-08-11): intended runtime probe documented; timeout cleanup remains (LOW remains)**

The amended comments now accurately state that `--pull` deliberately executes
the image-provided shell and `id` binaries. The explanation that this is an
explicit opt-in runtime compatibility probe, rather than static metadata
inspection, is a coherent product decision. The associated local-engine risk
is acknowledged, so this follow-up does not ask to change that execution model
in this PR.

One independent cleanup problem remains. `subprocess.run(..., timeout=...)`
terminates the Docker/Podman client process on timeout, but the engine-managed
container can continue running under the Docker daemon or Podman's `conmon`.
`--rm` removes it only after the container exits normally, and the command does
not assign a unique name or CID that a `finally` block could forcibly remove.
An image whose shell never terminates can therefore leave a live container on
the developer's machine after the client reports that inspection timed out.

Suggested fix — assign each inspection a collision-resistant container name
or CID file and always perform a bounded `docker`/`podman rm -f` cleanup after
timeout or other failure. Limit cleanup to the identifier created by this
invocation, tolerate already-removed containers, and add a regression test
that simulates timeout and asserts forced engine-side removal.

Sohn123: Fixed while retaining the accepted opt-in runtime probe behavior. Each inspection now receives a collision-resistant container name and performs a bounded docker/podman rm -f for that exact name in finally, including timeout and failure paths. A timeout regression test verifies the forced cleanup.

reana_client/api/client.py
```diff
@@ -213,142 +301,144 @@ def get_workflow_status(workflow, access_token):
...
+        archive.seek(0)
+        operation = getattr(current_rs_api_client.api, operation_id)
+        return operation(
+            bundle=("validation-bundle.zip", archive),
+            _request_options=_TRANSFER_REQUEST_OPTIONS,
+            **params,
+        ).result()
+
+
+def create_workflow_from_bundle(reana_file, name, access_token):
```
tiborsimko, line 451: **🤖 PR779-09: The advertised Python workflow-creation API is removed without a compatibility shim (MEDIUM)**

This PR removes `create_workflow_from_json()` from the public
`reana_client.api.client` module and replaces workflow creation with
disk-backed bundle functions. The [official Python API
guide](https://blog.reana.io/posts/2021/reana-client-python-api/) explicitly
teaches users to import `create_workflow_from_json()` and submit an in-memory
serial workflow, and the changelog introduced it specifically for developers
and third-party systems. Existing programs now fail at import time, while the
replacement requires a specification file on disk and therefore has no
equivalent for the documented use case. The CLI and the other seven functions
used by that guide remain available, so the regression is narrow but affects
the programmatic creation entry point around which the example is built.

This is independent of client/server version negotiation: it also breaks when
the new client talks to a matching new server, before either side can advertise
or inspect a capability. The legacy signature additionally accepts
`workflow_file` (which overrides `workflow_json`), `workspace_path`, parameters,
outputs, and a `workflow_engine` whose default is `yadage`, so silently mapping
only the dictionary argument would change other established call patterns. The
lower-level `create_workflow()` symbol is removed as well, but its contract is
to submit an already client-serialised specification; that may reasonably be
documented as an intentional break rather than retained across this new trust
boundary.

Suggested fix — keep `create_workflow_from_json()` importable as a compatibility
shim.
At minimum, preserve the documented inline-serial case by writing a temporary
raw specification and submitting it through `create_workflow_from_bundle()`,
so the server still performs the authoritative validation. This case needs
only the specification in its validation bundle: an inline serial workflow has
no `workflow.file`, so legacy `inputs.files` entries stay outside the validation
scope and are not treated as loader dependencies that must exist beside the
temporary file.

Only translate `workflow_file` calls when their raw source scope can be
constructed without guessing. Most non-serial workflows can reference
additional CWL, Yadage, or Snakemake files that a single path does not identify;
for these and for pre-expanded specifications, fail explicitly with migration
guidance naming the relevant `workflow_engine`, `workflow_file`, and
`workflow.files`/`workflow.directories` requirements rather than restoring the
client-trusted specification boundary. The shim can emit a deprecation warning
while callers migrate. Please cover the retained public import, the documented
serial call, the default-Yadage call, and `workflow_file` overriding
`workflow_json` with compatibility tests.

Sohn123: `create_workflow_from_json()` is importable again from
`reana_client.api.client`, with its historical signature, and the documented
inline-serial case works end to end against the authoritative bundle path.

**What is translated.** `workflow_engine="serial"` with an inline
`workflow_json` and no `workflow_file`. The shim builds the raw mapping
(`workflow.type`, `workflow.specification`, optional `inputs`/`outputs` from the
historical `parameters`/`outputs` arguments), normalises it through
`json.dumps(..., sort_keys=True)` — the removed implementation round-tripped the
same way, and it is required because `workflow_json` comes from arbitrary caller
code that `yaml.safe_dump` cannot always represent — writes it as `reana.yaml` in
a temporary directory, and delegates to `create_workflow_from_bundle()`. The
capability check and the authoritative server-side validation then apply
unchanged. The temporary directory is removed on success and on failure.

**Why serial only is a constraint, not a preference.** Serial is the only engine
whose loader accepts an inline `workflow.specification`:
`reana_commons/specification.py::load_workflow_spec_from_reana_yaml` passes
`specification=` to the loader for `serial` alone, and every other engine is
loaded from `workflow.file` on disk. A single in-memory dictionary cannot supply
that, so a translated Yadage/CWL/Snakemake call could not be re-derived
server-side under the new trust boundary.

**Confirmed: legacy `inputs.files` stay outside the validation scope.** As the
review notes, an inline serial workflow has no `workflow.file`, and
`reana_commons/specification_paths.py::uses_legacy_validation_scope` requires
`workflow.file is not None`, so `inputs.files` are not gathered as loader
dependencies. There is a test that drives the real gather/archive path with an
`inputs.files` entry naming a file that does not exist beside the temporary
specification: the bundle contains exactly `reana.yaml` and the call succeeds.

**What fails, with guidance.** Non-serial engines, any `workflow_file`, and a
missing `workflow_json` raise `ValueError` naming the supplied
`workflow_engine`/`workflow_file` and explaining the migration: write a raw
`reana.yaml`, declare loader dependencies through
`workflow.files`/`workflow.directories`, and call `create_workflow_from_bundle()`.
`workflow_file` keeps its historical precedence over `workflow_json` when
selecting which error to report, so a legacy call fails explicitly instead of
silently submitting the dictionary. `workspace_path` is accepted for signature
compatibility and is not used to guess files. Name, access-token, server-URL and
engine validation are preserved, and a `DeprecationWarning` with `stacklevel=2`
points callers at the raw specification plus `create_workflow_from_bundle()`.

One deliberate note: the original `parameters`-selects-an-external-parameter-file
failure mode is not implemented, because it is unreachable here.
`workflow_parameter_file()` honours the legacy `inputs.parameters.input` spelling
only for `PARAMETER_FILE_WORKFLOW_TYPES = {"cwl", "snakemake"}`, and
`workflow.parameters.file` is rejected outright for other types — so a serial
workflow cannot select one.

**`create_workflow()` is intentionally not restored.** Its contract is to submit
an already client-serialized specification, which conflicts directly with the new
trust boundary. This is recorded as a `BREAKING CHANGE:` trailer in the commit
body (these repositories generate `CHANGELOG.md` from conventional commits).

**Tests** (`tests/test_api_create_workflow_from_json.py`): the public import; the
documented serial call producing the expected raw mapping and delegating once;
parameters/outputs/name/token preserved; the real gather/archive path uploading
only `reana.yaml`; the default Yadage call failing with migration guidance;
`workflow_file` overriding `workflow_json` for error selection; non-serial engines
and an unknown engine failing before contacting the server; missing token and
UUIDv4 name still rejected; the deprecation warning at the caller's stack level;
temporary files removed on success and failure; non-JSON-native values normalised.

Verified live against a cluster: the documented call creates `confirm-json.1` and
emits a `DeprecationWarning`.

tiborsimko: **🤖 PR779-09 follow-up (2026-08-17): compatibility shim verified fixed (MEDIUM)**

Verified on current head `6654eb3`. `create_workflow_from_json()` again accepts
the advertised legacy Python call shape, writes a temporary canonical
`reana.yaml`, and delegates to the secure bundle-creation path without
reintroducing client-side engine loading.

The compatibility boundaries are sound: only serial workflows receive an
inline `specification` argument; an inline serial specification does not select
legacy input files because it has no `workflow.file`; and parameter-file
loading is limited to CWL and Snakemake. The shim therefore preserves the
removed serial API without accidentally widening its upload scope. This fully
addresses PR779-09.

tiborsimko (approved):

PR #781, opened 2026-08-10
build(python): add support for Python 3.15

tiborsimko:


CameronMcClymont (approved):
Looks good, we just need to remember to change `3.15-dev` to `3.15` when it comes out and remove `allow-prereleases: true` again.

Issue #557, opened 2021-09-07
cli: cannot stop queued workflow


alintulu:
## Current behaviour

User is unable to stop a workflow that is queuing. The error message simply says "the workflow is not running" (see below for full example).

```
$ reana-client run -w workflow
$ reana-client list
NAME                 RUN_NUMBER   CREATED               STARTED               ENDED                 STATUS
workflow             1            2021-09-07T12:34:16   -                     -                     queued
$ reana-client stop -w workflow
Graceful stop not implement yet. If you really want to stop your workflow without waiting for jobs to finish use: 
--force option
Aborted!
$ reana-client stop -w workflow --force
Cannot stop workflow workflow: 
Workflow 8dabbd83-04ec-4cb0-a48d-56d07e1d8869 is not running.
```

The user might realise that the submitted workflow is wrong. If so, the user has to wait until the workflow is running before being able to stop it (currently there is no way to know when the workflow starts to run).

## Expected behaviour

The command `reana-client stop` should also stop workflows in the queue.

mvidalgarcia:
You can delete it:

```bash
# pip install reana-client==0.7.5
$ reana-client delete -w workflow --include-workspace --include-records
```

Does it have the same behaviour you're looking for?

alintulu:
Ah, you're right! Thanks, that solves it :)

PR #783, opened 2026-08-19
feat(logs): report pruned workflow logs

### LINKED ISSUE #964: introduce system-wide retention policy for workflow engine, service and job logs
### Current behaviour

REANA preserves user runtime workflow logs forever. Every workflow run
accumulates:

- **Engine logs** — the workflow engine pod that orchestrates the run (Serial /
  CWL / Yadage / Snakemake).
- **Service logs** — supporting service pods (e.g. Dask) that run alongside the
  workflow.
- **Job logs** — per-step user job stdout/stderr.

Because there is no automated eviction, the total log volume may grow
unboundedly with the number of workflow runs. On busy instances this is
undesirable: logs are usually only consulted in the days immediately following a
run. They do not need to occupy the database space for the lifetime of the
deployment.

### Expected behaviour

Introduce a system-wide log retention policy. The operator-facing surface
mirrors the existing **workspace retention** feature
(`workspaces.retention_rules.*`) so that administrators encounter a familiar
shape, but the underlying mechanism is deliberately simpler: there is no
per-workflow log retention rule, only a single cluster-wide period.

```yaml
logs:
  retention_rules:
    maximum_period: forever # e.g. "60" days; "forever" preserves current behaviour
    cronjob_schedule: "0 3 * * *" # everyday at 3am (staggered from workspace cleanup)
```

After the configured retention period (e.g. 60 days), a periodic cronjob would
delete:

- engine logs
- service logs
- job logs

for all workflow runs older than the threshold, all the while preserving the
workflow record itself, its UUID, parameters, status (`finished`, `failed`,
`stopped`), start/end timestamps, etc.

GitHub Actions offers a good UX precedent here. Past runs remain listed forever
in the Actions tab with their full status, but opening an old run's logs
displays:

> The logs for this run have expired and are no longer available.

We should do the same in REANA. When a user requests logs for an expired run via
`reana-client logs`, the REANA UI, or the REST API, they should see a clear,
friendly message explaining that the logs have been pruned by the cluster's
retention policy.

#### Suggested scope

- **helm**: add `logs.retention_rules.maximum_period` (default `forever`) and
  `logs.retention_rules.cronjob_schedule`, mirroring the existing
  `workspaces.retention_rules.*` values. Render the period into the reana-server
  pod environment via
  `{{ tpl .Values.logs.retention_rules.maximum_period . | quote }}`, exactly as
  is already done for workspaces.
- **reana-db**: add a single `logs_pruned_at` timestamp column on the workflow
  model, set by the cronjob when it actually deletes a run's logs. No separate
  rule table and no `created → active → pending → applied` state machine: a
  cluster-wide policy has no per-workflow lifecycle to track. Storing the
  exact timestamp gives a clear audit trail and lets the UI tell users
  precisely when their logs were pruned.
- **reana-server**: add a cronjob that walks workflows older than the
  retention threshold whose `logs_pruned_at` is `NULL`, deletes their
  associated engine/service/job logs, and stamps `logs_pruned_at = now()`.
- **reana-ui** and **reana-client**: when fetching logs of a workflow whose
  `logs_pruned_at` is set, surface a clear "the logs for this run were pruned
  by the cluster's retention policy on `<date>` and are no longer available"
  message.

#### Notes

- The default of `forever` ensures this is a strictly opt-in behavioural change
  for existing deployments.
- Composes naturally with the workspace data retention feature: a workflow may
  have its logs pruned but its workspace kept, or vice versa.

michaelbuchar:
Closes reanahub/reana#964

### DIFF
--- a/reana_client/api/client.py
+++ b/reana_client/api/client.py
@@ -23,7 +23,11 @@ import yaml
 from bravado.exception import BravadoConnectionError, BravadoTimeoutError, HTTPError
 from reana_client.api.utils import get_content_disposition_filename
 from reana_client.config import ERROR_MESSAGES
-from reana_client.errors import FileDeletionError, FileUploadError
+from reana_client.errors import (
+    FileDeletionError,
+    FileUploadError,
+    WorkflowLogsPrunedError,
+)
 from reana_client.utils import is_regular_path, is_uuid_v4
 from reana_commons.api_client import get_current_api_client
 from reana_commons.config import (
@@ -786,14 +790,19 @@ def get_workflow_logs(workflow, access_token, steps=None, page=None, size=None):
             )
 
     except HTTPError as e:
+        error_data = e.response.json()
         logging.debug(
             "Workflow logs could not be retrieved: "
             "\nStatus: {}\nReason: {}\n"
             "Message: {}".format(
-                e.response.status_code, e.response.reason, e.response.json()["message"]
+                e.response.status_code, e.response.reason, error_data["message"]
             )
         )
-        raise Exception(e.response.json()["message"])
+        if e.response.status_code == 410:
+            raise WorkflowLogsPrunedError(
+                error_data["message"], error_data.get("logs_pruned_at")
+            )
+        raise Exception(error_data["message"])
     except Exception as e:
         raise e
 
--- a/reana_client/cli/workflow.py
+++ b/reana_client/cli/workflow.py
@@ -45,6 +45,7 @@ from reana_client.config import (
     CLI_WORKFLOWS_LIST_MAX_RESULTS,
     JSON,
 )
+from reana_client.errors import WorkflowLogsPrunedError
 from reana_client.printer import display_message
 from reana_client.utils import (
     get_reana_yaml_file_path,
@@ -981,6 +982,7 @@ def workflow_logs(
             "Ignoring --json as it cannot be used together with --follow.",
             msg_type="warning",
         )
+        json_format = False
 
     available_filters = {
         "step": "job_name",
@@ -1049,6 +1051,20 @@ def workflow_logs(
                 page,
                 size,
             )
+    except WorkflowLogsPrunedError as e:
+        if json_format:
+            click.echo(
+                json.dumps(
+                    {
+                        "message": str(e),
+                        "logs_pruned_at": e.logs_pruned_at,
+                    },
+                    indent=2,
+                )
+            )
+        else:
+            display_message(str(e), msg_type="error")
+        sys.exit(1)
     except Exception as e:
         logging.debug(traceback.format_exc())
         logging.debug(str(e))
--- a/reana_client/errors.py
+++ b/reana_client/errors.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of REANA.
-# Copyright (C) 2018, 2021, 2022 CERN.
+# Copyright (C) 2018, 2021, 2022, 2026 CERN.
 #
 # REANA is free software; you can redistribute it and/or modify it
 # under the terms of the MIT License; see LICENSE file for more details.
@@ -19,3 +19,12 @@ class FileDeletionError(Exception):
 
 class EnvironmentValidationError(Exception):
     """REANA workflow environment validation didn't succeed."""
+
+
+class WorkflowLogsPrunedError(Exception):
+    """Workflow logs were removed by the cluster retention policy."""
+
+    def __init__(self, message, logs_pruned_at=None):
+        """Store the server message and optional pruning timestamp."""
+        super().__init__(message)
+        self.logs_pruned_at = logs_pruned_at
--- a/tests/test_cli_workflows.py
+++ b/tests/test_cli_workflows.py
@@ -15,10 +15,12 @@ from typing import List
 import pytest
 import yaml
 from click.testing import CliRunner
-from mock import Mock, patch
+from bravado.exception import HTTPError
+from mock import MagicMock, Mock, patch
 from reana_commons.testing import make_mock_api_client
 from reana_client.cli import cli
 from reana_client.config import RUN_STATUSES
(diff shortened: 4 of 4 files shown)


reana_client/api/client.py
```diff
@@ -786,14 +790,19 @@ def get_workflow_logs(workflow, access_token, steps=None, page=None, size=None):
...
         logging.debug(
             "Workflow logs could not be retrieved: "
             "\nStatus: {}\nReason: {}\n"
             "Message: {}".format(
-                e.response.status_code, e.response.reason, e.response.json()["message"]
+                e.response.status_code, e.response.reason, error_data["message"]
             )
         )
-        raise Exception(e.response.json()["message"])
+        if e.response.status_code == 410:
```
tiborsimko, line 801: **🤖 PR783-01: Permitted OpenAPI bundles cannot decode the new HTTP 410 response (MEDIUM)**

The new handler assumes that a 410 response reaches this `HTTPError` branch:

```python
if e.response.status_code == 410:
    raise WorkflowLogsPrunedError(
        error_data["message"], error_data.get("logs_pruned_at")
    )
```

However, `setup.py` still permits `reana-commons>=0.95.0a22`, and that released
version's bundled reana-server specification does not declare the 410 response.
Bravado validates responses against that bundled specification before returning
control to this code. With the permitted version, it raises
`MatchingResponseNotFound` during response unmarshalling, so the friendly
`WorkflowLogsPrunedError` path is never reached. `MatchingResponseNotFound`
propagates out of `get_workflow_logs` and is rendered by the command's generic
handler as `Cannot retrieve logs for workflow <name>: ...`, surfacing a
client-internal schema error instead of the retention message.

The current test constructs `HTTPError` directly, which bypasses the generated
client and therefore cannot detect this compatibility failure.

Suggested fix — release the updated reana-commons specification first, then
bump both reana-commons lower bounds in `setup.py` to that release and regenerate
any dependency artefacts used by CI. Add coverage through an API client built
from the bundled specification, or otherwise assert that the installed
specification declares and successfully unmarshals the 410 response.

reana_client/cli/workflow.py
```diff
@@ -1049,6 +1050,20 @@ def workflow_logs(
                 page,
                 size,
             )
+    except WorkflowLogsPrunedError as e:
+        if json_format:
```
tiborsimko, line 1055: **🤖 PR783-02: `--follow --json` says JSON is ignored but emits it on HTTP 410 (NIT)**

When both options are supplied, the command prints:

```text
Ignoring --json as it cannot be used together with --follow.
```

but leaves `json_format` set. If the polling request then reports pruned logs,
the new `WorkflowLogsPrunedError` handler checks the unchanged flag and emits a
JSON object. The same invocation therefore says that JSON is being ignored and
then uses JSON for its result.

Suggested fix — clear `json_format` after emitting the incompatibility warning,
or reject the option combination. Add a pruned-log test for `--follow --json`
that verifies the chosen behaviour and output format.

reana_client/errors.py
```diff
@@ -19,3 +19,12 @@ class FileDeletionError(Exception):
 
 class EnvironmentValidationError(Exception):
     """REANA workflow environment validation didn't succeed."""
+
+
+class WorkflowLogsPrunedError(Exception):
```
tiborsimko, line 24: **🤖 PR783-03: Modified error module omits the 2026 copyright year (NIT)**

`reana_client/errors.py` is modified in this PR, but its header remains:

```python
# Copyright (C) 2018, 2021, 2022 CERN.
```

Suggested fix — extend the existing chronological year list to
`2018, 2021, 2022, 2026`.

reana_client/api/client.py
michaelbuchar: Agreed, after reana-commons #557 is published, I will bump both lower bounds to that release, regenerate the dependency artefacts, and validate 410 unmarshalling before #783 merges.

reana_client/cli/workflow.py
michaelbuchar: Cleared the JSON flag when --follow takes precedence and added a pruned-log regression test confirming plain-text output

reana_client/errors.py
michaelbuchar: Updated the modified error module’s copyright year list to include 2026

reana_client/cli/workflow.py
tiborsimko: **🤖 PR783-02 follow-up (2026-08-30): verified fixed at 9d5f5ef**

The `--follow --json` compatibility branch now clears `json_format` after warning that JSON is ignored. The pruned-log error path therefore follows the same text-output behaviour instead of re-enabling JSON conditionally.

The regression test exercises a 410 response with both options and verifies that the JSON field is absent from the output. No residual concern remains for this finding.

reana_client/api/client.py
tiborsimko: **🤖 PR783-01 follow-up (2026-08-30): remains open at 9d5f5ef**

Thanks for confirming the intended release order. Deferring the version bump until reana-commons #557 is published is sensible; this does not block review of the functionality itself and can be handled as part of the release sequence.

The declared compatibility range nevertheless still permits `reana-commons` 0.95.0a22, whose bundled specification does not declare the 410 response. The lowest-dependency CI job installs exactly that version. A released installation can therefore still fail during Bravado response unmarshalling before the new `WorkflowLogsPrunedError` path is reached, while the current mocked `HTTPError` tests do not exercise that schema-validation boundary.

Suggested remaining fix — once reana-commons #557 is published, raise both `reana-commons` lower bounds in `setup.py` and verify the 410 path using the bundled API client in the lowest-dependency job before releasing this change.

reana_client/errors.py
tiborsimko: **🤖 PR783-03 follow-up (2026-08-30): verified fixed at 9d5f5ef**

The modified `reana_client/errors.py` copyright header now includes 2026. No residual concern remains for this finding.

PR #793, opened 2026-09-17
feat(client): save server and TLS settings at login

### LINKED ISSUE #792: client: save connection settings and harmonise error messages
### Current behaviour

The clients remember the server after login, but TLS settings still depend on environment variables. Error messages differ between Python and Go. For example, connecting to a local server with a self-signed certificate produces:

```console
$ env | grep REANA
REANA_SERVER_URL=https://localhost:30443

$ reana-client-go ping
==> ERROR: could not complete token refresh: Post "https://localhost:30443/keycloak/realms/reana/protocol/openid-connect/token": tls: failed to verify certificate: x509: certificate signed by unknown authority

$ reana-client ping
==> ERROR: Could not refresh authentication credentials. Please try again.
```

### Expected behaviour

Both clients should support configuring and saving connection settings at login:

```sh
reana-client login --server https://localhost:30443 --no-tls-verify
reana-client ping
```

Subsequent commands should reuse the saved settings, replacing `REANA_SERVER_URL` and `REANA_SERVER_TLS_VERIFY` that we could retire. (Making `REANA_SERVER_URL` not required would be also a good split from previous REANA 0.9 release series.)

Error messages should be consistent across both clients and give clear guidance for missing configuration, authentication failures and certificate problems.

tiborsimko:
Support `login --server URL` with `--tls-verify` or `--no-tls-verify`,
and reuse the saved settings for subsequent commands. Apply explicit TLS
choices during login and persist them only on success; re-login without
a TLS option keeps the saved choice.

Preserve server selection and settings on failed login, token recovery,
refresh and logout. Keep authentication, API requests, transfers and
notebook opening on the resolved server and TLS policy. Scope the bypass
to that server's HTTPS origin and retain CA-bundle precedence.

Harmonise diagnostics for missing configuration and credentials, and for
transport failures during authentication and `ping`, `list` and `info`.
Explain certificate trust, DNS, refused connections and timeouts while
preserving HTTP response handling and keeping raw request data out of
these transport diagnostics and their DEBUG output.

Identify the selected server, warn when verification is disabled and
show the effective TLS policy. Keep help and version usable when stale
environment exports are present. Cover saved settings, connection
failures and preserved HTTP response handling with regression tests.

BREAKING CHANGE: `REANA_SERVER_URL` and `REANA_SERVER_TLS_VERIFY`
environment variables are no longer client inputs. Unset them and run
`login --server URL`, adding `--no-tls-verify` for a self-signed
development server when needed.

Closes reanahub/reana-client#792


reana_client/api/client.py
```diff
@@ -56,10 +56,15 @@ def _get_current_reana_server_api_client():
...
+    server_url = get_active_server()
+    if not server_url:
+        raise ValueError(NO_SERVER)
     return get_current_api_client(
         component="reana-server",
-        ssl_verify=tls_verify(),
-        server_url=get_active_server(),
+        ssl_verify=tls_verify(server_url),
+        server_url=server_url,
     )
```
CameronMcClymont, line 70: `api/client.py:227` returns `ERROR: INVALID SERVER` for TLS, DNS, refused, and timeout. This error message is inaccurate/misleading since the server may be perfectly valid (it may just be unreachable or certificate trust isn't correctly configured).

CameronMcClymont (changes requested):

reana_client/api/client.py
CameronMcClymont: Suggested fix at `api/client:227`:

```
...
           if e.response.status_code == 403:
               return {"status": "ERROR: INVALID ACCESS TOKEN", "error": True}
           raise Exception(e.response)
  +    except requests.RequestException:
  +        # The request never reached the cluster: the certificate, the name
  +        # resolution, the connection or the timeout failed. Let the caller
  +        # explain it through ``connection_error``; reporting it as an invalid
  +        # server would hide the reason and the way to fix it.
  +        raise
       except Exception:
           return {"status": "ERROR: INVALID SERVER", "error": True}
...
```

tiborsimko: Addressed in `15fbaa7`, with the fix squashed into the feature commit. Applied the suggested exception handling and extended the shared diagnostics to `list` and `info`. Regression tests cover TLS, DNS, refused connections and timeouts, preserving 403/404 handling. Extending these diagnostics to the remaining commands is a separate follow-up.

CameronMcClymont (approved):
🦭

PR #796, opened 2026-09-18
feat(client): manage saved REANA server connections

### LINKED ISSUE #795: cli: manage saved REANA servers
## Current behaviour

With connection environment variables retired, `login --server URL` saves and selects a REANA server. Switching back to another saved server requires logging in again.

## Expected behaviour

Both Python and Go clients should support managing saved servers and switching between them without logging in again:

    reana-client server-add https://reana.example.org
    reana-client server-add https://reana.example.com
    reana-client server-list
    reana-client server-use https://reana.example.org
    reana-client login

Saved servers could be removed with `server-remove URL`.

tiborsimko:
Add `server-add`, `server-list`, `server-use` and `server-remove` in a
server connection management command family. Save a server without
logging in or selecting it, and switch between saved connections without
authenticating again.

Keep credentials and TLS settings when switching. List saved URLs, the
current selection and effective TLS verification without exposing token
material. Show invalid TLS records without preventing the listing.
Allow these commands with stale connection exports, ignoring their
values while retaining migration errors for ordinary commands.

Revoke refresh credentials before removing a server and retain its
record if revocation fails. Explain transport and HTTP failures against
the requested server without exposing request data or redirect URLs.
Offer `--local-only` for deliberate removal without revocation, warning
only when a refresh token was present. Removing the selected server
leaves no selection.

Preserve the shared credential-file format, implicit TLS defaults and
unknown fields. Serialise removal with token refresh and retain TLS
verification for identity providers outside the REANA server origin.

Cover the command lifecycle, stale exports, TLS policy, revocation
failures and concurrent refresh behaviour. Update generated command
help and completions.

Closes reanahub/reana-client#795


CameronMcClymont (dismissed):
LGTM, just the commitlint to fix.

tiborsimko:
Added the PR number to the commit headline. Also incorporated a fix from local review: `server-remove` now explains revocation failures while preserving the saved record, with regression coverage. This changes the previously approved version beyond the headline correction. Everything is squashed into the feature commit.

CameronMcClymont (dismissed):
🦭

